PR #2, opened 2020-09-23
Feature/initial types

calebzulawski:
Adds initial type implementations, plus some fundamental conversion traits.

Most of the work is in `macros.rs`, the modules for each type just calls the definition macro and any implements any type-specific conversion traits.  I only implemented x86 conversions for now--not sure how to handle unstable conversions, if at all.


crates/core_simd/Cargo.toml
```diff
@@ -0,0 +1,5 @@
+[package]
```
Lokathor, line 1: technically we will also need `license` and `description` here if we want to publish to crates.io, but that's no big deal at the moment, and we don't necessarily want that at all anyway. Just that some other parts of the standard library are also available as individual crates (libm, hashbrown, etc), so maybe we could do that if we thought it would make it easier to have people test out intermediate work.

crates/core_simd/src/macros.rs
```diff
@@ -0,0 +1,143 @@
...
+    };
+    { @def $(#[$attr:meta])* | $name:ident | $type:ty | $($itype:ty,)* | $($ivar:ident,)* } => {
+        $(#[$attr])*
+        #[allow(non_camel_case_types)]
+        #[derive(Copy, Clone, Debug, Default, PartialEq, PartialOrd)]
+        #[repr(simd)]
+        pub struct $name($($itype),*);
+
+        impl $name {
+            /// Construct a vector by setting each lane to a single value.
```
Lokathor, line 129:
```suggestion
            /// Construct a vector by setting all lanes to the given value.
```

crates/core_simd/src/macros.rs
```diff
@@ -0,0 +1,143 @@
...
+            #[inline]
+            pub fn splat(value: $type) -> Self {
+                Self($(value as $itype),*)
+            }
+
+            /// Construct a vector by setting each lane.
```
Lokathor, line 135:
```suggestion
            /// Construct a vector by setting each lane to the given values.
```

crates/core_simd/src/macros.rs
```diff
@@ -0,0 +1,143 @@
...
+        $(#[$attr])*
+        #[allow(non_camel_case_types)]
+        #[derive(Copy, Clone, Debug, Default, PartialEq, PartialOrd)]
+        #[repr(simd)]
+        pub struct $name($($itype),*);
+
+        impl $name {
+            /// Construct a vector by setting each lane to a single value.
+            #[inline]
+            pub fn splat(value: $type) -> Self {
```
Lokathor, line 131:
```suggestion
            pub const fn splat(value: $type) -> Self {
```

crates/core_simd/src/macros.rs
```diff
@@ -0,0 +1,143 @@
...
+                Self($(value as $itype),*)
+            }
+
+            /// Construct a vector by setting each lane.
+            #[allow(clippy::too_many_arguments)]
+            #[inline]
+            pub fn new($($ivar: $itype),*) -> Self {
```
Lokathor, line 138:
```suggestion
            pub const fn new($($ivar: $itype),*) -> Self {
```

Lokathor (changes requested):
* Oof, 58 modules is a lot. It occurs to me that the widths are actually not *supposed* to be an important part of what's available on a type. So, maybe we should collapse it down so that all the `f32` variants are one module, all the `i32` variants are one module, etc.
* The generic doc comment of "Vector of 16 `i32` *types*" should probably be "Vector of 16 `i32` *values*" instead.
* I'm suspicious of a derived PartialEq in this case. We could make it faster by doing a simd compare and then checking the mask value we get out, but speed isn't an issue for that right now, so i guess this is just a note on what we might want to check later.

**Importantly:** we'll need a set of "mask" types too, which is the output of doing a SIMD comparison. Masks can be combined with other masks using simd bitwise ops similar to combining bools using bitwise ops, and also masks have a "blend" operation which lets you marge an `a` and `b` value according to which lanes in the mask are true or not.

The masks should be a separate type from the main simd types. This is both for correctness (eg: forcing bitwise ops with floats behaves weirdly on `i586`), but also I believe that there's optimization potential to be had in some cases, if I remember properly from reading stuff in the `packed_simd` issue tracker.

I think our name for the mask types should be a very simply `maskWIDTHxLANES`, so types like `mask32x4`, `mask16x8`, etc. The `packed_simd` crate uses just `m`, so `m32x4`, but I feel like *maybe* that's getting too terse. And if people *really* want that terseness they can just write `type m32x4 = mask32x4;` at the top of their crate.

**Less Importantly:** In addition to `usize` and `isize`, the `packed_simd` crate also supports `*const` and `*mut` pointers. I think, since we're currently fully non-generic, that we should *not* support those pointer types initially. You can already use `usize` if you want. Still, it is maybe something to note for people wondering about what we've done differently or not.

crates/core_simd/src/macros.rs
```diff
@@ -0,0 +1,143 @@
...
+    { unsafe $from:ty => $to:ty } => {
+        impl core::convert::From<$from> for $to {
+            #[inline]
+            fn from(value: $from) -> $to {
+                assert_eq!(core::mem::size_of::<$from>(), core::mem::size_of::<$to>());
+                unsafe { (&value as *const $from as *const $to).read_unaligned() }
+            }
+        }
+    }
+}
```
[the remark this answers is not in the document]
calebzulawski: I completely forgot it validates the size already.  I also didn't realize that it would fix the alignment, but that makes sense.  The simplified macros are much simpler.  I also added a macro specifically for converting to x86 types.

crates/core_simd/Cargo.toml
calebzulawski: I agree--for now I just left it my cargo default.  Everyone else can also add their preferred name/email

calebzulawski:
>     * Oof, 58 modules is a lot. It occurs to me that the widths are actually not _supposed_ to be an important part of what's available on a type. So, maybe we should collapse it down so that all the `f32` variants are one module, all the `i32` variants are one module, etc.

Agreed, I realized this partway through the implementation but left it to illustrate that point :) I've combined it by scalar type now.

> I'm suspicious of a derived PartialEq in this case. We could make it faster by doing a simd compare and then checking the mask value we get out, but speed isn't an issue for that right now, so i guess this is just a note on what we might want to check later.

Once we start implementing SIMD ops I agree this is presumably better.

Regarding the mask types, I actually purposely left them out.  Once concern I have is that (at least on x86) float comparisons, return float values, not integers.  Obviously this can be cast, but my understanding is that the cast will force it to break up the uops.  Though if this is transparent to the optimizer it may not matter anyway.

bjorn3:
According to https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_cmp_ps&expand=743 the float comparisons return 0 or !0 bitcasted to a float. Int and float vectors are in the same registers so bitcasting from float to int should emit no instructions.

Lokathor:
Same register, different ALU, so it leads to a pipeline stall when you're doing a series of ops in a row. That is why there's separate intrinsics/asm for f32 & f32 vs i32 & i32.

However, on i586 there's no floating bitand, so you end up casting to u32, doing the and in an integer register, and casting back to f32. Except that cast back to float is *not* bit preserving if the bits would be a signaling NaN: the bits are changed to be a quiet NaN instead.

Which seems to be a problem unique to i586 as far as I know, but we do have to support that target anyway, so... *something*. I'm not positive what to do, but `packed_simd` has mask types and seems to get along fine.

calebzulawski:
One possibility (that I haven't fully considered the implications of) is to have separate fmask types that convert to and from the mask types. On i586 etc you can define this type to be integer if this makes more sense.

calebzulawski:
On further inspection, rustc's `simd_select` platform intrinsic, for example, only accepts signed integer masks anyway from what I can tell.  I suppose the masks must be integer, but I hope it codegens properly.

calebzulawski:
Maybe I'm getting a little too off into the weeds, but the optimizer appears to see straight through it:

https://rust.godbolt.org/z/zYe7je

Note that `vblendvps` is used for the integer example too because there's no i32 blend instruction.  Upping to avx512vl doesn't actually emit `vpblendmd` as I would have expected because it can simply use `vmovdqa` with writemasks and skip a separate blend instruction.

Lokathor:
Can we get impls for `core::fmt::{Binary, Display, LowerExp, LowerHex, Octal, Pointer, UpperExp, UpperHex};`? Only when the element-type supports the given formatter that is.

Each impl should take in the formatter and then use it to call the trait impl of the element type once per element (this lets people apply per-element formatting), with square braces around the output and commas and all that like it's an array/slice. [Example](https://github.com/Lokathor/wide/blob/main/src/lib.rs#L501-L549) from the `wide` crate.

As for `Debug`, I think we should use the same macro impl we come up with for the above just to maintain consistency.

calebzulawski:
Good idea with the additional formatting traits.  I came up with a bit of a hack that lets us still use `core::fmt::DebugList` so it should come out identical to a slice (I don't think there's an easier way?)

Lokathor (approved):
namspacing the intermediate struct and impl to inside of a function. beautiful.

calebzulawski:
Note that I didn't implement `Display`, I'm still not 100% sure that's appropriate and might need a little more discussion (but I assume it would be identical to `Debug`)

crates/core_simd/src/masks.rs
```diff
@@ -0,0 +1,73 @@
...
+        impl core::fmt::Debug for $name {
+            fn fmt(&self, f: &mut core::fmt::Formatter) -> core::fmt::Result {
+                self.test().fmt(f)
+            }
+        }
+    }
+}
+
+define_mask! {
+    #[doc = "8-bit mask"]
```
Stupremee, line 46: Is there a reason why you use `#[doc = "..."]` here instead of `/// ...`? Because `/// ...` will compile to `#[doc]` and thus can be matched by the macro

crates/core_simd/src/pointers.rs
```diff
@@ -0,0 +1,116 @@
...
+        }
+    };
+}
+
+define_pointer_vector! { #[doc = "Vector of two mutable pointers"] mptrx2 => isizex2 => 2, mut }
+define_pointer_vector! { #[doc = "Vector of four mutable pointers"] mptrx4 => isizex4 => 4, mut }
+define_pointer_vector! { #[doc = "Vector of eight mutable pointers"] mptrx8 => isizex8 => 8, mut }
+define_pointer_vector! { #[doc = "Vector of two const pointers"] cptrx2 => isizex2 => 2, const }
+define_pointer_vector! { #[doc = "Vector of four const pointers"] cptrx4 => isizex4 => 4, const }
+define_pointer_vector! { #[doc = "Vector of eight const pointers"] cptrx8 => isizex8 => 8, const }
```
bjorn3, line 116: What is exactly the purpose of pointer vectors?

bjorn3 (approved):

crates/core_simd/src/masks.rs
calebzulawski: I wasn't sure about the expansion order (is that guaranteed to expand before the macro?)

crates/core_simd/src/fmt.rs
```diff
@@ -0,0 +1,106 @@
...
+              (LowerExp, format_lower_exp),
+              (UpperExp, format_upper_exp);
+            )*
+        }
+    };
+    { masks: $($type:ty,)* } => {
+        impl_fmt_trait! {
+            $($type =>
+              (Debug, format);
+            )*
```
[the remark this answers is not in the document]
calebzulawski: That's actually exactly what should happen.  The Debug impl will defer to the inner type's impl, and that's how I implemented Debug for the scalar masks.

crates/core_simd/src/masks.rs
Lokathor: yeah /// becomes a meta attribute before the macro.

crates/core_simd/src/pointers.rs
calebzulawski: I've never done it, but there certainly may be a good reason to perform simultaneous ops on pointers (strided algorithms come to mind).  One thought (@Lokathor ?) is it actually worth providing both a `*const T` and `*mut T` vector?  `core::sync::atomic::AtomicPtr` only operates on `*mut T` and leaves it up to the user to ensure correct mutability.  I propose doing that instead, and just calling it `pointerxN`.

Lokathor: `packed_simd` has them... but I think we can leave them out for now. Let's wait for someone to ask for it.

Because people can also use `usize` if they want.

I don't recall seeing any op at the 128-bit or 256-but level that's like, "take a vector of pointers and load each one into a vector of whatever"

workingjubilee:
I know that packed_simd does it, and LLVM supports a similar intensity of abstraction, but do we *want* to have < 64-bit vector types? I'm not aware of any actual 32-bit vector registers, it creates a lot of API noise to cover every 2^N, and I am ever-so-slightly concerned about introducing interesting bugs in the parts of a vector register we "aren't" using.

I agree that SIMD types should be effectively organized by T, not by T *and* W, in either case.

Lokathor:
Even the 64-bit types feel dubious at best.

calebzulawski:
I think there is some benefit to it when it comes to writing generic algorithms.  I have a feeling LLVM may use partially empty vector registers to implement it, so it would still be generally considered SIMD.

Lokathor (approved):

workingjubilee:
> I think there is some benefit to it when it comes to writing generic algorithms. I have a feeling LLVM may use partially empty vector registers to implement it, so it would still be generally considered SIMD. Remember that with SSE, scalar float ops still use vector registers. I imagine it would be similar for say f32x2.

Using partially empty vector registers is the concern that inspired me to ask, actually.
https://stackoverflow.com/questions/63125919/how-to-avoid-floating-point-exceptions-in-unused-simd-lanes

And scalar code *can* actually beat SIMD code in speed, because it requires less motion to/from the vector registers.

bjorn3:
This PR as is causes rustc to consume 3GB when using `cargo check`. Removing all vector types smaller than 128bit reduces this to 1.5GB.

bjorn3:
Item body checking is a minor bump:

```
time: 0.031; rss: 184MB item_types_checking
time: 3.965; rss: 245MB item_bodies_checking
```

Borrowck is a larger bump:

```
time: 0.073; rss: 248MB misc_checking_2
time: 5.704; rss: 401MB MIR_borrow_checking
```

Encoding the `typeck` query is a huge bump:

```
time: 0.000; rss: 502MB encode_query_results_for(rustc_middle::ty::query::queries::unsafety_check_result)
time: 0.852; rss: 1604MB        encode_query_results_for(rustc_middle::ty::query::queries::typeck)
```

The memory consumed during encoding of the `typeck` query is not freed together with `TyCtxt` for some reason:

```
time: 3.749; rss: 1604MB        codegen_crate
time: 0.040; rss: 1484MB        free_global_ctxt
```

packed_simd also requires a lot of memory, but there the macro expansion already caused the huge memory usage. (https://github.com/rust-lang/rust/issues/76647)

calebzulawski:
A few things.  First of all, I don't really understand the concern with using partially filled vectors. If you are primarily using vertical ops it certainly can still outperform scalar code.  On x86, SSE is just partially filled AVX vectors and AVX is partially filled AVX512.  Additionally, we definitely shouldn't remove 64-bit vectors since NEON has 64-bit registers (and some CPUs actually only perform ops 64-bits at a time, even in the 128-bit registers)

Lokathor:
IIRC, Neon has 128-bit registers only, but the instructions allow for acting on the lower 64 bits with the high bits being zeroed out.

I mean the 64-bit stuff is fine I guess.

@bjorn3 I'm not sure that's a huge deal since people won't generally be building this whole crate locally. They'll have it provided as a part of the pre-built standard library.

Still, some people use `build-std` and for those people we should make an effort.

calebzulawski:
My understanding is that on some CPUs, instructions on the D registers are twice as fast as the Q registers, but yes I believe they're the same register bank.

bjorn3:
> @bjorn3 I'm not sure that's a huge deal since people won't generally be building this whole crate locally. They'll have it provided as a part of the pre-built standard library.

The memory usage is already very bad at this stage. Adding actual operations will probably make it significantly worse. 3GB is already too much to compile it on a 32bit system. It is also bad for everyone that contributes to the standard library or rustc. For me personally like 2GB is already enough to cause swapping. Adding stdsimd as of this PR to libcore would be horrible for me when working on cg_clif.

Lokathor:
Hmm, so what's spiking the ram usage that high? I figured you just had a ton of jobs going from a high-core cpu, and then cutting the jobs down would bring things back in line.

Though when I built the crate on my rpi4 just now it didn't use more than one core. It just maxed out a single core for 50 seconds.

But I'm not sure how to make it speed up short of like, cutting out the 512-bit types or something.

calebzulawski:
I'm not sure I understand why simply having a fairly large number of types is an issue.  Don't other crates like libc have a similarly large number of types, and doesn't core/std itself have significantly more?

workingjubilee:
> A few things. First of all, I don't really understand the concern with using partially filled vectors. If you are primarily using vertical ops it certainly can still outperform scalar code.

I mentioned a performance concern because I believe it affects library usability. Code was committed to notable open-source libraries like OpenSSL that failed to improve performance (and due to throttling issues, even hurt it) using vector registers, but at least in those cases it was easy to form expectations. I would want to think more carefully about usability before supporting more niche cases. Rust does not have explicit support for all hardware numerics that its targets support, either, mostly those which are most common and useful across all platforms.

However, the core concern is that LLVM appears to emit code that actually **does** perform floating point operations in unused SIMD lanes, and this can lead to unexpected behavior by throwing floating point exceptions when a zeroed lane divides by zero. I believe we are quite likely to have our hands full delivering literally just 128-bit vectors, so I am keen to suggest a more minimal API surface area that has less surprising behavior, and to gather intelligence on how that works, before moving to support other uses.

> On x86, SSE is just partially filled AVX vectors and AVX is partially filled AVX512.

The 129th bit onward in the vector register is dark when performing SSE ops, though, and it requires a transition period to activate them. Are you sure it is not the other way around, that AVX is not just 128 bits tacked on to the first 128 bits of an SSE register, and AVX512 is another 256 bits on that?

bjorn3:
> Hmm, so what's spiking the ram usage that high? I figured you just had a ton of jobs going from a high-core cpu, and then cutting the jobs down would bring things back in line.

Only a single rustc instance at a time is active. When not compiled in incremental mode the max-rss is 667MB. Initial incremental build has a max-rss of 1037MB. Second incremental build has a max-rss of 1975MB and takes just 49s. This is when using massif. If I don't use massif the max-rss is 2364MB. I can't reproduce the 3GB, but 2GB is still too big.

Incremental compilation of the standard library is very useful for cg_clif. cg_clif has an option to prevent storing object files in the incremental cache. This allows checking a change to cg_clif that would affect the generated object files, but not type-checking and other passes without unnecessarily run those passes again.

Lokathor:
> I'm not sure I understand why simply having a fairly large number of types is an issue. Don't other crates like libc have a similarly large number of types, and doesn't core/std itself have significantly more?

I'm no expert, but I *suspect* that part of it has to do with the fact that many of the types here are very large on the field count, and rustc doesn't handle 32-field and 64-field types as well as we'd like.

When we combine this with what workingjubilee said, perhaps for the first pass of things we should limit the scope of things.

**Proposal:** we cut down the scope of the types we generated to just *two* widths by default. Either 64/128 or 128/256 would work. I don't think we want a *single* width because we want to stay familiar with being lane-count agnostic as we work, but two settings is probably all we need. Then the other widths can be behind feature flags and we can sort out the build requirements issue down the road.

calebzulawski:
> However, the core concern is that LLVM appears to emit code that actually **does** perform floating point operations in unused SIMD lanes, and this can lead to unexpected behavior by throwing floating point exceptions when a zeroed lane divides by zero. I believe we are quite likely to have our hands full delivering literally just 128-bit vectors, so I am keen to suggest a more minimal API surface area that has less surprising behavior, and to gather intelligence on how that works, before moving to support other uses.

This is a good point.  Playing around in godbolt seems to indicate this is the case.  My understanding, however, is that LLVM (and by extension, Rust) simply doesn't support the floating-point environment and this probably isn't relevant until it does (and would be just one of many issues that need to be solved in LLVM).  I'm concerned about this becoming x86-centric, since supporting 64-bit NEON vectors is just as important as supporting 128-bit vectors on AVX, in my opinion.

workingjubilee:
The floating point semantics for Rust are undecided. To the extent that they are, we are supposed to support IEEE754, which means we should support the floating point environment. Rust may not do so directly right now, but I am thinking carefully about our decisions here and how they may affect future maneuvers around fixing our floating point situation.

I don't think focusing more on certain vector widths is "x86_64 centric", I think it's **64-bit scalar + 128-bit vector** centric, and I don't think it's inappropriate to do that, actually. I would like to see us support not only ARM Neon but PowerPC's AltiVec/VSX as well, and certainly of PowerPC architectures it will be easiest to get support for powerpc64le, and certainly of ARM it will be easiest to support aarch64, and the concerns for 64-bit vectors will be less there as well. ARM SVE(2) and Helium are also their own barrels of fun that we haven't even cracked open yet.

If we isolate certain bitwidths, I think we should focus initially on 128-bit and 256-bit vectors, and then fan out from there.

calebzulawski:
Fair enough, I just think we don't want to preclude it because of an LLVM bug that can't even appear in stable rust.  FWIW, there are still some fairly common places that Cortex-A9s (which operate 64-bits at a time) are still used, Xilinx SoCs come to mind.

Lokathor:
We could probably do 64/128/256 and just skip on 512 initially.

calebzulawski:
The culprit seems to have been the derived traits.  I'm guessing they provided extra bounds on the fields that we don't really need to explicitly list as part of the trait.  Compiling for me is now about 6x faster and uses a "normal" amount of memory.  @bjorn3 can you confirm on your end?

Lokathor:
the manual derives will need `#[inline]` too

calebzulawski:
Added.  Now I think the matter of including sub-128-bit or sub-64-bit vectors is separated from the build time issue.  Personally I vote for keeping them in with the option of either removing them in the future or specifically addressing divide-by-zero.  Note that integer vectors will already need to be specialized for `Div` and panic on zeros in the divisor.

bjorn3:
> The culprit seems to have been the derived traits. I'm guessing they provided extra bounds on the fields that we don't really need to explicitly list as part of the trait. Compiling for me is now about 6x faster and uses a "normal" amount of memory. @bjorn3 can you confirm on your end?<

Thanks that fixed it! Non-incremental uses 199MB and takes 1.8s. First incremental build uses 290MB and takes 2.3s. Next incremental build uses 242MB and takes 0.7s.

workingjubilee:
> Personally I vote for keeping them in with the option of either removing them in the future or specifically addressing divide-by-zero.
Well, what would change to provoke us to remove it later? I mentioned that as a specific example of a general concern re: API surface area.

Lokathor:
sub-64 sounds... not useful? Like i think it's completely fair to have 64-bit simd be the minimum we support.

calebzulawski:
Okay, I've removed sub-64-bit vectors.

At the risk of extending this already long PR, one final question: should we further namespace the types?  Something like `std::simd::mask::*` for the mask types and `std::simd::vector::*` for the vector types?  I figure this may get even more confusing once we add traits.

Lokathor:
Nah, keep it flat.

I'd like to have at least one test somewhere so that the CI setup can be checked. How about just a simple "f32x4 defaults to all 0.0" sample test and then we merge it?

workingjubilee:
Agree with keep it flat for now, I'm happy to rearrange things a bunch of times once the rubber hits the road but a slightly big but thematically coherent module is better than the `std::sync` situation.

workingjubilee (approved):

PR #8, opened 2020-09-27
Add a beginner's guide

Lokathor:
Surely there will be more to say later, but i think this is already useful now, and getting it started at all will hopefully percolate more tips from others.


beginners-guide.md
```diff
@@ -0,0 +1,65 @@
+
+# Beginner's Guide To SIMD
+
+Hello and welcome to our SIMD basics guide!
+
+Because SIMD is a subject that many programmers haven't worked with before, we thought that it's best to outline some terms and other basics for you to get started with.
+
+## Quick Background
+
+**SIMD** stands for *Single Instruction, Multiple Data*. In other words, SIMD is when the CPU performs a single action on more that one logical pieces of data at the same time. Instead of adding two registers that each contain one `f32` value and getting an `f32` as the result, you might add two registers that each contain `f32x4` (128 bits of data) and then you get an `f32x4` as the output.
```
calebzulawski, line 10: s/that/than

beginners-guide.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Terms
+
+SIMD has a few special vocabulary terms you should know:
+
+* **Vector:** A SIMD value is called a vector. This shouldn't be confused with the `Vec<T>` type. A SIMD vector has a fixed size, known at compile time. All of the elements within the vector are of the same type. This makes vectors *similar to* arrays. One difference is that a vector is generally aligned to its *entire* size (eg: 16 bytes, 32 bytes, etc), not just the size of an individual element. Sometimes vector data is called "packed" data.
+
+* **Lane:** A single element position within a vector is called a lane. If you have `N` lanes available then they're numbered from `0` to `N-1` when referring to them, again like an array. The biggest difference between an array element and a vector lane is that it is *relatively costly* to access an individual lane value. Generally, the vector has to be pushed out of register onto the stack, then an individual lane is accessed while it's on the stack. For this reason, when working with SIMD you should avoid reading or writing the value of an individual lane during hot loops.
```
calebzulawski, line 22: I'm not sure this is always true. My understanding is some architectures (like ARM + VFP) can operate directly on a lane.  Maybe just say that it _may_ require extra instructions to move it into a GP register?

beginners-guide.md
```diff
@@ -0,0 +1,65 @@
...
+
+* **Horizontal:** When an operation is "horizontal", the lanes within a single vector interact in some way. A "horizontal add" might add up lane 0 of `a` with lane 1 of `a`, with the total in lane 0 of `out`.
+
+* **Target Feature:** Rust calls a CPU architecture extension a `target_feature`. Proper SIMD requires various CPU extensions to be enabled (details below). Don't confuse this with `feature`, which is a Cargo crate concept.
+
+## Target Features
+
+When using SIMD, you should be familiar with the CPU feature set that you're targeting.
+
+On `arm` and `aarch64` it's fairly simple. There's just one CPU feature that controls if SIMD is available: `neon` (or "NEON", all caps, as the ARM docs often put it). Neon registers are 128-bit, but they can also operate as 64-bit (the high lanes are just zeroed out).
```
calebzulawski, line 36: Each Q register actually maps to a pair of D registers, so it doesn't zero out the other lanes, they operate independently as far as I know.

beginners-guide.md
```diff
@@ -0,0 +1,65 @@
...
+
+If you know that you're targeting a specific CPU you can instead use the [target-cpu](https://rust-lang.github.io/packed_simd/perf-guide/target-feature/rustflags.html#target-cpu) flag and the compiler will enable the correct set of features for that CPU.
+
+The [Steam Hardware Survey](https://store.steampowered.com/hwsurvey/Steam-Hardware-Software-Survey-Welcome-to-Steam) is one of the few places with data on how common various CPU features are. The dataset is limited to "the kinds of computers owned by people who play computer games", so the info only covers `x86`/`x86_64`, and it also probably skews to slightly higher quality computers than average. Still, we can see that the `sse` levels have very high support, `avx` and `avx2` are quite common as well, and the `avx-512` family is still so early in adoption you can barely find it in consumer grade stuff.
+
+## Running a program compiled for a CPU feature level that the CPU doesn't support is automatic undefined behavior.
```
calebzulawski, line 55: Should we mention function multiversioning at all? It's not really part of this project, but it is relevant to this issue.

calebzulawski (changes requested):

beginners-guide.md
Lokathor: Alright, I tried to make it more "generally" and "usually"

beginners-guide.md
Lokathor: Ah, yeah, sorta. I have to [look it up](https://community.arm.com/developer/tools-software/oss-platforms/b/android-blog/posts/arm-neon-programming-quick-reference) because they describe it kinda weird:
* Armv7-A and AArch32 have 32 x 64-bit NEON registers (D0-D31). These registers can also be viewed as 16x128-bit registers (Q0-Q15). Each of the Q0-Q15 registers maps to a pair of D registers, as shown in the following figure.
* AArch64 has 32 x 128-bit NEON registers (V0-V31). These registers can also be viewed as 32-bit Sn registers or 64-bit Dn registers.

beginners-guide.md
Lokathor: Yes we should.

But i'm gonna call it a "todo for later" thing.

calebzulawski (approved):

beginners-guide.md
workingjubilee: Yeah, ARMv8 Neon SIMD registers are enums, not structs like ARMv7 Neon SIMD, which means that even in aarch32, its compatibility mode with ARMv7, we will require careful testing.

Lokathor: @workingjubilee can you open an issue for this so we can track the info?

PR #3, opened 2020-09-23
Initial CI using travis

KodrAus:
This PR bootstraps some CI configuration for us. There's very little here right now.

I've gone with Travis instead of GitHub Actions here because Travis [has ARM support](https://blog.travis-ci.com/2019-10-07-multi-cpu-architecture-support).

This PR is still WIP and will probably be noisy, but I've opened it up directly on the repository here and not marked it as WIP just to make sure we start getting our builds triggered properly.

I'll ping when it's ready!

PR #10, opened 2020-09-28
Implement `core::ops`

### LINKED ISSUE #5: impl `core::ops` for `core::simd`
* [ ] Add / AddAssign
* [ ] BitAnd / BitAndAssign
* [ ] BitOr / BitOrAssign
* [ ] BitXor / BitXorAssign
* [ ] Div / DivAssign
* [ ] Index/IndexMut
* [ ] Mul / MulAssign
* [ ] Neg
* [ ] Not
* [ ] Rem / RemAssign
* [ ] Shl / ShlAssign
* [ ] Shr / ShrAssign
* [ ] Sub / SubAssign

Most of these are self-explanitory.

* Generally, `vector op scalar` should "auto-splat" the scalar value and then perform the op between the two vectors. This is just good quality of life / ergonomics.
* I'd like if *all* types support bit ops (including Not), even though normally `f32` / `f64` don't support bit ops. I consider the lack of bit ops on the floating primitives to be a "not cool" part of `core`.

calebzulawski:
Closes #5.

Note that none of this is tested--I'm not actually sure where to start with testing.  @workingjubilee any recommendations from packed-simd?

Also note that I didn't implement bitwise ops for floats.  We can discuss it, but a solution that mirrors `f32` and `f64` are `to_bits` and `from_bits` fns that use `u32` and `u64` vectors.

### DIFF
--- a/crates/core_simd/src/intrinsics.rs
+++ b/crates/core_simd/src/intrinsics.rs
@@ -1,0 +1,12 @@
+extern "platform-intrinsic" {
+    pub(crate) fn simd_add<T>(x: T, y: T) -> T;
+    pub(crate) fn simd_sub<T>(x: T, y: T) -> T;
+    pub(crate) fn simd_mul<T>(x: T, y: T) -> T;
+    pub(crate) fn simd_div<T>(x: T, y: T) -> T;
+    pub(crate) fn simd_rem<T>(x: T, y: T) -> T;
+    pub(crate) fn simd_shl<T>(x: T, y: T) -> T;
+    pub(crate) fn simd_shr<T>(x: T, y: T) -> T;
+    pub(crate) fn simd_and<T>(x: T, y: T) -> T;
+    pub(crate) fn simd_or<T>(x: T, y: T) -> T;
+    pub(crate) fn simd_xor<T>(x: T, y: T) -> T;
+}
--- a/crates/core_simd/src/intrinsics.rs
+++ b/crates/core_simd/src/intrinsics.rs
@@ -1,12 +1,39 @@
+//! This module contains the LLVM intrinsics bindings that provide the functionality for this
+//! crate.
+//!
+//! The LLVM assembly language is documented here: https://llvm.org/docs/LangRef.html
+
+/// These intrinsics aren't linked directly from LLVM and are mostly undocumented, however they are
+/// simply lowered to the matching LLVM instructions by the compiler.  The associated instruction
+/// is documented alongside each intrinsic.
 extern "platform-intrinsic" {
+    /// add/fadd
     pub(crate) fn simd_add<T>(x: T, y: T) -> T;
+
+    /// sub/fsub
     pub(crate) fn simd_sub<T>(x: T, y: T) -> T;
+
+    /// mul/fmul
     pub(crate) fn simd_mul<T>(x: T, y: T) -> T;
+
+    /// udiv/sdiv/fdiv
     pub(crate) fn simd_div<T>(x: T, y: T) -> T;
+
+    /// urem/srem/frem
     pub(crate) fn simd_rem<T>(x: T, y: T) -> T;
+
+    /// shl
     pub(crate) fn simd_shl<T>(x: T, y: T) -> T;
+
+    /// lshr/ashr
     pub(crate) fn simd_shr<T>(x: T, y: T) -> T;
+
+    /// and
     pub(crate) fn simd_and<T>(x: T, y: T) -> T;
+
+    /// or
     pub(crate) fn simd_or<T>(x: T, y: T) -> T;
+
+    /// xor
     pub(crate) fn simd_xor<T>(x: T, y: T) -> T;
 }
--- a/crates/core_simd/src/masks.rs
+++ b/crates/core_simd/src/masks.rs
@@ -66,34 +66,6 @@ macro_rules! define_mask {
                 self.test().fmt(f)
             }
         }
-
-        impl core::ops::BitAnd for $name {
-            type Output = Self;
-            fn bitand(self, rhs: Self) -> Self {
-                Self(self.0 & rhs.0)
-            }
-        }
-
-        impl core::ops::BitOr for $name {
-            type Output = Self;
-            fn bitor(self, rhs: Self) -> Self {
-                Self(self.0 | rhs.0)
-            }
-        }
-
-        impl core::ops::BitXor for $name {
-            type Output = Self;
-            fn bitxor(self, rhs: Self) -> Self {
-                Self(self.0 ^ rhs.0)
-            }
-        }
-
-        impl core::ops::Not for $name {
-            type Output = Self;
-            fn not(self) -> Self {
-                Self(!self.0)
-            }
-        }
     }
 }
 
--- a/crates/core_simd/src/masks.rs
+++ b/crates/core_simd/src/masks.rs
@@ -66,6 +66,30 @@ macro_rules! define_mask {
                 self.test().fmt(f)
             }
         }
+
+        impl core::fmt::Binary for $name {
+            fn fmt(&self, f: &mut core::fmt::Formatter) -> core::fmt::Result {
+                <$type as core::fmt::Binary>::fmt(&self.0, f)
+            }
+        }
+
+        impl core::fmt::Octal for $name {
+            fn fmt(&self, f: &mut core::fmt::Formatter) -> core::fmt::Result {
+                <$type as core::fmt::Octal>::fmt(&self.0, f)
+            }
+        }
+
+        impl core::fmt::LowerHex for $name {
+            fn fmt(&self, f: &mut core::fmt::Formatter) -> core::fmt::Result {
+                <$type as core::fmt::LowerHex>::fmt(&self.0, f)
+            }
+        }
+
+        impl core::fmt::UpperHex for $name {
+            fn fmt(&self, f: &mut core::fmt::Formatter) -> core::fmt::Result {
+                <$type as core::fmt::UpperHex>::fmt(&self.0, f)
+            }
+        }
     }
 }
 
--- a/crates/core_simd/src/masks.rs
+++ b/crates/core_simd/src/masks.rs
@@ -87,6 +87,13 @@ macro_rules! define_mask {
                 Self(self.0 ^ rhs.0)
             }
         }
+
+        impl core::ops::Not for $name {
+            type Output = Self;
+            fn not(self) -> Self {
+                Self(!self.0)
+            }
+        }
     }
 }
 
--- a/crates/core_simd/src/masks.rs
+++ b/crates/core_simd/src/masks.rs
@@ -1,3 +1,13 @@
+/// The error type returned when converting an integer to a mask fails.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub struct TryFromMaskError(());
+
+impl core::fmt::Display for TryFromMaskError {
+    fn fmt(&self, f: &mut core::fmt::Formatter) -> core::fmt::Result {
+        write!(f, "mask must have all bits set or unset")
+    }
+}
+
 macro_rules! define_mask {
     { $(#[$attr:meta])* struct $name:ident($type:ty); } => {
         $(#[$attr])*
@@ -34,9 +44,47 @@ macro_rules! define_mask {
             }
         }
 
+        impl core::convert::TryFrom<$type> for $name {
+            type Error = TryFromMaskError;
+            fn try_from(value: $type) -> Result<Self, Self::Error> {
+                if value == 0 || !value == 0 {
+                    Ok(Self(value))
+                } else {
+                    Err(TryFromMaskError(()))
+                }
+            }
+        }
+
+        impl core::convert::From<$name> for $type {
+            fn from(value: $name) -> Self {
+                value.0
+            }
+        }
+
         impl core::fmt::Debug for $name {
             fn fmt(&self, f: &mut core::fmt::Formatter) -> core::fmt::Result {
                 self.test().fmt(f)
+            }
+        }
+
+        impl core::ops::BitAnd for $name {
+            type Output = Self;
+            fn bitand(self, rhs: Self) -> Self {
+                Self(self.0 & rhs.0)
+            }
+        }
+
+        impl core::ops::BitOr for $name {
+            type Output = Self;
+            fn bitor(self, rhs: Self) -> Self {
+                Self(self.0 | rhs.0)
+            }
+        }
+
+        impl core::ops::BitXor for $name {
+            type Output = Self;
+            fn bitxor(self, rhs: Self) -> Self {
+                Self(self.0 ^ rhs.0)
             }
         }
     }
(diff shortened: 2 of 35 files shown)


Lokathor:
I think that this is effectively blocked on having CI basics in place to run the tests. Maybe we can have that be a subject at today's meeting.

The tests themselves can be simple math tests in a `tests/` folder like any normal integration test suite. Example inputs, expected output, assert that the actual output is equal to (or within a tolerance of) the actual output.

For the specific organization of this sort of "many small code bits" crate, I usually make one `tests/` file per `src/` file, and then each test name is generally one of "impl_[trait]_for_[type]", "type_[method]", or "test_[property]".

I'd be happy to just donate the entire test suite from `wide` to this crate. Other than possible renames here or there, I expect that `stdsimd` should be able to pass any test that `wide` does. It doesn't cover all the stuff that we'd want since it's focused on actual hardware ops, so there are no tests for integer simd division for example, but it's a good start probably.

workingjubilee:
We will eventually want to build up to running most of the same set of [example projects](https://github.com/rust-lang/packed_simd/tree/master/examples) that packed_simd did, as well. They look a lot like real programs and will provide useful full-scale integration testing and benchmarks, essentially.

Lower things will be useful for revealing analytical diagnostics, and there I am mostly indifferent as long as the tests are easily findable. packed_simd had macroed tests as part of its impls, and there I often found it difficult to figure out where the test and the implementation code diverged.

Lokathor:
Heavy macros in the tests sounds pretty gross.

However, a macro that can take in the inputs+outputs for widest possible value we want to test (say: `f32x16`), and then automatically generate the same test test at lower lane counts (eg: `f32x8`, `f32x4`, `f32x2`) by just skipping the values at the end of the list, well that does sound extremely useful.

workingjubilee:
This PR is waiting on #3 to go first.

crates/core_simd/src/intrinsics.rs
```diff
@@ -0,0 +1,12 @@
+extern "platform-intrinsic" {
```
Lokathor, line 1: If we have a link to LLVM's docs about how these work we should add it here.

calebzulawski:
I'll take that opportunity to clean up a couple things.

crates/core_simd/src/masks.rs
```diff
@@ -1,3 +1,13 @@
+/// The error type returned when converting an integer to a mask fails.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub struct TryFromMaskError(());
```
Lokathor, line 3: should this carry the bad value as a payload and then you can give the bad value as part of the `Display` impl?

crates/core_simd/src/ops.rs
```diff
@@ -0,0 +1,518 @@
...
+        impl core::ops::$trait for &'_ $type {
+            type Output = <$type as core::ops::$trait>::Output;
+            fn $fn($self_tok) -> Self::Output {
+                core::ops::$trait::$fn(*$self_tok)
+            }
+        }
+    }
+}
+
+macro_rules! impl_mask_ops {
```
Lokathor, line 98: Does this also impl the `FooAssign` traits for the masks as well? There's so much macro stuff I can't quite tell.

crates/core_simd/src/ops.rs
```diff
@@ -0,0 +1,518 @@
...
+        impl_op! { @binary $type, $rhs, BitXor::bitxor, BitXorAssign::bitxor_assign, simd_xor }
+    };
+
+    // Neg and Not implementations
+    { impl Not for $type:ty } => {
+        impl_ref_ops! {
+            impl core::ops::Not for $type {
+                type Output = Self;
+                fn not(self) -> Self::Output {
+                    self ^ <$type>::splat(true.into())
```
Lokathor, line 213: How is `true.into()` becoming an all 1 bits value for... well any type at all?

crates/core_simd/src/ops.rs
```diff
@@ -0,0 +1,518 @@
...
+            }
+        }
+    };
+
+    { impl Neg for $type:ty } => {
+        impl_ref_ops! {
+            impl core::ops::Neg for $type {
+                type Output = Self;
+                fn neg(self) -> Self::Output {
+                    self * <$type>::splat(-1 as <$type as crate::Vector>::Scalar)
```
Lokathor, line 224: I normally do `0 - val` for Neg, since Sub is usually cheaper than Mul.

Would that give wrong values for any of the types?

KodrAus:
Alrighty, after a rebase this should pick up our initial CI

crates/core_simd/src/intrinsics.rs
calebzulawski: I found the language guide at one point--let me find that.

crates/core_simd/src/masks.rs
calebzulawski: I copied `TryFromIntError` for this, which doesn't hold the value.  Since the types are `Copy` I'm guessing it's easy enough for the user to do that if they want it.

crates/core_simd/src/ops.rs
calebzulawski: Masks implement `From<bool>`, which then gets splatted

crates/core_simd/src/ops.rs
calebzulawski: I wasn't sure how to implement `Neg` and this is what `packed-simd` does.  I assume that in reality both would get optimized to an actual neg op (like an xor or something?).  Maybe this optimizes better? Or maybe it doesn't matter at all?

crates/core_simd/src/ops.rs
calebzulawski: Whoops, nope, doesn't look like it.  I tried to leave the macros fairly explicit.  `impl_ref_ops` only adds matching implementations on references.

crates/core_simd/src/ops.rs
Lokathor: Oh, do the integer vectors also get a `Not` impl elsewhere that I missed?

crates/core_simd/src/ops.rs
Lokathor: Let's put a TODO to check on it more carefully later on.

crates/core_simd/src/intrinsics.rs
calebzulawski: Added a link to the language guide and documented which LLVM instructions are being used.

crates/core_simd/src/ops.rs
calebzulawski: Oh whoops, this is definitely wrong for integers.

crates/core_simd/src/ops.rs
calebzulawski: I actually just changed it to use `0 - val` since it was much simpler to implement (`$type::default - self`)

Lokathor (commented):
I think most of the vec/scalar ops should be commutative when the op is.

that is, if you can do `f32x4 + f32 -> f32x4`, you should probably also be able to do `f32 + f32x4 -> f32x4`.

Obviously this would only apply to Op, not OpAssign variants.

Lokathor (approved):

Lokathor:
All set to merge?

calebzulawski:
I just pushed some floating point tests (which uncovered that `Neg` can't be implemented with `0 - val`, since it doesn't work for -0).  If we're okay with this test style I'll expand it to the other types as well.

programmerjake:
`-0.0 - val` works for neg assuming LLVM's default rounding mode (round to nearest, ties to even).

workingjubilee:
I am not sure how I feel about the f32x4 + f32 example. I understand the intent behind including the operation but I am not sure if that is the best API for it, as my internal sense of what that would express actually *is* non-commutative:
```c
f32x4 + f32 // -> f32x4, the rhs f32 added to each lane of lhs
f32 + f32x4 // -> f32, lhs added to the summation of rhs
```
And I feel the rhs vector is more ambiguous in general. And I must note, in general Rust is very conservative with extending Add impls even in "obvious" cases, e.g. compare i32 + i64. Of course you could have it return an i64, always, but 
```rust
error[E0277]: cannot add `i32` to `i64`
 --> src/main.rs:2:18
  |
2 |     let n = 4i64 + -3i32;
  |                  ^ no implementation for `i64 + i32`
  |
  = help: the trait `std::ops::Add<i32>` is not implemented for `i64`
```

I don't believe packed_simd implemented Add for scalar + vector, but that might have been a technical limitation and not on purpose? In any case, I feel like having a fn to widen a scalar into an appropriate vector would be better than implementing coercion, and I believe with appropriate const generics madness it would be possible to have it handled by type inference(!) so as to make it as ergonomic as possible.

calebzulawski:
Personally, I'm fine with removing the mixed scalar/vector ops.  `a + f32x4::splat(b)` or simply `a + b.into()` isn't too bad in my opinion and _might_ be less likely to be confusing.

Lokathor:
Does `a + b.into()` infer properly?

I'll say that auto-splatting ops impls are one of the things I was almost *immediately* asked about adding when I released `wide-0.5.0` without them. Code can really quickly get noisey without the auto-splatting. Particularly, if you have a scalar function, it's easier to convert to being vector if the literals will automatically promote themselves.

Also, C++ libs generally do have this, and that's another language and all, but it does mean that people using SIMD are generally comfortable with it.

calebzulawski:
Good point, I just tried it and using `Into` doesn't infer correctly.  I'm a little less sold since that doesn't work.  You could always implement a function/trait on scalars to splat them, like `a + b.splat()` but I'm not sure that's any better.

workingjubilee:
...is there a For/Into impl here for that?

C++ and C implement some pretty intense mathematical coercion by default and Rust deliberately chose another path. I would like at least to experiment with an alternative API here because I think there is a fiendish trick I can make work.

Lokathor:
What makes Rust's minimal coercion work out well is that you can still quite casually convert numbers by throwing `as _` onto anything. Since that's not available with simd types, things get wordy fast.

Feel free to continue to try and think of good alternatives, but literals particularly are usually a pain point like this.

calebzulawski:
> ...is there a For/Into impl here for that?
> 
> C++ and C implement some pretty intense mathematical coercion by default and Rust deliberately chose another path. I would like at least to experiment with an alternative API here because I think there is a fiendish trick I can make work.

All of the vectors implement `From<Scalar>`.  I guess it's getting caught up on the generic `Rhs` type.

I'm also not sure that it's really considered coersion--while the scalar gets splatted as far as the actual implementation goes, conceptually it can be thought of without any splat at all.  I'm also not sure that behavior is unique to C++ libs, MATLAB and Wolfram automatically broadcast scalars for elementwise ops.

calebzulawski:
Unrelated to the scalar-vector ops, are we ok with tests in this form?  Should I expand them to the rest of the types?

Lokathor:
It's... quite full of macros, but seems to still make sense.

Also, our sample data probably needs a little more flair than just small positive whole numbers. As soon as we start mixing it up a bit, we'll probably need an ApproxEq as well as just BitEq.

Lokathor (approved):

crates/core_simd/tests/helpers/biteq.rs
```diff
@@ -0,0 +1,127 @@
+pub(crate) trait BitEq {
+    fn biteq(&self, other: &Self) -> bool;
+    fn fmt(&self, f: &mut core::fmt::Formatter) -> core::fmt::Result;
+}
+
+macro_rules! impl_biteq {
+    { integer impl BitEq for $($type:ty,)* } => {
+        $(
+        impl BitEq for $type {
+            fn biteq(&self, other: &Self) -> bool {
```
workingjubilee, line 10:
```suggestion
            fn bit_eq(&self, other: &Self) -> bool {
```

crates/core_simd/tests/ops_impl/float_macros.rs
```diff
@@ -0,0 +1,237 @@
+macro_rules! float_tests {
+    { $vector:ident, $scalar:ident } => {
+        #[cfg(test)]
+        mod $vector {
+            use super::*;
+            use helpers::lanewise::*;
+
+            // TODO change to read from slice
```
workingjubilee, line 8: What does this mean?

crates/core_simd/src/ops.rs
```diff
@@ -0,0 +1,624 @@
...
+                    core::ops::$assign_trait::$assign_trait_fn(self, <$type>::splat(rhs));
+                }
+            }
+        }
+    };
+}
+
+/// Implements relevant operators for the provided types.
+macro_rules! impl_op_meta {
+    { float: $($scalar:ty => $($vector:ty),*;)* } => {
```
workingjubilee, line 271: Using a magic word inside the macro which must be given when the macro is called and causes it to branch to a different logical implementation suggests that what is desired is to simply use a different name.

crates/core_simd/tests/helpers/biteq.rs
Lokathor: That would actually run against the convention for the bit ops, where and, or, and xor were defined with methods bitand, bitor, and bitxor.

I mean it could have the underscore anyway.

workingjubilee:
We will always want to test:
desired interactions with ty::MAX and ty::MIN
desired interactions with 1 and -1
desired interactions with `0 as ty`
Because these cases are exceedingly simple and exceedingly obvious at the same time.

workingjubilee:
I think for very large and small values we can use a crate like [quickcheck](https://github.com/BurntSushi/quickcheck) to test our assumptions there, and we can defer those until later. NaN and the infinities are important also though, but we're going to have a hard time testing NaN due to LLVM's canonicalizing behavior with them.

Lokathor:
Well theoretically all the NaN are equally "you'll get nonsense", we shouldn't have to worry about specific nan bit patterns. That would itself be non-portable i think.

Also, it seems like we've got 16 lanes or so of test data space, and we can run the op a second time to get a second set of 16 cases, so there's no big harm in having some preset "weird input" cases. We can also have quickcheck of course.

crates/core_simd/tests/helpers/biteq.rs
workingjubilee: Oh, I didn't actually notice that the bit ops were weird like that. I retract my suggestion then!

crates/core_simd/tests/ops_impl/float_macros.rs
calebzulawski: Hm.  Could clear up this comment.  Presumably in the future there will be a function that reads vectors from slices.

workingjubilee: Ah, much clearer!

crates/core_simd/src/ops.rs
workingjubilee: Looks good!

workingjubilee (approved):

calebzulawski:
Ok, didn't implement any edge cases (and I definitely like the idea of something like quickcheck) but everything should work, nominally.

workingjubilee:
I think we can TODO more tests then and rebase-and-merge this on green to unblock parallelized work on vector math and more tests, then.

PR #20, opened 2020-10-02
Add CI for wasm targets using wasm-bindgen

KodrAus:
This PR introduces CI for `wasm32-unknown-unknown` along with its `simd128` target feature. The SIMD support is tested through Chrome, only because it seemed easiest to set up with experimental flags.

Adding WebAssembly tests is unfortunately a bit intrusive, we need an extra attribute on every `#[test]`, but it all seems to work!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -230,6 +230,27 @@ matrix:
       arch: amd64
       env:
         - TARGET=x86_64-apple-darwin
+    
+    # WebAssembly (wasm-bindgen)
+
+    - name: "wasm32-unknown-unknown (node)"
+      os: linux
+      arch: amd64
+      install:
+        - curl https://rustwasm.github.io/wasm-pack/installer/init.sh -sSf | sh
+      script:
+        - wasm-pack test --node crates/core_simd
+    
+    - name: "wasm32-unknown-unknown+simd128 (chrome)"
+      os: linux
+      arch: amd64
+      addons:
+        chrome : stable
+      install:
+        - curl https://rustwasm.github.io/wasm-pack/installer/init.sh -sSf | sh
+      script:
+        - RUSTFLAGS="-C target-feature=+simd128"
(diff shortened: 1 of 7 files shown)


.travis.yml
```diff
@@ -230,6 +230,27 @@ matrix:
...
+    - name: "wasm32-unknown-unknown+simd128 (chrome)"
+      os: linux
+      arch: amd64
+      addons:
+        chrome : stable
+      install:
+        - curl https://rustwasm.github.io/wasm-pack/installer/init.sh -sSf | sh
+      script:
+        - RUSTFLAGS="-C target-feature=+simd128"
+        - wasm-pack test --chrome --headless
```
bjorn3, line 253: Would it make sense to also test in firefox?

bjorn3: `wasm-pack` seems to be able to test in multiple environments using a single invocation. 

https://rustwasm.github.io/docs/wasm-pack/commands/test.html?highlight=firefox#test-environment

> `wasm-pack test --node --firefox --chrome --safari --headless`

KodrAus: Do we know if Firefox has SIMD support? I'm testing in Chrome here because it has experimental support with `--enable-features=WebAssemblySimd`.

KodrAus: We could technically test in Node too, but I couldn't figure out how to get command-line Node args passed through `wasm-pack test`

calebzulawski: I went through this process a while ago, I wasn't able to enable SIMD in node or Firefox either, just chrome.  If there is a way to pass node flags it's definitely not obvious.

calebzulawski (approved):
Would it make sense to run the tests without simd128 as well?

KodrAus:
We could consider adding Firefox and Chrome to the existing Node target, which doesn’t use the SIMD feature.

calebzulawski:
Whoops--missed the node target.  I think it's probably fine as-is!

PR #31, opened 2020-10-07
Use xor to implement Neg::neg for floats

thomcc:
This returns the correct on all architectures, and avoids raising float exceptions, canonicalizing values, and setting status registers no matter the argument in both debug/release builds. This is required by IEEE-754, see<sup>1</sup>

Note that without the impl change, I believe we'd need to bring in the CI changes from #30 to actually trigger a failure, but https://github.com/thomcc/stdsimd/runs/1220021695 has examples of the failure cases (don't look at the ones that failed to build, I'm referring to the ones that failed the test this adds).

Ultimately though, this is a placeholder until some sort of neg intrinsic is provided by rustc, as while this is fairly optimal on Intel, that is not true for ARM. That said, this fixes the issue with the old implementation, even if it's not optimal everywhere.

---

1: From IEEE754-2019 section 5.5.1 "Sign bit operations"

> `negate(x)` copies a floating-point operand x to a destination in the same format, reversing the sign bit. negate(x) is not the same as subtraction(0, x)

It also says (when introducing negate/abs/copysign) a few lines earlier:

> The operations treat floating-point numbers and NaNs alike, and signal no exception. These operations may propagate non-canonical encodings.

Note that subtraction/multiplication would violate these requirements.

### DIFF
--- a/crates/core_simd/src/ops.rs
+++ b/crates/core_simd/src/ops.rs
@@ -212,7 +212,18 @@ macro_rules! impl_op {
             impl core::ops::Neg for $type {
                 type Output = Self;
                 fn neg(self) -> Self::Output {
-                    <$type>::splat(-<$scalar>::default()) - self
+                    <$type>::splat(0) - self
+                }
+            }
+        }
+    };
+
+    { impl Neg for $type:ty, $scalar:ty, @float } => {
+        impl_ref_ops! {
+            impl core::ops::Neg for $type {
+                type Output = Self;
+                fn neg(self) -> Self::Output {
+                    Self::from_bits(<$type>::splat(-0.0).to_bits() ^ self.to_bits())
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


crates/core_simd/src/ops.rs
```diff
@@ -212,7 +212,18 @@ macro_rules! impl_op {
             impl core::ops::Neg for $type {
                 type Output = Self;
                 fn neg(self) -> Self::Output {
-                    <$type>::splat(-<$scalar>::default()) - self
+                    <$type>::splat(0) - self
```
thomcc, line 215: I changed this line to {integer} 0 and the line later to -0.0 so that if you forgot or wrongly added the `@float` you'd get a compile error.

Lokathor (approved):

thomcc:
travis seems to have stalled, weirdly.

crates/core_simd/src/ops.rs
```diff
@@ -212,7 +212,18 @@ macro_rules! impl_op {
...
+            }
+        }
+    };
+
+    { impl Neg for $type:ty, $scalar:ty, @float } => {
+        impl_ref_ops! {
+            impl core::ops::Neg for $type {
+                type Output = Self;
+                fn neg(self) -> Self::Output {
+                    Self::from_bits(<$type>::splat(-0.0).to_bits() ^ self.to_bits())
```
programmerjake, line 226: There should probably be a comment to the effect of: replace with `fneg` intrinsic once Rustc exposes that LLVM instruction.

calebzulawski (approved):

Issue #41, opened 2020-10-16
Some floating point test failures on Raspberry Pi 4 with Neon


KodrAus:
Probably because #39 🙂 

I've been testing `stdsimd` a bit on a Raspberry Pi 4 and ran into a couple float test failures with `+neon`:

```
cargo test --target armv7-unknown-linux-gnueabihf
```

succeeds, but:

```
RUSTFLAGS="-C target-feature=+neon" cargo test --target armv7-unknown-linux-gnueabihf
```

fails with:

```
failures:

---- ops_impl::f32::f32x16::fract_odd_floats stdout ----
thread 'ops_impl::f32::f32x16::fract_odd_floats' panicked at 'assertion failed: `(left == right)`
  left: `[0.0 (0), 0.0 (0), 0.0 (0), 0.0 (0), 0.0 (0), 0.0 (0), NaN (7fc00000), NaN (7fc00000), 0.000000000000000000000000000000000000011754944 (800000), -0.000000000000000000000000000000000000011754944 (80800000), 0.00000011920929 (34000000), -0.00000011920929 (b4000000), NaN (7fc00000), NaN (7fc00000), 0.33333206 (3eaaaa80), -0.33333206 (beaaaa80)]`,
 right: `[0.0 (0), 0.0 (0), 0.0 (0), 0.0 (0), 0.0 (0), 0.0 (0), NaN (7fc00000), NaN (7fc00000), 0.000000000000000000000000000000000000011754944 (800000), -0.000000000000000000000000000000000000011754944 (80800000), 0.00000011920929 (34000000), -0.00000011920929 (b4000000), NaN (7fc00000), NaN (ffc00000), 0.33333206 (3eaaaa80), -0.33333206 (beaaaa80)]`', crates/core_simd/tests/ops_impl/f32.rs:6:1
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- ops_impl::f32::f32x2::fract_odd_floats stdout ----
thread 'ops_impl::f32::f32x2::fract_odd_floats' panicked at 'assertion failed: `(left == right)`
  left: `[NaN (7fc00000), NaN (7fc00000)]`,
 right: `[NaN (7fc00000), NaN (ffc00000)]`', crates/core_simd/tests/ops_impl/f32.rs:3:1

---- ops_impl::f32::f32x4::fract_odd_floats stdout ----
[...]

PR #42, opened 2020-10-16
add armv7 target

KodrAus:
In #41 I ran into some test case failures with ARMv7 on a Raspberry Pi. I'm just opening this up to see if I can replicate those on the targets available to use in CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,12 @@ matrix:
       arch: arm64
       env: RUSTFLAGS=-Ctarget-cpu=native
 
+    # Linux (armv7)
+    - name: "armv7-unknown-linux-gnueabihf (neon)"
+      os: linux
+      arch: arm64
+      env: RUSTFLAGS=-Ctarget-feature=+neon
+
     # Linux (powerpc64le)
     - name: "powerpc64le-unknown-linux-gnu (altivec, vsx, power8-*)"
       os: linux


KodrAus:
Alas it succeeds 🙃

thomcc:
I'm hoping to replicate the CI setup from https://github.com/rust-lang/stdarch/ and/or https://github.com/rust-lang/packed_simd this weekend, which (if successful) will help here. That said looking at what you have there, I think you're running on arm64 which is armv8, no?

.travis.yml
```diff
@@ -20,6 +20,12 @@ matrix:
       arch: arm64
       env: RUSTFLAGS=-Ctarget-cpu=native
 
+    # Linux (armv7)
+    - name: "armv7-unknown-linux-gnueabihf (neon)"
+      os: linux
+      arch: arm64
+      env: RUSTFLAGS=-Ctarget-feature=+neon
```
hsivonen, line 27: Adding `+neon` after the fact to the `armv7-unknown-linux-gnueabihf` target results in terrible performance due to the mismatch between the pre-built `std` and what's being built with `+neon`.

The target that doesn't suffer from this problem is `thumbv7neon-unknown-linux-gnueabihf`, which bakes the `neon` option into the target definition and, therefore, comes with `neon`-enabled `std`.

KodrAus:
@thomcc I believe so yes, but am not super familiar with ARM so am deferring to you on it 😄 I was curious to see if I could quickly make it fail the same way in CI for us.

I should have signaled better than this was a quick experiment, and will go ahead and close this.

PR #47, opened 2020-12-01
Use platform intrinsics, not LLVM, for floor & ceil

workingjubilee:
This PR removes the direct linkage to LLVM for trunc and round intrinsics, while replacing that link with rustc's platform intrinsics for floor and ceil functions, namely `simd_floor` and `simd_ceil`. Tests that are no longer testable are removed. In doing so it resolves the riscv64gc compilation problems.

This PR will close #35.

PR #48, opened 2020-12-04
[TEST ONLY] I am bors

workingjubilee:
This is a PR just to test if the problems in #47 are solved by #44.


workingjubilee:
Answer: yes, #47 needs the mask PR to patch it. Also one of them will likely require a rebase.

KodrAus:
It looks like everything is back to green on our main branch after #49 was merged.

PR #49, opened 2020-12-06
Feature/const generics

calebzulawski:
This PR adds a few things:
* Switch to generic-width vectors!
* Overhaul masks (adds an opaque mask type with arbitrary layout, and a trait for using particular layouts)
* Adds comparison operations (supersedes PR #44)

Quite a bit is going on in this PR, so it might be worth building the docs and reading those first to get a hold of the interface, then moving on to the implementation.


programmerjake:
Looks good to me!

KodrAus:
Looks like we're hitting a bug in the codegen:

```
Invalid bitcast

  %4 = bitcast <4 x i32> %1 to [4 x i32], !dbg !272

in function _ZN9core_simd11vectors_i3216SimdI32$LT$_$GT$8to_array17h1010e6a6315f7e36E

LLVM ERROR: Broken function found, compilation aborted!
```

calebzulawski:
I think this is the same issue from #48.  I'm pretty sure it's a compiler bug and not our fault.  Is the bitcast invalid because the source and destination types are the same?

bjorn3:
The source is a vector I think, the destination is an array,

calebzulawski:
It looks like arrays don't cast?  This doesn't compile at all (https://llvm.godbolt.org/z/4xM1MY):
```
define [4 x i32] @cast([4 x float] %x) local_unnamed_addr #0 {
    %out = bitcast [4 x float] %x to [4 x i32]
    ret [4 x i32] %out
}
```
The equivalent with vectors works just fine.

calebzulawski:
Opened https://github.com/rust-lang/rust/issues/80108 to address the error.  Going to try to find a workaround :)

Issue #54, opened 2021-01-07
Unsound Div, Rem


calebzulawski:
The following should panic for integers:
* ~~`neg(MIN)`~~
* `div(MIN, -1)`
* `rem(MIN, -1)`
* `div(x, 0)`
* `rem(x, 0)`

calebzulawski:
This should probably be fixed after or with #49

workingjubilee:
NB: these are all considered UB by LLVM and currently these scalar Rust ops panic (or are even prevented with a "this will panic at runtime" warning).

calebzulawski:
looking again, `neg` in llvm is implemented with `sub` and appears to be allowed to wrap, so maybe that needs to be checked in the compiler for debug?

thomcc:
Is there any chance of having a wrapping_simd_foo version that we could use instead? I suspect in practice the branching here will be very bad for performance.

calebzulawski:
For `neg` I think we want to extend the compiler intrinsics eventually.  For `div` and `rem` the branching is unavoidable (but we can try to optimize it in the future)

workingjubilee:
> looking again, `neg` in llvm is implemented with `sub` and appears to be allowed to wrap, so maybe that needs to be checked in the compiler only in debug mode?

[Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=c86d343408108f9f4f1f73f309d2953e)
```rust
fn main() {
    println!("{}", -i32::MIN);
}
```

>    Compiling playground v0.0.1 (/playground)
    Finished dev [unoptimized + debuginfo] target(s) in 1.69s
     Running `target/debug/playground`
thread 'main' panicked at 'attempt to negate with overflow', src/main.rs:2:20
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

However...

>    Running `target/release/playground`

> -2147483648
So yeah! That one only panics in debug, actually.

thomcc:
> For neg I think we want to extend the compiler intrinsics eventually. For div and rem the branching is unavoidable (but we can try to optimize it in the future)

div/rem are heavyweight operations that this seems fine for, fwiw.

workingjubilee:
I think we have a separate issue on our hands which is re: whether or not add/sub/neg should overflow in Debug mode, and as such have removed that from this issue and into #56.

workingjubilee:
Resolved by #55.

Issue #51, opened 2020-12-08
Improve wasm32 to_array performance


workingjubilee:
It's not clear why these are happening, they seem to be fixed in recent PRs to the mask API but not on purpose? This is being opened to make sure they get addressed eventually anyways.

workingjubilee:
These apparently are mysteriously OK now. ¯\\\_\(ツ\)\_\/¯

workingjubilee:
Oh, it's because of this bodge. Let's fix this... someday. https://github.com/rust-lang/stdsimd/blob/master/crates/core_simd/src/macros.rs#L54-L73

workingjubilee:
https://github.com/rust-lang/rust/issues/80108 remains an issue on the latest LLVM. I do not know if this is an issue with rustc or with LLVM.

calebzulawski:
Fairly confident it's an LLVM bug, since that's a completely valid bitcast.  I never actually reduced it or submitted it to LLVM though.  I should do that...

workingjubilee:
My rustc can now compile the trivial example given in #80108, so I believe we can take another crack at removing this bodge, but we should rebuild our wasm test suite to guarantee it's actually correct.

PR #55, opened 2021-01-26
WIP implement guards in rem and div unsound cases

miguelraz:
I'm leaving this WIP to address the rest of https://github.com/rust-lang/stdsimd/issues/54
I have to leave for a bit so I'm just opening this for now and will be back briefly to shape it up.

I'm doing this to get started with the contributing workflow and to test out my initial idea of just adding a `panic!` if any of the lanes match the int::MIN.

There's a lot of moving parts and I probably missed a lot of easy stuff, so please, all feedback is welcome.

### DIFF
--- a/crates/core_simd/src/ops.rs
+++ b/crates/core_simd/src/ops.rs
@@ -283,12 +283,19 @@ macro_rules! impl_unsigned_int_ops {
 
                         #[inline]
                         fn div(self, rhs: Self) -> Self::Output {
-                            // TODO there is probably a better way of doing this
-                            if AsRef::<[$scalar]>::as_ref(&rhs)
+                            if rhs.as_slice()
                                 .iter()
                                 .any(|x| *x == 0)
                             {
                                 panic!("attempt to divide by zero");
+                            }
+
+                            // Guards for div(MIN, -1),
+                            // this check only applies to signed ints
+                            if <$scalar>::MIN != 0 && self.as_slice().iter()
+                                    .zip(rhs.as_slice().iter())
+                                    .any(|(x,y)| *x == <$scalar>::MIN && *y == -1 as _) {
+                                panic!("attempt to divide with overflow");
                             }
                             unsafe { crate::intrinsics::simd_div(self, rhs) }
                         }
@@ -303,6 +310,11 @@ macro_rules! impl_unsigned_int_ops {
                         fn div(self, rhs: $scalar) -> Self::Output {
                             if rhs == 0 {
                                 panic!("attempt to divide by zero");
+                            }
+                            if <$scalar>::MIN != 0 &&
+                                self.as_slice().iter().any(|x| *x == <$scalar>::MIN) &&
+                                rhs == -1 as _ {
+                                    panic!("attempt to divide with overflow");
                             }
                             let rhs = Self::splat(rhs);
                             unsafe { crate::intrinsics::simd_div(self, rhs) }
@@ -353,6 +365,14 @@ macro_rules! impl_unsigned_int_ops {
                             {
                                 panic!("attempt to calculate the remainder with a divisor of zero");
                             }
+
+                            // Guards for rem(MIN, -1)
+                            // this branch applies the check only to signed ints
+                            if <$scalar>::MIN != 0 && self.as_slice().iter()
+                                    .zip(rhs.as_slice().iter())
+                                    .any(|(x,y)| *x == <$scalar>::MIN && *y == -1 as _) {
+                                panic!("attempt to calculate the remainder with overflow");
+                            }
                             unsafe { crate::intrinsics::simd_rem(self, rhs) }
                         }
                     }
@@ -366,6 +386,11 @@ macro_rules! impl_unsigned_int_ops {
                         fn rem(self, rhs: $scalar) -> Self::Output {
                             if rhs == 0 {
                                 panic!("attempt to calculate the remainder with a divisor of zero");
+                            }
+                            if <$scalar>::MIN != 0 &&
+                                self.as_slice().iter().any(|x| *x == <$scalar>::MIN) &&
+                                rhs == -1 as _ {
+                                    panic!("attempt to calculate the remainder with overflow");
                             }
                             let rhs = Self::splat(rhs);
                             unsafe { crate::intrinsics::simd_rem(self, rhs) }
--- a/crates/core_simd/tests/ops_impl/int_macros.rs
+++ b/crates/core_simd/tests/ops_impl/int_macros.rs
@@ -229,6 +229,39 @@ macro_rules! int_tests {
             }
 
             #[test]
+            #[should_panic]
+            fn div_min_panics() {
+                let a = from_slice(&vec![$scalar::MIN; 64]);
+                let b = from_slice(&vec![-1; 64]);
+                let _ = a / b;
+            }
+
+            #[test]
+            #[should_panic]
+            fn div_by_all_zeros_panics() {
+                let a = from_slice(&A);
+                let b = from_slice(&vec![0 ; 64]);
+                let _ = a / b;
+            }
+
+            #[test]
+            #[should_panic]
+            fn div_by_one_zero_panics() {
+                let a = from_slice(&A);
+                let mut b = from_slice(&B);
+                b[0] = 0 as _;
+                let _ = a / b;
+            }
+
+            #[test]
+            #[cfg_attr(target_arch = "wasm32", wasm_bindgen_test)]
+            fn div_min_neg_one_no_panic() {
+                let a = from_slice(&A);
+                let b = from_slice(&vec![-1; 64]);
+                let _ = a / b;
+            }
+
+            #[test]
             #[cfg_attr(target_arch = "wasm32", wasm_bindgen_test)]
             fn rem() {
                 let a = from_slice(&A);
@@ -273,6 +306,30 @@ macro_rules! int_tests {
                 let expected = apply_binary_scalar_rhs_lanewise(a, b, core::ops::Rem::rem);
                 a %= b;
                 assert_biteq!(a, expected);
+            }
+
+            #[test]
+            #[cfg_attr(target_arch = "wasm32", wasm_bindgen_test)]
+            fn rem_min_neg_one_no_panic() {
+                let a = from_slice(&A);
+                let b = from_slice(&vec![-1; 64]);
+                let _ = a % b;
+            }
+
+            #[test]
+            #[should_panic]
+            fn rem_min_panic() {
+                let a = from_slice(&vec![$scalar::MIN; 64]);
+                let b = from_slice(&vec![-1 ; 64]);
+                let _ = a % b;
+            }
+
+            #[test]
+            #[should_panic]
+            fn rem_min_zero_panic() {
+                let a = from_slice(&A);
+                let b = from_slice(&vec![0 ; 64]);
+                let _ = a % b;
             }
 
             #[test]


miguelraz:
Thanks @thomcc !
This is my first contribution to the project so I'm still picking things up.
Concretely, is your comment proposing an API for future contributions, or to ammend this PR?

thomcc:
I'm mostly noting it for future reference.

miguelraz:
Awesome. Thanks to @workingjubilee for inspecting the Travis failures on the webassembly builds #51 .

crates/core_simd/src/ops.rs
```diff
@@ -144,8 +144,15 @@ macro_rules! impl_op {
...
                 type Output = Self;
                 fn neg(self) -> Self::Output {
-                    Self::splat(0) - self
-                }
+                // TODO probs a smarter way to do this too
+                    if AsRef::<[$scalar]>::as_ref(&self)
+                        .iter()
+                        .any(|x| *x == <$scalar>::MIN)
+                        {
+                            panic!("neg(MIN) for integers is Undefined Behaviour");
```
workingjubilee, line 152: We might as well match the scalar panic messaging.

>thread 'main' panicked at 'attempt to negate with overflow', src/main.rs:3:24
so "attempt to negate with overflow" would be the preferred messaging here.

crates/core_simd/src/ops.rs
```diff
@@ -144,8 +144,15 @@ macro_rules! impl_op {
...
+                        }
```
workingjubilee, line 153: This should probably work in release mode... can this become a `debug_assert!` or something similar?

workingjubilee (changes requested):

crates/core_simd/src/ops.rs
calebzulawski: Unfortunately `std` is only compiled once so I'm not sure it matters?

thomcc: yeah the debug_asserts in libstd are a lie. there are ways to fake them though (ironically, by exploiting integer overflow behavior), but I don't quite remember. @scottmcm had something that did this at one point.

crates/core_simd/tests/ops_impl/int_macros.rs
```diff
@@ -228,6 +228,28 @@ macro_rules! int_tests {
                 assert_biteq!(a, expected);
             }
 
+            #[test]
+            #[cfg_attr(target_arch = "wasm32", wasm_bindgen_test)]
+            #[should_panic]
```
workingjubilee, line 233: Wasm does not support #[should_panic] tests. https://github.com/rustwasm/wasm-bindgen/issues/2286

I am personally OK with not running these tests on wasm if my collaborators do not object.

crates/core_simd/src/ops.rs
miguelraz: Being consistent makes sense.
However, this is the only value that triggers the behavior - wouldn't it make sense to have the error say that? Or we just assume the user is aware of the edge case?
I'm in favor either way.

crates/core_simd/src/ops.rs
```diff
@@ -144,8 +144,15 @@ macro_rules! impl_op {
             impl<const LANES: usize> core::ops::Neg for crate::$type<LANES> {
                 type Output = Self;
                 fn neg(self) -> Self::Output {
-                    Self::splat(0) - self
-                }
+                // TODO probs a smarter way to do this too
+                    if AsRef::<[$scalar]>::as_ref(&self)
```
calebzulawski, line 148: I think you should be able to use `self.as_slice()` or `self.to_array()` here instead of UFCS.  This is fine but just a bit more verbose.

crates/core_simd/src/ops.rs
scottmcm: There are some magic hacks in the compiler for integer checks to get this to work.  In `core` the usual thing to do (AFAIK) is to just phrase things in terms of `Add::add` instead of `+`, to pick up its behaviour.  That might only work for things that do MIR->LLVM in the context of the user's crate, though...

thomcc:
Given that it's not UB, that SIMD is a performance-oriented API, and that we don't panic for int overflow on adds and such (I think? or we'd have solved this already), it seems worth kicking the overflow checking in debug mode down the road a bit.

crates/core_simd/tests/ops_impl/int_macros.rs
calebzulawski: Oh right, no unwinding in wasm.  Fine with me.

crates/core_simd/src/ops.rs
workingjubilee: I think the phrase used by rustc has an advantage in that it is "non-technical", it uses no special terminology from Rust (like function terminology or referencing a constant in std that in a given user's code might have been shadowed) to express the value from the set and there are several other ops which raise the "attempt to ${op} with overflow" pattern, making it easier to comprehend if it appears beside similar errors that don't involve Neg specifically.

thomcc: Is it actually accurate that it's UB anyway? If we were going to debug_assert instead my assumption would be that it's not.

workingjubilee: Well, we thought that it wasn't, but then we checked and it's apparently not, or at least rustc doesn't care. :^)

crates/core_simd/tests/ops_impl/int_macros.rs
miguelraz: OK - I'll wait for others to chime in.

crates/core_simd/src/ops.rs
miguelraz: Noted and changed.

me22:
> and that we don't panic for int overflow on adds and such

This was my other question here.  If `neg` is implemented in terms of `sub`, then it arguably should have already been doing the right thing (if perhaps with a suboptimal message).

But I also thing that it would be fine for you as a WG to decide that checked-in-debug isn't worth doing for simd, and define all the `+`/`-`/etc to always be wrapping.  (I don't immediately know whether that's the right answer, but I assume you'll make a persuasive case for whichever direction you pick.)

scottmcm:
> and that we don't panic for int overflow on adds and such

This was my other question here. If neg is implemented in terms of sub, then it arguably should have already been doing the right thing (if perhaps with a suboptimal message).

But I also thing that it would be fine for you as a WG to decide that checked-in-debug isn't worth doing for simd, and define all the +/-/etc to always be wrapping. (I don't immediately know whether that's the right answer, but I assume you'll make a persuasive case for whichever direction you pick.)

thomcc:
> But I also thing that it would be fine for you as a WG to decide that checked-in-debug isn't worth doing for simd

Yeah, I'd argue in favor of this, mostly because `checked_add` and stuff would be viable performantly (returning `[Option<i32>; 4]`...). I also think in practice people are more likely to want wrapping behavior on SIMD than normal integers. I think it should be a deliberate decision though.

crates/core_simd/src/ops.rs
calebzulawski: Yeah, I originally mistakenly said it was UB (lumped it in with `div` and `rem` overflow)

workingjubilee:
While we're likely going to settle the issue regarding Neg some other day, that leaves the rest of issue #54, if you wish to continue to take this on. To recap, in the Zulip, I [said](https://rust-lang.zulipchat.com/#narrow/stream/257879-project-portable-simd/topic/MGR.20opens.20a.20PR!.20(issue.20.2354)/near/224138817):
```quote
"apply_binary_scalar_rhs_lanewise" is doing scalar ops, which of course panic.
```

so when that is removed and only the SIMD op is compiled in, running that test instead gets something like:

>process didn't exit successfully: `~/projects/stdsimd/target/debug/deps/ops-6770e1bc9a8f6e58` (signal: 8, SIGFPE: erroneous arithmetic operation)

So the tests should be updated to not include the scalar comparisons, and instead just run the ops without checking the results, since we're only looking for a "successful" panic at runtime, not that the arithmetic results of the SIMD operation would match the arithmetic results of a comparable scalar operation.

miguelraz:
Tests now all pass, but the branch is still requesting a review from @workingjubilee from a resolved convo. 
Re-requesting review.

crates/core_simd/src/ops.rs
```diff
@@ -367,6 +394,10 @@ macro_rules! impl_unsigned_int_ops {
                             if rhs == 0 {
                                 panic!("attempt to calculate the remainder with a divisor of zero");
                             }
+                            if <$scalar>::MIN != 0  && rhs == -1 as _ {
+                                panic!("rem(MIN, -1) is undefined");
```
workingjubilee, line 398: The panic message should use a consistent formatting between the division and remainder cases, and ideally between the scalar and vector cases as well, so either here it should be
```rust
                                 panic!("MIN modulo -1 is undefined"); 
```

or the `fn div` case should use the `div(MIN, -1)` notation, so back there it would be
```rust
                                 panic!("div(MIN, -1) is undefined"); 
```

but in any case, the actual message here is
```rust
thread 'main' panicked at 'attempt to divide with overflow', src/main.rs:3:17
```

But honestly I can be persuaded on any messaging as long as it's consistent.

crates/core_simd/src/ops.rs
```diff
@@ -144,8 +144,8 @@ macro_rules! impl_op {
             impl<const LANES: usize> core::ops::Neg for crate::$type<LANES> {
                 type Output = Self;
                 fn neg(self) -> Self::Output {
-                    Self::splat(0) - self
-                }
+                            Self::splat(0) - self
+                    }
```
workingjubilee, line 148: Hm, this indentation change isn't terribly important, but it's odd. Did you run `cargo fmt`?

crates/core_simd/src/ops.rs
```diff
@@ -290,6 +290,17 @@ macro_rules! impl_unsigned_int_ops {
                             {
                                 panic!("attempt to divide by zero");
                             }
+
+                            // Guards for div(MIN, -1),
+                            // this check only applies to signed ints
+                            if <$scalar>::MIN != 0 {
+                                if AsRef::<[$scalar]>::as_ref(&self).iter()
```
workingjubilee, line 297: Can this be joined with an `&&`?

workingjubilee, line 297: Can this be joined with an `&&`?

crates/core_simd/src/ops.rs
miguelraz: Oooh, did not know that existed.
Thanks, will clean up.

crates/core_simd/src/ops.rs
miguelraz: Yup.

crates/core_simd/src/ops.rs
miguelraz: Sure, I'm going with the first on option you suggested.

crates/core_simd/src/ops.rs
calebzulawski: This can be `as_slice`

crates/core_simd/src/ops.rs
```diff
@@ -353,6 +369,17 @@ macro_rules! impl_unsigned_int_ops {
                             {
                                 panic!("attempt to calculate the remainder with a divisor of zero");
                             }
+
+                            // Guards for rem(MIN, -1)
+                            // this branch applies the check only to signed ints
+                            if <$scalar>::MIN != 0 {
+                                if AsRef::<[$scalar]>::as_ref(&self).iter()
```
calebzulawski: This can be `as_slice`, too

crates/core_simd/src/ops.rs
miguelraz: Oh right - nice catch.

crates/core_simd/src/ops.rs
miguelraz: Fixing!

crates/core_simd/src/ops.rs
```diff
@@ -304,7 +311,12 @@ macro_rules! impl_unsigned_int_ops {
                             if rhs == 0 {
                                 panic!("attempt to divide by zero");
                             }
+                            if <$scalar>::MIN != 0  && rhs == -1 as _ {
+                                panic!("dividing MIN by -1 is undefined");
+
```
calebzulawski, line 316:
```suggestion

```

crates/core_simd/src/ops.rs
```diff
@@ -367,6 +387,10 @@ macro_rules! impl_unsigned_int_ops {
                             if rhs == 0 {
                                 panic!("attempt to calculate the remainder with a divisor of zero");
                             }
+                            if <$scalar>::MIN != 0  && rhs == -1 as _ {
+                                panic!("MIN modulo -1 is undefined");
+
```
calebzulawski, line 392:
```suggestion

```

crates/core_simd/src/ops.rs
```diff
@@ -304,7 +311,12 @@ macro_rules! impl_unsigned_int_ops {
                             if rhs == 0 {
                                 panic!("attempt to divide by zero");
                             }
+                            if <$scalar>::MIN != 0  && rhs == -1 as _ {
+                                panic!("dividing MIN by -1 is undefined");
+
+                            }
                             let rhs = Self::splat(rhs);
+
```
calebzulawski, line 319:
```suggestion

```

crates/core_simd/src/ops.rs
```diff
@@ -367,6 +387,10 @@ macro_rules! impl_unsigned_int_ops {
                             if rhs == 0 {
                                 panic!("attempt to calculate the remainder with a divisor of zero");
                             }
+                            if <$scalar>::MIN != 0  && rhs == -1 as _ {
+                                panic!("MIN modulo -1 is undefined");
```
calebzulawski, line 389:
```suggestion
                                panic!("attempt to calculate the remainder with overflow");
```

crates/core_simd/src/ops.rs
```diff
@@ -353,6 +365,14 @@ macro_rules! impl_unsigned_int_ops {
                             {
                                 panic!("attempt to calculate the remainder with a divisor of zero");
                             }
+
+                            // Guards for rem(MIN, -1)
+                            // this branch applies the check only to signed ints
+                            if <$scalar>::MIN != 0 && self.as_slice().iter()
+                                    .zip(rhs.as_slice().iter())
+                                    .any(|(x,y)| *x == <$scalar>::MIN && *y == -1 as _) {
+                                panic!("MIN modulo -1 is undefined");
```
calebzulawski, line 372:
```suggestion
                                panic!("attempt to calculate the remainder with overflow");
```

crates/core_simd/src/ops.rs
```diff
@@ -283,13 +283,20 @@ macro_rules! impl_unsigned_int_ops {
...
                             {
                                 panic!("attempt to divide by zero");
                             }
+
+                            // Guards for div(MIN, -1),
+                            // this check only applies to signed ints
+                            if <$scalar>::MIN != 0 && self.as_slice().iter()
+                                    .zip(rhs.as_slice().iter())
+                                    .any(|(x,y)| *x == <$scalar>::MIN && *y == -1 as _) {
+                                panic!("dividing MIN by -1 is undefined");
```
calebzulawski, line 298:
```suggestion
                                panic!("attempt to divide with overflow");
```

crates/core_simd/src/ops.rs
```diff
@@ -304,7 +311,12 @@ macro_rules! impl_unsigned_int_ops {
                             if rhs == 0 {
                                 panic!("attempt to divide by zero");
                             }
+                            if <$scalar>::MIN != 0  && rhs == -1 as _ {
+                                panic!("dividing MIN by -1 is undefined");
```
calebzulawski, line 315:
```suggestion
                                panic!("attempt to divide with overflow");
```

workingjubilee (approved):
Much better! My quibbles appear to be resolved and things appear to do what they say they do.

calebzulawski (approved):

miguelraz:
Shoutout to Mario Carneiro for spotting a missing conditional.

crates/core_simd/src/ops.rs
```diff
@@ -367,6 +387,11 @@ macro_rules! impl_unsigned_int_ops {
                             if rhs == 0 {
                                 panic!("attempt to calculate the remainder with a divisor of zero");
                             }
+                            if <$scalar>::MIN != 0 &&
+                                self.as_slice().iter().any(|x| *x == <$scalar>::MIN) &&
+                                rhs == -1 as _ {
+                                    panic!("MIN modulo -1 is undefined");
```
workingjubilee, line 393:
```suggestion
                                    panic!("attempt to calculate the remainder with overflow");
```

crates/core_simd/src/ops.rs
```diff
@@ -304,6 +311,11 @@ macro_rules! impl_unsigned_int_ops {
                             if rhs == 0 {
                                 panic!("attempt to divide by zero");
                             }
+                            if <$scalar>::MIN != 0 &&
+                                self.as_slice().iter().any(|x| *x == <$scalar>::MIN) &&
+                                rhs == -1 as _ {
+                                    panic!("dividing MIN by -1 is undefined");
```
workingjubilee, line 317:
```suggestion
                                    panic!("attempt to divide with overflow");
```

workingjubilee (changes requested):
Somewhere in the review flow these were altered again.

workingjubilee (approved):

PR #61, opened 2021-02-04
Add bitmasks and simplify mask API

calebzulawski:
This should simplify the mask API.

* The `MaskExt` trait was deleted.  Only opaque masks can be created directly with `eq` etc.  Use `From`/`Into` to produce the specific layouts.
* Masks are limited to up to 64 lanes, since the bitmask is backed by a `u64`.  It may be possible to lift this limitation in the future.
* Intrinsics still need to be added to rustc for truncating/sign-extending between bitmasks and full-width masks.


workingjubilee:
Are there any tests that could go here? Literally just "does this still compile when you actually pull it into a usage and compilation context instead of just having it sit 'at rest' in the crate?" is fine, I'm just starting to get a bit suspicious after having noticed some sharp differences when actually trying to use some of the API's I've experimented with.

calebzulawski:
Yeah good point, added some tests

crates/core_simd/src/masks/full_masks.rs
```diff
@@ -0,0 +1,301 @@
...
+        impl<const $lanes: usize> $name<$lanes> {
+            /// Construct a mask by setting all lanes to the given value.
+            pub fn splat(value: bool) -> Self {
+                Self(<$type>::splat(
+                    if value {
+                        -1
+                    } else {
+                        0
+                    }
+                ))
```
workingjubilee, line 32: I did some poking with the version with the u64 mask and it seems to usually optimize to be branchless up there, but I wonder if this version does as consistently? Could be
```suggestion
                Self(<$type>::splat(value as $type).neg())
```
if it doesn't? I don't know for sure.

workingjubilee (commented):
Looks good, just one nit and some idle wondering about perf that came to mind (not backed by any tests or anything so it should be taken as _just_ idle wondering... felt I should voice it out loud, though).

crates/core_simd/tests/masks.rs
```diff
@@ -0,0 +1,67 @@
...
+
+    assert_eq!(ints_returned, ints);
+}
+
+macro_rules! test_mask_api {
+    { $name:ident } => {
+        #[allow(non_snake_case)]
+        mod $name {
+            #[test]
+            #[cfg_attr(target_arch = "wasm32", wasm_bindgen_test)]
```
workingjubilee, line 33: Ooh, almost forgot. Need to bring wasm_bindgen_test into scope in this mod for this wasm test to work, a `use super::*;` inside the mod should suffice.

```suggestion
        mod $name {
            use super::*;

            #[test]
            #[cfg_attr(target_arch = "wasm32", wasm_bindgen_test)]
````

**Review by workingjubilee — CHANGES_REQUESTED** [review]

File: crates/core_simd/src/masks/full_masks.rs [hunk 3]
```diff
@@ -0,0 +1,301 @@
...
+            fn bitand(self, rhs: Self) -> Self {
+                Self(self.0 & rhs.0)
+            }
+        }
+
+        impl<const LANES: usize> core::ops::BitAnd<bool> for $name<LANES> {
+            type Output = Self;
+            #[inline]
+            fn bitand(self, rhs: bool) -> Self {
+                self & Self::splat(rhs)
```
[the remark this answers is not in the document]
calebzulawski: Personally I think this relays intent better.  The splat should just branch once, anyway

crates/core_simd/src/masks/full_masks.rs
calebzulawski: Yeah, I suppose this is more explicitly branchless.

calebzulawski: Ah nevermind, I remember why I wrote it this way, I just didn't want to pass in the scalar type as well.  You can't do `value as $type`, `$type` is the vector type.

crates/core_simd/src/masks/full_masks.rs
```diff
@@ -0,0 +1,301 @@
...
+        pub struct $name<const $lanes: usize>($type);
+
+        impl<const $lanes: usize> $name<$lanes> {
+            /// Construct a mask by setting all lanes to the given value.
+            pub fn splat(value: bool) -> Self {
+                Self(<$type>::splat(
+                    if value {
+                        -1
+                    } else {
+                        0
```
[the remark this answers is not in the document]
calebzulawski: I think it needs to be -1, since -1 doesn't have a specific type.  A const would actually need to be one const per scalar type, I think.

workingjubilee (approved):

workingjubilee:
Then we should be good after a rebase!

PR #62, opened 2021-02-04
Add SIMD shuffles for SimdType{2,4,8,16,32,64}

workingjubilee:
This const generic implementation for certain lane sizes represents
a more limited interface than what LLVM's shufflevector instruction
can handle, as normally the length of U can be different from the
length of T, but offers an interface that it is expected to be able
to expand the capabilities of in the future.

### DIFF
--- a/crates/core_simd/src/intrinsics.rs
+++ b/crates/core_simd/src/intrinsics.rs
@@ -52,4 +52,12 @@ extern "platform-intrinsic" {
     pub(crate) fn simd_le<T, U>(x: T, y: T) -> U;
     pub(crate) fn simd_gt<T, U>(x: T, y: T) -> U;
     pub(crate) fn simd_ge<T, U>(x: T, y: T) -> U;
+
+    // shufflevector
+    pub(crate) fn simd_shuffle2<T, U>(x: T, y: T, idx: [u32; 2]) -> U;
+    pub(crate) fn simd_shuffle4<T, U>(x: T, y: T, idx: [u32; 4]) -> U;
+    pub(crate) fn simd_shuffle8<T, U>(x: T, y: T, idx: [u32; 8]) -> U;
+    pub(crate) fn simd_shuffle16<T, U>(x: T, y: T, idx: [u32; 16]) -> U;
+    pub(crate) fn simd_shuffle32<T, U>(x: T, y: T, idx: [u32; 32]) -> U;
+    pub(crate) fn simd_shuffle64<T, U>(x: T, y: T, idx: [u32; 64]) -> U;
 }
--- a/crates/core_simd/src/permute.rs
+++ b/crates/core_simd/src/permute.rs
@@ -1,0 +1,29 @@
+macro_rules! impl_shuffle_lane {
+    { $name:ident, $fn:ident, $n:literal } => {
+        impl $name<$n> {
+            /// A const SIMD shuffle that takes 2 SIMD vectors and produces another vector, using
+            /// the indices in the const parameter. The first or "self" vector will have its lanes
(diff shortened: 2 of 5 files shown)


workingjubilee:
Check failures seem to spurious, i.e. be due to under-resourced computers rather than actually being compile or test failures.

crates/core_simd/src/intrinsics.rs
```diff
@@ -52,4 +52,12 @@ extern "platform-intrinsic" {
...
     pub(crate) fn simd_gt<T, U>(x: T, y: T) -> U;
     pub(crate) fn simd_ge<T, U>(x: T, y: T) -> U;
+
+    // shufflevector
+    pub(crate) fn simd_shuffle2<T, U>(x: T, y: T, idx: [u32; 2]) -> U;
+    pub(crate) fn simd_shuffle4<T, U>(x: T, y: T, idx: [u32; 4]) -> U;
+    pub(crate) fn simd_shuffle8<T, U>(x: T, y: T, idx: [u32; 8]) -> U;
+    pub(crate) fn simd_shuffle16<T, U>(x: T, y: T, idx: [u32; 16]) -> U;
+    pub(crate) fn simd_shuffle32<T, U>(x: T, y: T, idx: [u32; 32]) -> U;
+    pub(crate) fn simd_shuffle64<T, U>(x: T, y: T, idx: [u32; 64]) -> U;
```
calebzulawski, line 62: So if we extend rustc to have `fn simd_shuffle<T, U, M>(x: T, y: T, idx: M) -> U` we could make this fully generic, right?

calebzulawski (approved):

crates/core_simd/src/permute.rs
```diff
@@ -0,0 +1,28 @@
...
+            /// A const SIMD shuffle that takes 2 SIMD vectors and produces another vector, using
+            /// the indices in the const parameter. The first or "self" vector will have its lanes
+            /// indexed from 0, and the second vector will have its first lane indexed at $n.
+            ///
+            /// Some SIMD shuffle instructions can be quite slow, so avoiding them by loading data
+            /// into the desired patterns in advance is preferred, but shuffles are still faster
+            /// than storing and reloading from memory.
+            #[inline]
+            pub fn shuffle<const IDX: [u32; $n]>(self, second: Self) -> Self {
+                unsafe { crate::intrinsics::$fn(self, second, IDX) }
```
calebzulawski, line 13: How does rustc treat the indices? What happens if one goes out of range?  Is it UB that we somehow need to handle, or does rustc do something to it?  I vaguely recall it turns the index into `idx % lanes` but I'm not sure if that's actually true.

calebzulawski: Perhaps the behavior to the answer to those questions should be documented

calebzulawski (changes requested):

crates/core_simd/src/intrinsics.rs
workingjubilee: Possibly! I am currently poking people in the const generics crew to better understand what is required, so I'm optimistic but not _completely_ certain what's going to happen next.

crates/core_simd/src/permute.rs
workingjubilee: It errors, actually, and there are tests to assure it does so, see https://github.com/rust-lang/rust/pull/76768

calebzulawski (approved):

crates/core_simd/src/permute.rs
workingjubilee: Documented this requirement, since I would not be surprised if people also expect this.

PR #66, opened 2021-02-09
WIP trigonometric vectorized functions

miguelraz:
Very WIP... but a first draft was made!

### DIFF
--- a/crates/core_simd/src/ops.rs
+++ b/crates/core_simd/src/ops.rs
@@ -186,6 +186,30 @@ macro_rules! impl_op {
         }
     };
 
+    // trigonometric unary functions
+    { impl Sin for $type:ident, $scalar:ty } => {
+        impl_ref_ops! {
+            impl<const LANES: usize> core::ops::Sin for crate::$type<LANES> {
+                type Output = Self;
+                fn sin(self) -> Self::Output {
+                    Self::splat(self.sin())
+                }
+            }
+        }
+    };
+
+    { impl Cos for $type:ident, $scalar:ty } => {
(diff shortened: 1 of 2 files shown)


crates/core_simd/src/ops.rs
```diff
@@ -186,6 +186,30 @@ macro_rules! impl_op {
         }
     };
 
+    // trigonometric unary functions
+    { impl Sin for $type:ident, $scalar:ty } => {
+        impl_ref_ops! {
+            impl<const LANES: usize> core::ops::Sin for crate::$type<LANES> {
```
thomcc, line 192: Would we be adding these ops? I'd be personally against that — Not really our place and after a while I've come to believe the lack of core::num traits was the right call.

(If this was just a mistake caused by too much copypasta feel free to ignore)

miguelraz: Yes, I'm still grokking the codebase and am not sure how to add new functionality. I put up this PR to at least get the ball rolling. Recommendations on how to procceed are appreciated.

workingjubilee: For `fn sin` and `fn cos`, you want to have them directly be part of `impl SimdF32` and `impl SimdF64`, see my shuffle PR for something closer to how that can look. https://github.com/rust-lang/stdsimd/pull/62/files

workingjubilee:
Raz does not know when he will be able to get back to this so closing.

miguelraz:
Yup, bit off more than I can chew for the moment. Thanks for following up and others should feel free to take up the draft instead.

Issue #63, opened 2021-02-07
Support non-power-of-two vector lengths.


calebzulawski:
In https://github.com/rust-lang/rust/pull/80652 I disabled non-power-of-two vector lengths to deconflict `stdsimd` and `cg_clif` development.  Power-of-two vectors are typically sufficient, but there was at least one crate using length-3 vectors (https://github.com/yoanlcq/vek/issues/66).  It would be nice to reenable these at some point, which would require support from cranelift.  Perhaps a feature request should be filed with cranelift?

cc @bjorn3

thomcc:
TBH I worry that supporting this will cause way more pain by complicating the semantics compared to the value in supporting the use cases.

1. Concretely, with this we'd have explicit always-padding/uninitialized/undefined fields in some of the simd types.

    This feels dangerous, like it will lead to (likely latent) UB in peoples code the same way working with structs that have padding bytes is a headache in low level code, which I'd generally expect a lot of SIMD code to be.

2. The underlying intrinsics will still read these padding/uninit fields them and mix them into the result in a way that might be problematic in a number of ways.

    Consider comparing 2 vec3s and producing a integer bitmask.
    
    The ideal ideal would probably be the least significant 3 bits set and the rest are zeros. Comparison would produce a 3xN mask with one uninitialized field, but conversion to in integer will produce `0b0000_UZYX`, where U is uninitialized.
    
    That is to say, there's an undef bit, but as (at least) LLVM doesn't track this stuff at the bitwise level, the whole byte is undef... So, maybe `simd_bitmask` is implemented in a way to fix this, but I'd be very concerned, because as the real operations have to read the bit in question, it seems hard to guarantee robustly.

3. I suspect these would be semantically second-class still. That is:

    There are several behaviors that are currently consistent across all vectors which now would not be consistent for these. Example: size being easily computable from the element count, alignment, repr(simd) being soundly transmutable to/from bytes...
    
    Several questions exist whose answer is: "it behaves like the next power of two up" that the actual behavior should probably be that way too. For example: everything about layout again, simd_ffi, asm, ...

4. What if there's an arch in the future that ends up modeling non-POT vectors and then doesn't map well to the "uses next power of two up"? I think this is unlikely, but I think we should shy away from defining behavior if we can avoid it.

[...]

PR #67, opened 2021-02-10
Limit all types to 64 lanes

calebzulawski:
In https://github.com/rust-lang/rust/issues/27731, @oli-obk indicated some concern over allowing monomorphization errors (which are definitely not user-friendly).

This PR extends the `LanesAtMost64` trait to be implemented over _all_ vector types.  This adds a bit of complexity to the implementation.

This solution seems pretty good, but before going with it I want to be sure the following are true:
* Vectors will be useful after stabilization even if the `LanesAtMost64` trait never gets stabilized (much like the now-removed `std::array::LengthAtMost32`).  Considering I didn't need to update the tests, I think this is true.
* We can remove all traces of this trait without breaking the API
* We can change the bound in the future to add more support (such as wider vectors, or non-powers-of-two #63)

### DIFF
--- a/crates/core_simd/src/lib.rs
+++ b/crates/core_simd/src/lib.rs
@@ -15,7 +15,7 @@ mod ops;
 mod round;
 
 mod lanes_at_most_64;
-pub use lanes_at_most_64::*;
+pub use lanes_at_most_64::LanesAtMost64;
 
 mod masks;
 pub use masks::*;
--- a/crates/core_simd/src/lib.rs
+++ b/crates/core_simd/src/lib.rs
@@ -14,6 +14,9 @@ mod intrinsics;
 mod ops;
 mod round;
 
+mod lanes_at_most_64;
+pub use lanes_at_most_64::*;
+
 mod masks;
 pub use masks::*;
 
--- a/crates/core_simd/src/vectors_u8.rs
+++ b/crates/core_simd/src/vectors_u8.rs
@@ -2,7 +2,7 @@
 
 /// A SIMD vector of containing `LANES` `u8` values.
 #[repr(simd)]
-pub struct SimdU8<const LANES: usize>([u8; LANES]);
+pub struct SimdU8<const LANES: usize>([u8; LANES]) where Self: crate::LanesAtMost64;
 
 impl_integer_vector! { SimdU8, u8 }
 
--- a/crates/core_simd/src/fmt.rs
+++ b/crates/core_simd/src/fmt.rs
@@ -33,7 +33,10 @@ macro_rules! impl_fmt_trait {
     { $($type:ident => $(($trait:ident, $format:ident)),*;)* } => {
         $( // repeat type
             $( // repeat trait
-                impl<const LANES: usize> core::fmt::$trait for crate::$type<LANES> {
+                impl<const LANES: usize> core::fmt::$trait for crate::$type<LANES>
+                where
+                    Self: crate::LanesAtMost64,
+                {
                     fn fmt(&self, f: &mut core::fmt::Formatter) -> core::fmt::Result {
                         $format(self.as_ref(), f)
                     }
--- a/crates/core_simd/src/lanes_at_most_64.rs
+++ b/crates/core_simd/src/lanes_at_most_64.rs
@@ -1,0 +1,35 @@
+/// Implemented for bitmask sizes that are supported by the implementation.
+pub trait LanesAtMost64 {}
+
+macro_rules! impl_for {
+    { $name:ident } => {
+        impl LanesAtMost64 for $name<1> {}
+        impl LanesAtMost64 for $name<2> {}
+        impl LanesAtMost64 for $name<4> {}
+        impl LanesAtMost64 for $name<8> {}
+        impl LanesAtMost64 for $name<16> {}
+        impl LanesAtMost64 for $name<32> {}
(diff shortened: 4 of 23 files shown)


oli-obk:
So, once this is merged, the post-monomorphization error in the compiler is unreachable, right? If that is so, we should change it to a `span_bug!`

calebzulawski:
From stdsimd yes, but it's still possible if you use the repr_simd feature directly.

workingjubilee:
@oli-obk People writing code that depends on #[feature(repr_simd)] directly can always directly access the error in question, and there are multiple such nightly-only libraries and users of them, and many of them are quite fragile, being out-of-step with the compiler and dependent on highly unstable and infrequently developed nightly features, which is exactly why we're trying to design a replacement for those.

At the current moment, however, monomorphization errors are fairly common in such code when an error does occur, which is part of why we don't think it's clear that this is a huge diagnostics regression.

crates/core_simd/src/lib.rs
```diff
@@ -14,6 +14,9 @@ mod intrinsics;
 mod ops;
 mod round;
 
+mod lanes_at_most_64;
+pub use lanes_at_most_64::*;
```
KodrAus, line 18: Do we need to actually publicly export this trait?

I guess ideally what we want is to be able to bound the const generic when that becomes possible.

If users do need to use this trait (such as in generic code) then maybe we could come up with a different name for it and seal it so that it's a bit less of a visible hack?

What do you think?

Lokathor: it needs to be visible for generics

name bikeshed begins

KodrAus: If it needs to be visible and possibly stabilized maybe we could possibly call it something like `FixedSimd` and turn it into a useful trait?

```rust
#[rustc_on_unimplemented="fixed SIMD vectors are implemented for types with lane sizes of 2, 4, 8, 16, 32, or 64"]
pub trait FixedSimd {
    type T;
    const N: usize;
}
```

Lokathor: I don't think it needs to live into stable. At least I hope it doesn't.

But for now, on nightly, users do need to see it.

calebzulawski: I agree that it doesn't need to be stable.  You can do quite a bit with generics without touching the trait: https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=46c81ab903379f78b15b03ec98eab68e

KodrAus: If we don't want to stabilize it then any name seems ok to me 🙂 We can at least use the `private::Sealed` trick to prevent anybody else from implementing it.

oli-obk:
> @oli-obk People writing code that depends on `#![feature(repr_simd)]` directly can always directly access the error in question, and there are multiple such nightly-only libraries and users of them, and many of them are quite fragile, being out-of-step with the compiler and dependent on highly unstable and infrequently developed nightly features, which is exactly why we're trying to design a replacement for those.
> 
> At the current moment, however, monomorphization errors are fairly common in such code when an error does occur, which is part of why we don't think it's clear that the original change is a huge diagnostics regression, even if we think it's worth fixing here.

Ah that makes sense. Maybe we should adjust the error message then and point them to the a-little-bit-more-stable stdsimd types once they hit nightly?

crates/core_simd/src/lib.rs
BurntSushi: Do we need to use a glob import here? Seems like it risks exposing something in the public API unintentionally?

Lokathor: No significant risk.

When adding items you don't want to leave the crate you should be marking them as `pub(crate)` anyway.

BurntSushi: Doesn't mean it will happen. Doesn't seem worth the risk to me if you're relying on folks using `pub(crate)`. I certainly don't always do that. The glob import doesn't seem worth it to me.

calebzulawski: For this particular glob, yes it's trivial to turn it into a plain import.  The other globs import quite a bit and we rely on using `pub` and `pub(crate)` to not forget anything.  I think it could go either way.

workingjubilee (commented):
I assume diffing absolutely every single op is required due to the bound leaking through to them? Looks fine if so.

crates/core_simd/src/vectors_u8.rs
```diff
@@ -2,7 +2,7 @@
 
 /// A SIMD vector of containing `LANES` `u8` values.
 #[repr(simd)]
-pub struct SimdU8<const LANES: usize>([u8; LANES]);
+pub struct SimdU8<const LANES: usize>([u8; LANES]) where Self: crate::LanesAtMost64;
```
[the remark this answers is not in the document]
calebzulawski: Whoops, this is a mistake on my part.  I was too excited to publish the PR... So good catch because this is critical 🙃 This bound is completely unnecessary, but this is the one that prevents using the type in a way that causes a monomorpisation error.

calebzulawski:
> I assume diffing absolutely every single op is required due to the bound leaking through to them? Looks fine if so.

Yeah, the ops can only be implemented when the type is valid, so the bounds leak through.

PR #72, opened 2021-02-13
proptest

calebzulawski:
This replaces most tests with proptest, and makes it easier to define tests generically over lane count.  This should provide much broader API coverage and give us more confidence in our implementation.

Unfortunately building with `cross`/`wasm-pack` seems to be relatively slow for some reason, but a regular native test takes just under a minute on my machine.


Lokathor (approved):
To the extent of what I was able to follow while on my phone on the bus, I approve.

Also the only thing worse than all these macros would be not having all these macros.

calebzulawski:
Some of them could potentially be attribute macros in the future.  I'm not sure if that would be better or worse...

calebzulawski:
@workingjubilee should be fixed, needed to update some trait bounds though so might be worth letting the CI run to completion to make sure everything's set

workingjubilee:
Travis PPC64LE builds hung without starting.

workingjubilee:
Repeatedly kicking over Travis eventually got me a pair of PPC builds which actually started and passed normally. Mildly concerning, but moreso about the CI config than this.

calebzulawski:
Weird. They passed before so I'm pretty confident they're good anyway.  Good to merge?

workingjubilee:
Yep, I just had to take a looong moment to make sure the float code was correct because my gnawing sense of paranoia required it. Thank you!

PR #73, opened 2021-02-13
Add to glossary: vectorize, scalar, vector register

workingjubilee:
This primarily adds to the documentation a handful of terms that were unclear to @miguelraz when he was trying to make a contribution, inching the guide towards greater usefulness to the next comer. It also directly explains the purpose of some helper functions for our tests since they might have confusing names for someone who isn't used to SIMD.

### DIFF
--- a/beginners-guide.md
+++ b/beginners-guide.md
@@ -19,9 +19,17 @@ SIMD has a few special vocabulary terms you should know:
 
 * **Vector:** A SIMD value is called a vector. This shouldn't be confused with the `Vec<T>` type. A SIMD vector has a fixed size, known at compile time. All of the elements within the vector are of the same type. This makes vectors *similar to* arrays. One difference is that a vector is generally aligned to its *entire* size (eg: 16 bytes, 32 bytes, etc), not just the size of an individual element. Sometimes vector data is called "packed" data.
 
+* **Vectorize**: An operation that uses SIMD instructions to operate over a vector is often referred to as "vectorized".
+
+* **Autovectorization**: Also known as _implicit vectorization_. This is when a compiler can automatically recognize a situation where scalar instructions may be replaced with SIMD instructions, and use those instead.
+
+* **Scalar:** "Scalar" in mathematical contexts refers to values that can be represented as a single element, mostly numbers like 6, 3.14, or -2. It can also be used to describe "scalar operations" that use strictly scalar values, like addition. This term is mostly used to differentiate between vectorized operations that use SIMD instructions and scalar operations that don't.
+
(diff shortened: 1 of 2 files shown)


beginners-guide.md
```diff
@@ -19,10 +19,18 @@ SIMD has a few special vocabulary terms you should know:
...
 
+* **Vectorize**: An operation that uses SIMD instructions to operate over a vector is often referred to as "vectorized".
+
+* **Autovectorization**: Also known as _implicit vectorization_. This is when a compiler can automatically recognize a situation where scalar instructions may be replaced with SIMD instructions, and use those instead. A compiler lacks the same knowledge about the purpose of the program and shape of the data as the programmer who is using it, and that kind of domain knowledge strongly affects when and where SIMD instructions are appropriate, so this compiler optimization can often fail to trigger, even when it seems obvious to the programmer.
+
+* **Scalar:** "Scalar" in mathematical contexts refers to values that have only one "dimension", like real numbers (which can be represented by a line), but in programming it can also be used to describe "scalar operations" that use strictly scalar values, like addition. This term is mostly used to differentiate between vectorized operations that use SIMD instructions versus scalar operations that don't.
```
calebzulawski, line 26: Some of this phrasing might be confusing since complex numbers are scalar. Might be better to just say a scalar is one of something, vs a vector which is many

calebzulawski (approved):

beginners-guide.md
workingjubilee: Complex numbers don't exist and can't hurt you, they're imaginary. (noted)

beginners-guide.md
```diff
@@ -19,10 +19,18 @@ SIMD has a few special vocabulary terms you should know:
...
 
 * **Bit Widths:** When talking about SIMD, the bit widths used are the bit size of the vectors involved, *not* the individual elements. So "128-bit SIMD" has 128-bit vectors, and that might be `f32x4`, `i32x4`, `i16x8`, or other variations. While 128-bit SIMD is the most common, there's also 64-bit, 256-bit, and even 512-bit on the newest CPUs.
 
+* **Vector Register:** The extra-wide registers that are used for SIMD operations are commonly called vector registers, though you may also see "SIMD registers", vendor names for specific features, or even "floating-point register" as it is common for the same registers to be used with both scalar and vectorized floating-point operations. A processor will execute a SIMD instruction with a vector register, but it may not be executed on a vector register with the same width as the vector that is actually being operated on, because an ISA may not have the same registers available on each processor, and a processor may be able to activate or deactivate registers at runtime. This can affect the performance of SIMD code.
```
miguelraz, line 32: I'd only add something about "They look like `xmm`, `ymm` or `zmm` when you look at assembly, and that's a way to know you are using the vector registers. Good addition.

workingjubilee: There's different names for different architectures, but I think it would be totally reasonable to give the names for NEON, SSE, and AVX{,2,512} registers since those are all the ones in tier1 support.

Lokathor: I think everything starting at "a processor will execute" and after is getting deep into the weeds for a glossary.

workingjubilee: Also true, I think probably this should be discussed somewhere but moved into another section, thinking about it. We want the glossary to be (relatively) laconic so it's quickly reviewed.

miguelraz (approved):

PR #77, opened 2021-02-18
Reorg vector types (nfc)

workingjubilee:
No functional changes.

This follows the reorganization of the code proposed on Zulip by @calebzulawski, I realized that this should ideally be done now, while there aren't any particularly active branches, before I did much of anything myself, because this is the biggest change.

It is slightly unpretty because my primary instrument for moving files and concatenating strings was `bash`, and I figured we could make it prettier whenever we happen to touch the code.


crates/core_simd/src/vector/float.rs
```diff
@@ -23,3 +23,24 @@ pub type f32x16 = SimdF32<16>;
...
 from_transmute_x86! { unsafe f32x8 => __m256 }
 //from_transmute_x86! { unsafe f32x16 => __m512 }
+
+/// A SIMD vector of containing `LANES` `f64` values.
+#[repr(simd)]
+pub struct SimdF64<const LANES: usize>([f64; LANES])
+where
+    Self: crate::LanesAtMost64;
+
+impl_float_vector! { SimdF64, f64, SimdU64 }
```
calebzulawski, line 33: Perhaps the float macro should just be defined here?

calebzulawski: (and int, uint in their appropriate places as well)

calebzulawski (approved):
One general comment, maybe all of the type aliases (`f32x4` etc) should all be moved to their own file?  They tend to clutter the "meaningful" code, imo.

crates/core_simd/src/vector/float.rs
workingjubilee: I moved the type aliases to the bottom of their respective files, the way everything works currently the type aliases are entangled with the code as-is. I will save that for a later refactoring.

calebzulawski (approved):

PR #80, opened 2021-03-06
Add classification functions

### LINKED ISSUE #36: Comparison functions
Lanewise comparisons:
- [x] `lanes_eq`
- [x] `lanes_ne`
- [x] `lanes_lt`
- [x] `lanes_le`
- [x] `lanes_gt`
- [x] `lanes_ge`
- [ ] `is_positive`/`is_sign_positive`
- [ ] `is_negative`/`is_sign_negative`
- [ ] `is_finite`
- [ ] `is_infinite`
- [ ] `is_nan`
- [ ] `is_normal`

Other comparisons:
- [ ] `any_of`/`all_of`
- [ ] `any_nan`/`all_nan`

I'm open to different naming conventions of the lane comparisons, but I think it's best if they're not ambiguous with `PartialEq` etc.  These functions may also be duplicated in traits if multiple mask widths end up being supported.

calebzulawski:
Closes #36.


calebzulawski:
Looks like aarch64 segfaulted during the travis build (not the test). No clue how that happened...

workingjubilee:
That's... odd. This requires rebasing anyways, can you do that and see if it repros?

workingjubilee:
Looks like we do indeed have a particularly gribbly bug!

calebzulawski:
I read somewhere that LLVM segfaults when an assert fails.  When I run this in docker (I don't have access to an actual aarch64 machine) it succeeds, so I thought maybe we're running out of memory and a malloc assert is failing.

Lokathor:
I'm unclear why `codegen-units=1` is in the build flags on some targets but not all targets, and I'm worried about that not being picked up by downstream, when this crate is built into core/std, or is inlined into a user's program.

If we need to reduce codegen-units for this to build without crashing LLVM, and a user of the crate leads the codegen-units at the default quantity of 256 of 16, will *their* build explode when they use this crate?

calebzulawski:
I read in an issue elsewhere that `codegen-units=1` could fix aarch64 issues, but it didn't here.  I just didn't remove it yet.  I'm still not completely certain what is causing the issue.  I can't replicate it locally with `cross`, so I think it's particular to the build agent in some way?

calebzulawski:
I think I've figured out the issue, but I'm not sure what's really happening.  The problem is the `simd_ne` intrinsic which is only used in `is_nan`, `is_normal`, and `is_subnormal`.  If I replace it with a naive implementation everything compiles fine on aarch64.  I think this is probably a bug in LLVM?

workingjubilee:
Following up from discussion: these two examples offer repro of our bug, but only when targeting aarch64, but seemingly only when generating tests, and only in release mode.

```rust
#[test]
fn is_normal() {
    assert!(core_simd::SimdF32::<64>::splat(0.).is_normal().to_array()[0]);
}
```

```rust
#[test]
fn bug() {
    use core_simd::{SimdF32, SimdU32};
    let v = SimdF32::<64>::splat(0.);
    let m = v.lanes_eq(SimdF32::splat(0.0)) & v.lanes_eq(SimdF32::splat(0.0));
    assert!(m.to_array()[0]);
}
```

calebzulawski:
I reduced the maximum lane count from 64 to 32, since the error only occurred with 64-length vectors.

This prevents AVX-512 vectors of `u8`, but this is really only temporary until we can get an LLVM fix (and get that fixed version into rustc)

workingjubilee:
Alright, looks like this is good now with that. Have you filed the LLVM bug somewhere? I think with that posted somewhere (anywhere) this can be merged.

calebzulawski:
There's an LLVM bug filed (though I'm not 100% sure it's the same bug).  I still need to file the rust and stdsimd bugs to track it.

calebzulawski:
Filed #90 and rust-lang/rust#84020.

PR #83, opened 2021-03-08
Add reductions

### LINKED ISSUE #71: any() and all() methods for mask types
It appears that mask types don't currently have `any()` and `all()` methods that `packed_simd` and `simd` had.

Filing this to keep track of adding those methods.

calebzulawski:
Fixes #71.

### DIFF
--- a/crates/core_simd/src/reduction.rs
+++ b/crates/core_simd/src/reduction.rs
@@ -55,11 +55,12 @@ macro_rules! impl_float_reductions {
         where
             Self: crate::LanesAtMost32
         {
+
             /// Produces the sum of the lanes of the vector.
             #[inline]
             pub fn sum(self) -> $scalar {
                 // f32 SIMD sum is inaccurate on i586
-                if cfg!(target_arch = "i586") && core::mem::size_of::<$scalar>() == 4 {
+                if cfg!(all(target_arch = "x86", not(target_feature = "sse2"))) && core::mem::size_of::<$scalar>() == 4 {
                     self.as_slice().iter().sum()
                 } else {
                     unsafe { crate::intrinsics::simd_reduce_add_ordered(self, 0.) }
@@ -70,7 +71,7 @@ macro_rules! impl_float_reductions {
             #[inline]
             pub fn product(self) -> $scalar {
                 // f32 SIMD product is inaccurate on i586
-                if cfg!(target_arch = "i586") && core::mem::size_of::<$scalar>() == 4 {
+                if cfg!(all(target_arch = "x86", not(target_feature = "sse2"))) && core::mem::size_of::<$scalar>() == 4 {
                     self.as_slice().iter().product()
                 } else {
                     unsafe { crate::intrinsics::simd_reduce_mul_ordered(self, 1.) }
--- a/crates/core_simd/src/reduction.rs
+++ b/crates/core_simd/src/reduction.rs
@@ -1,0 +1,142 @@
+macro_rules! impl_integer_reductions {
+    { $name:ident, $scalar:ty } => {
+        impl<const LANES: usize> crate::$name<LANES>
+        where
+            Self: crate::LanesAtMost32
+        {
+            /// Produces the sum of the lanes of the vector, with wrapping addition.
+            #[inline]
+            pub fn wrapping_sum(self) -> $scalar {
+                unsafe { crate::intrinsics::simd_reduce_add_ordered(self, 0) }
+            }
+
+            /// Produces the sum of the lanes of the vector, with wrapping multiplication.
+            #[inline]
+            pub fn wrapping_product(self) -> $scalar {
+                unsafe { crate::intrinsics::simd_reduce_mul_ordered(self, 1) }
+            }
+
+            /// Sequentially performs bitwise "and" between the lanes of the vector.
+            #[inline]
+            pub fn and_lanes(self) -> $scalar {
+                unsafe { crate::intrinsics::simd_reduce_and(self) }
+            }
+
+            /// Sequentially performs bitwise "or" between the lanes of the vector.
+            #[inline]
+            pub fn or_lanes(self) -> $scalar {
+                unsafe { crate::intrinsics::simd_reduce_or(self) }
+            }
+
+            /// Sequentially performs bitwise "xor" between the lanes of the vector.
+            #[inline]
+            pub fn xor_lanes(self) -> $scalar {
+                unsafe { crate::intrinsics::simd_reduce_xor(self) }
+            }
+
+            /// Returns the maximum lane in the vector.
+            #[inline]
+            pub fn max_lane(self) -> $scalar {
+                unsafe { crate::intrinsics::simd_reduce_max(self) }
+            }
+
+            /// Returns the minimum lane in the vector.
+            #[inline]
+            pub fn min_lane(self) -> $scalar {
+                unsafe { crate::intrinsics::simd_reduce_min(self) }
+            }
+        }
+    }
+}
+
+macro_rules! impl_float_reductions {
+    { $name:ident, $scalar:ty } => {
+        impl<const LANES: usize> crate::$name<LANES>
+        where
+            Self: crate::LanesAtMost32
+        {
+            /// Produces the sum of the lanes of the vector.
+            #[inline]
+            pub fn sum(self) -> $scalar {
+                unsafe { crate::intrinsics::simd_reduce_add_ordered(self, 0.) }
+            }
+
+            /// Produces the sum of the lanes of the vector.
+            #[inline]
+            pub fn product(self) -> $scalar {
+                unsafe { crate::intrinsics::simd_reduce_mul_ordered(self, 1.) }
+            }
+
+            /// Returns the maximum lane in the vector.
+            #[inline]
+            pub fn max_lane(self) -> $scalar {
+                unsafe { crate::intrinsics::simd_reduce_max(self) }
+            }
+
+            /// Returns the minimum lane in the vector.
+            #[inline]
+            pub fn min_lane(self) -> $scalar {
+                unsafe { crate::intrinsics::simd_reduce_min(self) }
+            }
+        }
+    }
+}
+
+macro_rules! impl_full_mask_reductions {
+    { $name:ident, $inner:ident } => {
+        impl<const LANES: usize> crate::$name<LANES>
+        where
+            crate::$inner<LANES>: crate::LanesAtMost32
+        {
+            /// Returns true if any lane is set, or false otherwise.
+            #[inline]
+            pub fn any(self) -> bool {
+                unsafe { crate::intrinsics::simd_reduce_any(self.to_int()) }
+            }
+
+            /// Returns true if all lanes are set, or false otherwise.
+            #[inline]
+            pub fn all(self) -> bool {
+                unsafe { crate::intrinsics::simd_reduce_all(self.to_int()) }
+            }
+        }
+    }
+}
+
+macro_rules! impl_opaque_mask_reductions {
+    { $name:ident, $inner:ident, $bits_ty:ident } => {
+        impl<const LANES: usize> $name<LANES>
+        where
+            $bits_ty<LANES>: crate::LanesAtMost32
+        {
+            /// Returns true if any lane is set, or false otherwise.
+            #[inline]
+            pub fn any(self) -> bool {
+                self.0.any()
+            }
+
+            /// Returns true if all lanes are set, or false otherwise.
+            #[inline]
+            pub fn all(self) -> bool {
+                self.0.all()
+            }
+        }
+    }
+}
+
+impl<const LANES: usize> crate::BitMask<LANES>
+where
+    crate::BitMask<LANES>: crate::LanesAtMost32,
+{
+    /// Returns true if any lane is set, or false otherwise.
+    #[inline]
+    pub fn any(self) -> bool {
+        self.0 != 0
+    }
+
+    /// Returns true if all lanes are set, or false otherwise.
+    #[inline]
+    pub fn all(self) -> bool {
+        self.0 == (!0) >> (64 - LANES)
+    }
+}
--- a/crates/core_simd/src/reduction.rs
+++ b/crates/core_simd/src/reduction.rs
@@ -4,45 +4,48 @@ macro_rules! impl_integer_reductions {
         where
             Self: crate::LanesAtMost32
         {
-            /// Produces the sum of the lanes of the vector, with wrapping addition.
+            /// Horizontal wrapping add.  Computes the sum of the lanes of the vector, with wrapping addition.
             #[inline]
             pub fn wrapping_sum(self) -> $scalar {
                 unsafe { crate::intrinsics::simd_reduce_add_ordered(self, 0) }
             }
 
-            /// Produces the sum of the lanes of the vector, with wrapping multiplication.
+            /// Horizontal wrapping multiply.  Computes the product of the lanes of the vector, with wrapping multiplication.
             #[inline]
             pub fn wrapping_product(self) -> $scalar {
                 unsafe { crate::intrinsics::simd_reduce_mul_ordered(self, 1) }
             }
 
-            /// Sequentially performs bitwise "and" between the lanes of the vector.
+            /// Horizontal bitwise "and".  Computes the cumulative bitwise "and" across the lanes of
+            /// the vector.
             #[inline]
-            pub fn and_lanes(self) -> $scalar {
+            pub fn horizontal_and(self) -> $scalar {
                 unsafe { crate::intrinsics::simd_reduce_and(self) }
             }
 
-            /// Sequentially performs bitwise "or" between the lanes of the vector.
+            /// Horizontal bitwise "or".  Computes the cumulative bitwise "or" across the lanes of
+            /// the vector.
             #[inline]
-            pub fn or_lanes(self) -> $scalar {
+            pub fn horizontal_or(self) -> $scalar {
                 unsafe { crate::intrinsics::simd_reduce_or(self) }
             }
 
-            /// Sequentially performs bitwise "xor" between the lanes of the vector.
+            /// Horizontal bitwise "xor".  Computes the cumulative bitwise "xor" across the lanes of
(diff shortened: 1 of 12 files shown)


workingjubilee:
These will be complicated further by https://github.com/rust-lang/rust/issues/83984

calebzulawski:
All ready to go with #80 merged.

Lokathor:
Are these reducing ops horizontal ops? This needs more docs on the methods to explain what's happening.

calebzulawski:
Yes they are, though I thought my documentation was pretty decent.  Do you have any specific examples of what you mean?

Lokathor:
I would include the word "horizontal" so that the difference is more noteable at a glance, and so that search results in rustdoc let you see it sooner (since they show item path and first line of docs but not type).

thomcc:
I'd also suggest that horizontal should be in the name, so that it can be more easily grouped when describing in e.g. whatever guide we end up writing on how to write good/efficient SIMD code.

(E.g. One of the guidelines here would be to, where possible, try to avoid these in batch processing by deferring them until the end, because they're comparatively inefficient on most hardware — often `O(log lanes)` or `O(lanes)`)

Also, it makes it easy to search for if you know what you want.

calebzulawski:
I kind of didn't want to use horizontal because to me it's lingo (intel specific maybe?) and not really descriptive, and I think any documentation using it would need to define what that means anyway.  I just looked at the documentation for arm `vaddv` and it doesn't use the word horizontal, it just defines it as "add across vector".

calebzulawski:
I think reduce is potentially a more universal term (but `sum` might still be more obvious than something like `reduce_add`?)

thomcc:
The term horizontal predates Intel's stuff certainly. It is jargon, but so is stuff like saturating.

Reduce can mean a lot of different things, and seems way more overloaded to me. I could imagine someone expecting reducing a vector to mean reducing the number of lanes, as just an example.

Horizontal, even if it is jargon, is pretty specific to SIMD, is defined in [our guide](https://github.com/rust-lang/stdsimd/blob/master/beginners-guide.md) (briefly) and [the guide that was written for packed_simd](https://rust-lang.github.io/packed_simd/perf-guide/vert-hor-ops.html) (in depth), and is something of a critical concept to understand when writing SIMD — I suspect we have to come up with a name for it and provide a definition somewhere, so I feel like we should use the common terminology for that, even if it's not perfect.

calebzulawski:
Do we maybe want all of the reductions to be named `horizontal_sum`, `horizontal_xor`, `horizontal_min`, etc and leave the docs mostly as-is?

Lokathor:
while that's certainly possible, consider that:
1) there's really no value in under-documenting a function. in `core` even the most obvious math functions have a full explanation of what it does along with a rustdoc example. Having less lines of rustdoc doesn't make your program run faster.
2) you yourself pointed out that "horizontal" is a jargon-y term just a few posts ago. If we take that as true (and honestly I think we can) then simply putting a jargony term on the front of the method names does not do a service to out beginner users.

That said, i got up from a nap like, *after* you posted your most recent reply 14 minutes ago, so i could be kinda half-asleep rigtht now.

calebzulawski:
Updated the function names and docs, I think the new names are more consistent and the docs a little clearer.

workingjubilee:
PPC errors are spurious.
Blocked, functionally, on a decision regarding wasm testing.

crates/core_simd/src/reduction.rs
```diff
@@ -0,0 +1,162 @@
...
+            pub fn sum(self) -> $scalar {
+                // LLVM sum is inaccurate on i586
+                if cfg!(all(target_arch = "x86", not(target_feature = "sse2"))) {
+                    self.as_slice().iter().sum()
+                } else {
+                    unsafe { crate::intrinsics::simd_reduce_add_ordered(self, 0.) }
+                }
+            }
+
+            /// Horizontal multiply.  Computes the sum of the lanes of the vector.
```
Lokathor, line 73: product of the lanes*

workingjubilee:
`Computes` -> `Returns`?
<details><summary>Applicable Patch</summary>

```diff
diff --git a/crates/core_simd/src/reduction.rs b/crates/core_simd/src/reduction.rs
index e728f8a..118d1d8 100644
--- a/crates/core_simd/src/reduction.rs
+++ b/crates/core_simd/src/reduction.rs
@@ -4,46 +4,46 @@ macro_rules! impl_integer_reductions {
         where
             Self: crate::LanesAtMost32
         {
-            /// Horizontal wrapping add.  Computes the sum of the lanes of the vector, with wrapping addition.
+            /// Horizontal wrapping add.  Returns the sum of the lanes of the vector, with wrapping addition.
             #[inline]
             pub fn wrapping_sum(self) -> $scalar {
                 unsafe { crate::intrinsics::simd_reduce_add_ordered(self, 0) }
             }
 
-            /// Horizontal wrapping multiply.  Computes the product of the lanes of the vector, with wrapping multiplication.
+            /// Horizontal wrapping multiply.  Returns the product of the lanes of the vector, with wrapping multiplication.
             #[inline]
             pub fn wrapping_product(self) -> $scalar {
                 unsafe { crate::intrinsics::simd_reduce_mul_ordered(self, 1) }
             }
 
-            /// Horizontal bitwise "and".  Computes the cumulative bitwise "and" across the lanes of
+            /// Horizontal bitwise "and".  Returns the cumulative bitwise "and" across the lanes of
             /// the vector.
             #[inline]
             pub fn horizontal_and(self) -> $scalar {
                 unsafe { crate::intrinsics::simd_reduce_and(self) }
             }
 
-            /// Horizontal bitwise "or".  Computes the cumulative bitwise "or" across the lanes of
+            /// Horizontal bitwise "or".  Returns the cumulative bitwise "or" across the lanes of
             /// the vector.
             #[inline]
             pub fn horizontal_or(self) -> $scalar {
                 unsafe { crate::intrinsics::simd_reduce_or(self) }
             }
 
-            /// Horizontal bitwise "xor".  Computes the cumulative bitwise "xor" across the lanes of
+            /// Horizontal bitwise "xor".  Returns the cumulative bitwise "xor" across the lanes of
             /// the vector.
             #[inline]
             pub fn horizontal_xor(self) -> $scalar {
                 unsafe { crate::intrinsics::simd_reduce_xor(self) }
             }
 
-            /// Horizontal maximum.  Computes the maximum lane in the vector.
+            /// Horizontal maximum.  Returns the maximum lane in the vector.
             #[inline]
             pub fn horizontal_max(self) -> $scalar {
                 unsafe { crate::intrinsics::simd_reduce_max(self) }
             }
 
-            /// Horizontal minimum.  Computes the minimum lane in the vector.
+            /// Horizontal minimum.  Returns the minimum lane in the vector.
             #[inline]
             pub fn horizontal_min(self) -> $scalar {
                 unsafe { crate::intrinsics::simd_reduce_min(self) }
@@ -59,7 +59,7 @@ macro_rules! impl_float_reductions {
             Self: crate::LanesAtMost32
         {
 
-            /// Horizontal add.  Computes the sum of the lanes of the vector.
+            /// Horizontal add.  Returns the sum of the lanes of the vector.
             #[inline]
             pub fn sum(self) -> $scalar {
                 // LLVM sum is inaccurate on i586
@@ -70,7 +70,7 @@ macro_rules! impl_float_reductions {
                 }
             }
 
-            /// Horizontal multiply.  Computes the sum of the lanes of the vector.
+            /// Horizontal multiply.  Returns the sum of the lanes of the vector.
             #[inline]
             pub fn product(self) -> $scalar {
                 // LLVM product is inaccurate on i586
@@ -81,7 +81,7 @@ macro_rules! impl_float_reductions {
                 }
             }
 
-            /// Horizontal maximum.  Computes the maximum lane in the vector.
+            /// Horizontal maximum.  Returns the maximum lane in the vector.
             ///
             /// Returns values based on equality, so a vector containing both `0.` and `-0.` may
             /// return either.  This function will not return `NaN` unless all lanes are `NaN`.
@@ -90,7 +90,7 @@ macro_rules! impl_float_reductions {
                 unsafe { crate::intrinsics::simd_reduce_max(self) }
             }
 
-            /// Horizontal minimum.  Computes the minimum lane in the vector.
+            /// Horizontal minimum.  Returns the minimum lane in the vector.
             ///
             /// Returns values based on equality, so a vector containing both `0.` and `-0.` may
             /// return either.  This function will not return `NaN` unless all lanes are `NaN`.
```
</details>

workingjubilee:
Re: nomenclature, vertical:lanewise::horizontal::vectorwise?

calebzulawski:
Do we think this is good to go?

crates/core_simd/src/reduction.rs
```diff
@@ -0,0 +1,162 @@
+macro_rules! impl_integer_reductions {
+    { $name:ident, $scalar:ty } => {
+        impl<const LANES: usize> crate::$name<LANES>
+        where
+            Self: crate::LanesAtMost32
+        {
+            /// Horizontal wrapping add.  Returns the sum of the lanes of the vector, with wrapping addition.
+            #[inline]
+            pub fn wrapping_sum(self) -> $scalar {
```
Lokathor, line 9: should this say `horizontal_` at the front of the name like the others?

crates/core_simd/src/reduction.rs
```diff
@@ -0,0 +1,162 @@
...
+                unsafe { crate::intrinsics::simd_reduce_add_ordered(self, 0) }
+            }
+
+            /// Horizontal wrapping multiply.  Returns the product of the lanes of the vector, with wrapping multiplication.
+            #[inline]
+            pub fn wrapping_product(self) -> $scalar {
```
Lokathor, line 15: ditto with the above, eg: `horizontal_wrapping_product`

crates/core_simd/src/reduction.rs
```diff
@@ -0,0 +1,162 @@
...
+macro_rules! impl_float_reductions {
+    { $name:ident, $scalar:ty } => {
+        impl<const LANES: usize> crate::$name<LANES>
+        where
+            Self: crate::LanesAtMost32
+        {
+
+            /// Horizontal add.  Returns the sum of the lanes of the vector.
+            #[inline]
+            pub fn sum(self) -> $scalar {
```
Lokathor, line 64: h_

crates/core_simd/src/reduction.rs
```diff
@@ -0,0 +1,162 @@
...
+                if cfg!(all(target_arch = "x86", not(target_feature = "sse2"))) {
+                    self.as_slice().iter().sum()
+                } else {
+                    unsafe { crate::intrinsics::simd_reduce_add_ordered(self, 0.) }
+                }
+            }
+
+            /// Horizontal multiply.  Returns the product of the lanes of the vector.
+            #[inline]
+            pub fn product(self) -> $scalar {
```
Lokathor, line 75: h_

Lokathor (commented):
Other than the possible name changes, looks solid.

crates/core_simd/src/reduction.rs
calebzulawski: I left it out of sum and product because I think the names are more self explanatory, since there's no "vertical" sum (and mirror the ones in `std::iter` a bit, too).

Kind of like how it wouldn't really make sense to make it `horizontal_all`.

Lokathor: well, these types should implement the Sum and Product traits: https://doc.rust-lang.org/std/iter/trait.Sum.html but if you think it won't be confusing we can skip it for now

calebzulawski: Unfortunately we can't implement those traits since they only work on iterators.  They get implemented on the iterator `Item`, not the iterator itself.

More specifically you can already do `vector.to_array().sum()` but it won't be SIMD.

Lokathor: https://docs.rs/wide/0.6.4/wide/struct.f32x4.html#impl-Sum%3CRHS%3E you just do it like this?

calebzulawski: That's for accumulating `Iter<Item=f32x4>` into a single `f32x4`, we should probably provide that too.  But there's no way to do horizontal ops with that trait

workingjubilee: I mean I am still in favor of implementing IntoIterator on vectors, fwiw, precisely for this reason.

Lokathor: Right, Sum and Product traits are about "if i have an iterator of you guys, can you put yourselves together?", which is very different from a horizontal sum.

However, the Sum and Product trait *name their methods* `sum` and `product`, so on the assumption that we'll support those traits on these types eventually, we shouldn't name the horizontal sum  and horizontal product methods some name *other* than the ones already used by standard library traits.

So these methods here should be `horizontal_sum` and `horizontal_product`

crates/core_simd/src/reduction.rs
workingjubilee: 🤪

crates/core_simd/src/reduction.rs
Lokathor: and I understand that there won't be a *literal* namespace clash because one is a method and one is an associated function on a trait but there's a *documentation* and *conceptual* clash for having two things name `sum` or `product` that are doing very different things. It's not normal, it's weird. You wouldn't really name an associated function named `from` and then have it do something *other than* infallibly convert a value from another type into the current type. That would just be a weird thing to do in Rust.

workingjubilee: There was some resistance the first time I pointed it out as an option but the reason I am noting it is precisely because things like horizontal operations are "you stop caring about the vector as a vector and are reducing it to a single value".

calebzulawski: @Lokathor I misunderstood your point, I think.  I agree that it's potentially a little confusing about the direction the accumulation occurs, so I renamed them.

@workingjubilee If you're working with iterators over lanes you're forgoing SIMD, so I think it's appropriate to call `to_array()` to make that explicit

calebzulawski:
Updated all of those fn names

Lokathor (approved):

crates/core_simd/src/reduction.rs
workingjubilee: Mmm, I think there's ways to get around that, but regardless...!

workingjubilee: To elaborate on how that would look:

An iterator over a 16 byte vector (yon typical SSE/Neon register) that iterated one step (doing whatever) and then had `product()` called on it could have the first lane set to 1 and then issue a horizontal mul instruction. Likewise, sum, but the additive identity is 0. This works because we can extract an ExactSizeIterator from a SIMD vector.

calebzulawski: Well, first of all that would require different iterators from Sum and Product, which doesn't really make sense.  But I think that's also just not a very good iterator for normal iteration, compared to an array.

programmerjake: how about having the iterator be something like:
```rust
struct Iter<SimdType: AsElements> {
    value: SimdType,
    range: Range<usize>,
}

impl<SimdType: AsElements> Iterator for Iter<SimdType> {
    type Item = SimdType::Element;
    fn next(&mut self) -> Option<Self::Item> {
        let index = self.range.next()?;
        unsafe { Some(self.value.get_element_unchecked(index)) }
    }
}

impl<SimdType: AsElements> DoubleEndedIterator for Iter<SimdType> {
    fn next_back(&mut self) -> Option<Self::Item> {
        let index = self.range.next_back()?;
        unsafe { Some(self.value.get_element_unchecked(index)) }
    }
}

trait SumProductHelper<I: Iterator<Item=Self>> {
    fn sum(iter: I) -> Self;
    fn product(iter: I) -> Self;
}

impl<I: Iterator<Item=Self>> SumProductHelper<I> for i32 {
    default fn sum(iter: I) -> Self {
        iter.fold(0i32, Add::add)
    }
    default fn product(iter: I) -> Self {
        iter.fold(1i32, Mul::mul)
    }
}

impl<SimdType: AsElements<Element = Self>> SumProductHelper<Iter<SimdType>> for i32 {
    fn sum(iter: Iter<SimdType>) -> Self {
        iter.value.fill_outside_range(iter.range, 0i32).sum()
    }
    fn product(iter: Iter<SimdType>) -> Self {
        iter.value.fill_outside_range(iter.range, 1i32).product()
    }
}

impl Sum for i32 {
    fn sum<I: Iterator<Item=Self>>(iter: I) -> Self {
        SumProductHelper::sum(iter)
    }
}
```

calebzulawski:
@workingjubilee okay with naming etc?

crates/core_simd/src/reduction.rs
```diff
@@ -0,0 +1,162 @@
...
+            #[inline]
+            pub fn horizontal_wrapping_sum(self) -> $scalar {
+                unsafe { crate::intrinsics::simd_reduce_add_ordered(self, 0) }
+            }
+
+            /// Horizontal wrapping multiply.  Returns the product of the lanes of the vector, with wrapping multiplication.
+            #[inline]
+            pub fn horizontal_wrapping_product(self) -> $scalar {
+                unsafe { crate::intrinsics::simd_reduce_mul_ordered(self, 1) }
+            }
```
workingjubilee, line 17: Wrapping is implied throughout the crate by default, is it not?

calebzulawski: For the ops, yes, but I figure it should be explicit for any named functions still

Lokathor: I lean toward "don't say wrapping in the name, it's the default and people should just learn that".

While saying it every time helps the small portion of people who are new to SIMD usage, it becomes a bother to people who are past the initial introduction.

calebzulawski: Yeah I agree this is probably right, there are no non-wrapping functions to distinguish from so it's unnecessary.

crates/core_simd/src/reduction.rs
calebzulawski: I suppose this would work, but it makes the iterator pretty big and wordy and I'm not convinced this is any better than just doing `as_array` or `as_slice`

workingjubilee:
Looks good!

PR #87, opened 2021-04-02
Add saturating abs/neg

workingjubilee:
Simple branchless software implementations for saturating abs/neg, for all the people using them on vectors that might contain i{N}::MIN.


Lokathor:
i don't even know what saturating abs and neg do.

can the docs have like, a words description without needing to puzzle out the code sample?

workingjubilee:
Added.
For every signed integer the lowest value is greater than the highest value, so if you do abs() you just get the minimum number back again in Rust (in other langs you get undef!).

Lokathor (approved):

calebzulawski (approved):

PR #89, opened 2021-04-08
Use neg intrinsics

### LINKED ISSUE #32: Replace bitxor in float Neg impl with an intrinsic
Issue #31 replaces/replaced a non-ieee754-compliant implementation of negate, but while it's more-or-less optimal on some platforms, on ARM it actually constructs a mask and does the XOR, even though there's a relevant instruction it should use.

In general this is all pretty hairy, and ideally rustc would just give us an intrinsic that will negate the float in optimal (but correct) way for the platform. This isn't urgent (the slowdown is minor) but isn't something we should just forget.

workingjubilee:
`simd_neg` was added in https://github.com/rust-lang/rust/pull/83271 so now we get to use it.
This PR fixes #32.

### DIFF
--- a/crates/core_simd/src/intrinsics.rs
+++ b/crates/core_simd/src/intrinsics.rs
@@ -40,6 +40,9 @@ extern "platform-intrinsic" {
     /// fptoui/fptosi/uitofp/sitofp
     pub(crate) fn simd_cast<T, U>(x: T) -> U;
 
+    /// neg/fneg
+    pub(crate) fn simd_neg<T>(x: T) -> T;
+
     // floor
     #[cfg(feature = "std")]
     pub(crate) fn simd_floor<T>(x: T) -> T;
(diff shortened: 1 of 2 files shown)


Lokathor (approved):

calebzulawski:
It looks like this is failing on wasm+simd?
```
Error: failed to deserialize wasm module

Caused by:
    0: failed to parse code section
    1: Unknown 0xfd opcode (at offset 576984)
error: test failed, to rerun pass '--test i16_ops'
Error: Running Wasm tests with wasm-bindgen-test failed
Caused by: failed to execute `cargo test`: exited with exit code: 1
  full command: "cargo" "test" "--release" "--target" "wasm32-unknown-unknown"
The command "wasm-pack test --chrome --headless crates/core_simd --release" exited with 1.
```

`0xfd` are the SIMD opcodes.  Not sure if this is a bug in LLVM or chrome

workingjubilee:
One way to help tell would be to add the ability to test against wasmtime and/or Firefox for reference.

calebzulawski:
I wonder if something broke in chrome, since #83 is failing for the same reason.

PR #95, opened 2021-04-14
Use fabs intrinsic

workingjubilee:
I noticed there's a `simd_fabs` even though there is no `simd_abs`.
Same story as #89.

### DIFF
--- a/crates/core_simd/src/intrinsics.rs
+++ b/crates/core_simd/src/intrinsics.rs
@@ -51,6 +51,9 @@ extern "platform-intrinsic" {
     #[cfg(feature = "std")]
     pub(crate) fn simd_ceil<T>(x: T) -> T;
 
+    /// fabs
+    pub(crate) fn simd_fabs<T>(x: T) -> T;
+
     pub(crate) fn simd_eq<T, U>(x: T, y: T) -> U;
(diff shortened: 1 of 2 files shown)


workingjubilee:
...Why do these tests pass locally but fail on GH?

Lokathor (approved):

Lokathor:
dunno but the code itself looks good.

calebzulawski:
I just tried this on my machine (AMD) and it also passed.

calebzulawski:
Oh wait, I rebased and ran it in release mode and it failed.  Anyway, I found the problem.  Look at what asm is generated for `f32x4::is_infinite()`:
```
 mov     rax, rdi
 xorps   xmm0, xmm0
 movaps  xmmword, ptr, [rdi], xmm0
 ret
```

workingjubilee:
wait
does that just... do nothing?

calebzulawski:
It always returns false :)

workingjubilee:
always finite but never normal around here, apparently. :^)

workingjubilee:
always finite but never normal around here, apparently. :^)

calebzulawski:
Opened rust-lang/rust#84268 and #97.

calebzulawski:
I posted it in the linked issue, but this looks like an LLVM 12 regression

Issue #92, opened 2021-04-11
Failing wasm tests


calebzulawski:
I tracked down the cause of the wasm test failures and it looks like it's due to the comparison ops added to tests in #80.  It looks like wasmparser has been updated to support these opcodes so it's just a matter of the changes propagating to wasm-bindgen-test.  I opened rustwasm/wasm-bindgen#2522 to track this.

I think we can just ignore the failures for now and they should start passing against once wasm-bindgen-test is updated.

cc @workingjubilee

workingjubilee:
I would _like_ to find all the wasm tests that are currently failing on Chrome and disable them individually, but that might be asking too much.

workingjubilee:
This currently hinges on https://github.com/rustwasm/wasm-pack/issues/998
To fix this we can either update our tests to no longer depend on using the `wasm-pack` binary, or wait.
I opened #96 to temporarily omit the tests in the meanwhile.

workingjubilee:
This _actually_ was waiting for `wasm-bindgen-cli` to do an actual release, as it turns out. As Travis is broken, I will move the relevant wasm CI to GHA when I patch the tests back in and force an appropriate minimal `wasm-bindgen` version.

PR #96, opened 2021-04-15
Reduce wasm test suite

workingjubilee:
Fighting CI over #92.

Node is removed from CI because it was apparently never actually being tested: when I ran the test suite locally, it appeared to skip all the tests. Chrome is removed from CI because Chrome Beta drops connections more often and Firefox Nightly allows us to test on the latest Firefox.

Something still fails inside wasm-bindgen, so I omitted the relevant tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,29 +42,28 @@ matrix:
       env: RUSTFLAGS=-Ctarget-feature=+avx512vl
 
     # WebAssembly (wasm-bindgen)
-    - name: "wasm32-unknown-unknown (node, firefox, chrome)"
+    - name: "wasm32-unknown-unknown (firefox)"
       os: linux
       arch: amd64
       addons:
-        firefox: latest
-        chrome: stable
+        firefox: latest-nightly
       install:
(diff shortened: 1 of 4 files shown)


calebzulawski:
Doesn't this have nothing to do with chrome?  It's failing because of #92

workingjubilee:
I guess? I don't fully understand the wasm toolchain yet.

calebzulawski:
This is where the error occurs: https://github.com/rustwasm/wasm-bindgen/blob/master/crates/cli/src/bin/wasm-bindgen-test-runner/main.rs#L69-L70

Lokathor:
*me checking the phone after work* "why am i being emailed about a Gibson novel?"

workingjubilee:
Latest trunk of wasm-bindgen doesn't appear to fix it.

workingjubilee:
I have come around to your point of view, @calebzulawski. I still dropped Chrome because I would rather test on the latest Chrome and Travis + ChromeDriver security features appear to make that hard for us, so I suspect we will have slightly fewer issues with our test suite failing to connect to the headless browser with GeckoDriver.

calebzulawski (approved):

Issue #50, opened 2020-12-08
Bounding const generic bitmasks


workingjubilee:
> You can do this just fine:
> ```
> #[repr(simd)]
> pub struct SimdU8<const LANES: usize>([u8; LANES]);
> ```
> But to make a bitmask you need to do this:
> ```
> #[repr(simd)]
> pub struct SimdBitMask<const LANES: usize>([u8; (LANES + 7) / 8]) where [(); (LANES + 7) / 8]: Sized;
> ```
—@calebzulawski in (https://rust-lang.zulipchat.com/#narrow/stream/257879-project-portable-simd/topic/Mask.20API/near/219272371)

As a reminder, because frankly I have lost track of this in a conversation at least once, a "bitmask" is a mask type that uses 1 bit per vector element (like a bitset) so that a single tightly-packed number defines a mask for an entire vector. Bitmasks can be more efficient, but in order to make an efficient bitmask, you need to know the size of the bitmasks you're supporting ahead of time. This is something we would prefer to not say, given that current architectures that are actually in use on commodity hardware (AVX512) already support 64b bitmasks, and future implementations of RISCV-V also will use bitmasks with pretty interestingly widths†, as I currently understand it.

This problem infects our attempt to design an "opaque", arch-agnostic mask type because we have to impose the same bound at that site as well. That makes it quite unpretty and non-ergonomic, even though we have some clear bounds that you would think we could prove to the compiler pretty effectively.

"Full masks" are comparatively simple... they're the size of a vector's elements, and so, are the size of a vector register. Which is a bit convenient, really. Caleb mentioned It might be nice to say that a mask uses an entire vector register and so is Many bits wide regardless, in terms of layout, and hope the optimizer handles it intelligently.

†  hypothetically, unless I misunderstand [the spec](https://github.com/riscv/riscv-v-spec/blob/master/v-spec.adoc#changes-from-v0-9), it requires a mask register to fit inside VLEN, the size of an individual vector register, but that means it could allow a VLEN-sized mask to mask byte-wide structures over a "vector register group" using that 1 vector register... or 128 elements! @programmerjake am I reading this correctly?

programmerjake:
[...]

Issue #46, opened 2020-11-29
Missing compiler intrinsics


calebzulawski:
I think it would be helpful to collect a list of all compiler intrinsics (the `extern "platform-intrinsics"` ones) that need to be added.

- [ ] `simd_neg`
- [ ] `simd_wrapping_add` etc
- [ ] `simd_overflowing_add` etc

workingjubilee:
On one hand, logically there isn't any reason it should yield the same results, as integers are simpler types. On the other hand, https://github.com/rust-lang/stdsimd/issues/97 still gives me the heebie-jeebies about adding `simd_abs` as an intrinsic.

What would `simd_lt_packed` be equivalent to?

calebzulawski:
I think I previously decided (but never updated this issue) that we don't need an intrinsics for operating on packed bit masks.  I think we only need two intrinsics, one for sign extending a packed bit mask to a full width mask, and one for truncating a full width mask to a bit mask.

calebzulawski:
Actually, looking at the rustc codegen, I think `simd_bitmask` might be the truncation I'm talking about.

PR #99, opened 2021-04-19
Feature/simplify masks

calebzulawski:
Remove explicit bitmasks.

Add `ToBitMask` trait which allows you to convert masks to `u64`.

### DIFF
--- a/crates/core_simd/src/comparisons.rs
+++ b/crates/core_simd/src/comparisons.rs
@@ -1,7 +1,7 @@
 use crate::LanesAtMost32;
 
 macro_rules! implement_mask_ops {
-    { $($vector:ident => $mask:ident ($inner_mask_ty:ident, $inner_ty:ident),)* } => {
+    { $($vector:ident => $mask:ident ($inner_ty:ident),)* } => {
         $(
             impl<const LANES: usize> crate::$vector<LANES>
             where
@@ -12,8 +12,7 @@ macro_rules! implement_mask_ops {
                 #[inline]
                 pub fn lanes_eq(self, other: Self) -> crate::$mask<LANES> {
                     unsafe {
-                        crate::$inner_mask_ty::from_int_unchecked(crate::intrinsics::simd_eq(self, other))
-                            .into()
+                        crate::$mask::from_int_unchecked(crate::intrinsics::simd_eq(self, other))
                     }
                 }
 
@@ -21,8 +20,7 @@ macro_rules! implement_mask_ops {
                 #[inline]
                 pub fn lanes_ne(self, other: Self) -> crate::$mask<LANES> {
                     unsafe {
-                        crate::$inner_mask_ty::from_int_unchecked(crate::intrinsics::simd_ne(self, other))
-                            .into()
+                        crate::$mask::from_int_unchecked(crate::intrinsics::simd_ne(self, other))
                     }
                 }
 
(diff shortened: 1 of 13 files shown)


calebzulawski:
I'm not sure. I'm still thinking about that.  Unfortunately there's basically no way to represent that in the rust type system as far as generics go.  Related to #50

programmerjake:
> I'm not sure. I'm still thinking about that. Unfortunately there's basically no way to represent that in the rust type system as far as generics go. Related to #50

An idea for fixing that:
https://github.com/rust-lang/stdsimd/issues/50#issuecomment-822634071

workingjubilee:
@programmerjake We will absolutely successfully work out a version that does not break on 2048-bit vectors with byte lanes before we stabilize anything (or else over my dead body, I suppose). You can expect to see this API change a lot, please do not read too much into any single version. Right now, we're trying things out a lot and seeing how things work.

"How to handle masks on platforms" is definitely going to be an Unresolved Question in our RFC.

calebzulawski:
I think there are two separate things here that we need to consider.

One is that masks of any (reasonable) vector length should be supported, regardless of their implementation.  That could be solved by using a trait like that (it may be appropriate as part of `LanesAtMost32` or whatever we end up calling it).

But I think there is a separate problem of explicitly exposing a mask as an integer bitmask.  This is the SSE movemask example, where you use your bitmask as an index to a lookup table or similar.  It's possible to have a 2048-lane vector, but I don't think it's useful to expose an entire 256-byte "integer".  Or, hypothetically, if we support Mask8<65536> in the future, I don't think there's any (portable!) way of exposing that.  On AVX-512 is that 2048 k registers placed into an array in memory?  I can't think of many (portable!) reasons you'd need to access that as a [u8].  This is what the `ToBitMask` trait is for, converting bit masks with up to 64 lanes into `u64`s, independently of the supported number of mask lanes.

calebzulawski:
Anyone know why the MIPS masks seem to be reversed when packed to an integer?  Somehow an LLVM bug or is there something peculiar about MIPS?

programmerjake:
IDK, but maybe it's related to Ada bit array ordering?

workingjubilee:
...Ada bit array?

programmerjake:
> ...Ada bit array?

They describe it some here:
https://en.wikibooks.org/wiki/Ada_Programming/Types/array#With_aliased_elements

calebzulawski:
I think that leaves me with more questions than I started with.  What do Ada arrays have to do with this?

programmerjake:
Well, when discussing which order to have bitmasks in SimpleV, lxo (one of the GCC Ada compiler engineers who's helping out with Libre-SOC) mentioned that Ada uses MSB to LSB ordering for bit arrays -- I figured that might be related since it uses the same order as what you apparently noticed in MIPS.

calebzulawski:
Hmm.  I'd hope that's not related to the problem.  I would expect LLVM to maintain consistent bit field ordering across architectures.

workingjubilee:
Can you rebase this?

calebzulawski:
Okay.  "Simplify" might be a bit of a misnomer now (but the API is definitely simplified!)

I used a `Mask` helper trait to mostly fix #50.  I didn't use `const_evaluatable_unchecked` to implement the trait over all masks, but it should be possible in the future, effectively solving that problem.

Bit masks are now implemented as arrays of `u8` and can be exposed to the users of stdsimd that way, too.  If you want an integer you'll have to do the conversion yourself (not that big of a deal).  Due to limitations in the `simd_bitmask` and `simd_select_bitmask` intrinsic, `Mask` has an extra `IntBitMask` associated type, which is purely used internally.  We should be able to remove that after improving the intrinsics to accept arrays of `u8` as bitmasks as well.

workingjubilee:
This wound up not simplifying things as much as intended internally but it looks much better to me as an external API, really, which is what counts here. Can you add "modify Rust SIMD intrinsics to accept u8-or-something bitmasks" to #46 with a note about all the ones that need to get touched?

PR #103, opened 2021-04-22
mask select

calebzulawski:
Ready to go pending #99

### DIFF
--- a/crates/core_simd/src/select.rs
+++ b/crates/core_simd/src/select.rs
@@ -1,0 +1,52 @@
+mod sealed {
+pub trait Sealed {}
+}
+use sealed::Sealed;
+
+/// Supporting trait for vector `select` function
+pub trait Select<Mask>: Sealed {}
+
+macro_rules! impl_select {
+    {
+        $mask:ident ($bits_ty:ident): $($type:ident),*
+    } => {
+        $(
+        impl<const LANES: usize> Sealed for crate::$type<LANES> where Self: crate::LanesAtMost32 {}
+        impl<const LANES: usize> Select<crate::$mask<LANES>> for crate::$type<LANES>
+        where
+            crate::$mask<LANES>: crate::Mask,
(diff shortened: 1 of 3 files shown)


crates/core_simd/src/select.rs
```diff
@@ -0,0 +1,52 @@
...
+/// Supporting trait for vector `select` function
+pub trait Select<Mask>: Sealed {}
+
+macro_rules! impl_select {
+    {
+        $mask:ident ($bits_ty:ident): $($type:ident),*
+    } => {
+        $(
+        impl<const LANES: usize> Sealed for crate::$type<LANES> where Self: crate::LanesAtMost32 {}
+        impl<const LANES: usize> Select<crate::$mask<LANES>> for crate::$type<LANES>
```
workingjubilee, line 15: So this looks pretty good and I want to merge this but, I have to ask, is there a reason we need to seal this trait? I am perhaps missing something / asking an obvious question, I suspect.

calebzulawski: Since the trait is all that's gating the types accepted by `select`, allowing a user to implement the trait (even on nightly) makes it pretty easy to get either compiler errors or maybe even do something broken that we haven't tested.  Also, it may be desirable to stabilize the trait at some point (so you can use select in generics more easily) and then I think it should definitely be sealed.

`FloatToInt` is a similar trait already in `std`:
https://doc.rust-lang.org/nightly/std/convert/trait.FloatToInt.html

workingjubilee: _thinking_ But if they implement the trait, it just does whatever they say it does, doesn't it?
Not that I'm firmly against it being sealed or anything, I just wanted to check if there was something I was missing.

workingjubilee: newayz having slept on it sealing this is fine, and implementing it as a trait is fine, but I think we should consider developing a more generalized approach to "this logic makes most sense on a trait" stuff, especially as it may help for organizing our API in general.

PR #111, opened 2021-04-30
(WIP): Add rotate_left and rotate_right

RDambrosio016:
Adds `rotate_left` and `rotate_right` per #98, However, it is currently broken because it relies on accessing `N` inside of the const fn, which rustc does not currently allow. Therefore, some compiler trickery will be required to make this work, which is why this is a draft.


calebzulawski:
Closing this because we support rotation now.  It just needed a little more help from the API.

PR #112, opened 2021-04-30
Feat: add shift_left and shift_right

RDambrosio016:
Adds `shift_left` and `shift_right` per #98, However it is broken in the same manner as #111 therefore it is also a draft.


sammysheep:
I saw that rotation methods were added a while back but for the life of me (user error?) couldn't find a whole vector shift in the API as implemented here. Obviously the within each element shift exists and it's easy to use.

Also curious, does the compiler optimize this well?

Issue #76, opened 2021-02-17
extern "platform-intrinsics" float functions often call libm


workingjubilee:
I tried this code [Godbolt](https://rust.godbolt.org/z/q3K3PM):
<details><summary>Rust Code</summary>
<p>

```rust
#![no_std]
#![allow(non_camel_case_types)]
#![feature(repr_simd, platform_intrinsics)]

#[repr(simd)]
#[derive(Debug)]
pub struct f32x4(f32, f32, f32, f32);

extern "platform-intrinsic" {
    fn simd_fsqrt<T>(x: T) -> T;
    fn simd_fabs<T>(x: T) -> T;
    fn simd_fsin<T>(x: T) -> T;
    fn simd_fcos<T>(x: T) -> T;
    fn simd_ceil<T>(x: T) -> T;
    fn simd_fexp<T>(x: T) -> T;
    fn simd_fexp2<T>(x: T) -> T;
    fn simd_floor<T>(x: T) -> T;
    fn simd_fma<T>(x: T, y: T, z: T) -> T;
    fn simd_flog<T>(x: T) -> T;
    fn simd_flog10<T>(x: T) -> T;
    fn simd_flog2<T>(x: T) -> T;
    fn simd_fpow<T>(x: T, y: T) -> T;
    fn simd_fpowi<T>(x: T, y: i32) -> T;
}

impl f32x4 {
    pub fn powi(self, exp: i32) -> Self {
        unsafe { simd_fpowi(self, exp) }
    }
    pub fn powf(self, exp: Self) -> Self {
        unsafe { simd_fpow(self, exp) }
    }
    pub fn flog2(self) -> Self {
        unsafe { simd_flog2(self) }
    }
    pub fn flog10(self) -> Self {
        unsafe { simd_flog10(self) }
    }
    pub fn flog(self) -> Self {
        unsafe { simd_flog(self) }
    }
    pub fn mul_add(self, y: Self, z: Self) -> Self {
        unsafe { simd_fma(self, y, z) }
    }
    pub fn floor(self) -> Self {
        unsafe { simd_floor(self) }
    }
    pub fn fexp(self) -> Self {
        unsafe { simd_fexp(self) }
    }
    pub fn fexp2(self) -> Self {
        unsafe { simd_fexp2(self) }
    }
    pub fn ceil(self) -> Self {
        unsafe { simd_ceil(self) }
    }
    pub fn sqrt(self) -> Self {
        unsafe { simd_fsqrt(self) }
[...]

Issue #109, opened 2021-04-25
Potentially share Vector Math library with the Kazan GPU driver (related to Libre-SOC and SimpleV)


programmerjake:
We're going to need a Vector Math library for the Kazan in order to implement all the functions that Vulkan requires (basically all the sin, cos, tan, atan, sinpi, cospi, exp, log2, pow, etc. functions), if we work together on the same library we could also use it for Rust's `std::simd`, potentially saving a bit of work. Both Kazan and rustc share backends (LLVM and cranelift) and would like to support having vector math functions inlined into calling code, giving a potentially substantial performance boost.

There is potentially funding available for this (see matching [bug on Libre-SOC's bugtracker](https://bugs.libre-soc.org/show_bug.cgi?id=632)).

programmerjake:
Funding available! Provisionally assigning EUR 2000, to be adjusted as needed.

calebzulawski:
I'm unfamiliar with Kazan, but if both projects use LLVM perhaps most of the implementation work here should be handled in LLVM?

It seems to me like the most universal approach would be to provide a library as part of the LLVM project that implements these functions, similar to (or maybe even part of?) compiler-rt.

programmerjake:
We'd also want it to work with cranelift, so having it be part of LLVM may not be best.

thomcc:
For clarity: It would still be written in Rust?

programmerjake:
@thomcc That's the plan...Rust seems like the best language for `core::simd`.

thomcc:
Right, I just wasn't sure it met your other requirements.

programmerjake:
Kazan is also written in Rust.

there can either be a vector abstraction layer that compiles out to nothing in `std::simd` and compiles to code for generating IR in Kazan, or Kazan can just use the spir-v backend for rustc, or save the generated llvm & cranelift ir.

programmerjake:
A rough sketch of how I think we could have the vector math library's API that the vector math functions would be implemented using:
```rust
pub struct F16(u16); // to be replaced by built-in type when Rust gains support

/// reference used to build IR for Kazan; an empty type for `core::simd`
pub trait Context: Copy {
    // todo: add missing type conversions such as i32 -> f64 and i64 -> u8
    // scalar types
    type Bool: Bool + Make<Self, Prim = bool> + Select<Self::Bool>;
    type U8: UInt<Self::U32> + Compare<Bool = Self::Bool> + Make<Self, Prim = u8>;
[...]

PR #120, opened 2021-05-18
add simd_fsqrt intrinsic

miguelraz:
Ref #76 
cc @workingjubilee

### DIFF
--- a/crates/core_simd/src/vector/float.rs
+++ b/crates/core_simd/src/vector/float.rs
@@ -35,6 +35,14 @@ macro_rules! impl_float_vector {
             pub fn abs(self) -> Self {
                 unsafe { crate::intrinsics::simd_fabs(self) }
             }
+
+            /// Produces a vector where every lane has the square root value
+            /// of the equivalently-indexed lane in `self`
+            #[inline]
+            #[cfg(feature = "std")]
+            pub fn sqrt(self) -> Self {
+                unsafe { crate::intrinsics::simd_fsqrt(self) }
+            }
         }
 
         impl<const LANES: usize> $name<LANES>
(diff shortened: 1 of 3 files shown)


crates/core_simd/src/vector/float.rs
```diff
@@ -35,6 +35,13 @@ macro_rules! impl_float_vector {
             pub fn abs(self) -> Self {
                 unsafe { crate::intrinsics::simd_fabs(self) }
             }
+
+            /// Produces a vector where every lane has the square root value
+            /// of the equivalently-indexed lane in `self`
+            #[inline]
+            pub fn sqrt(self) -> Self {
```
calebzulawski, line 42: I think this should possible go in another file instead of the main float.rs file.  Also, it needs to be gated on the std cargo feature.

calebzulawski (changes requested):

crates/core_simd/src/vector/float.rs
miguelraz: Got it - does `vector/fops.rs` sound good?

calebzulawski: Hm, I browsed the files and maybe this file is the right place to put it (for now at least).  But it should still be gated on the std feature.

Lokathor: So will we make sqrt available in core later on? or is there some sort of plan there.

calebzulawski: Yeah, I think it will using the vector libm (#109), but will require LLVM changes too.

miguelraz:
@calebzulawski I pusehd the `std` gating feature changes. Does that resolve your review?

calebzulawski (approved):

Issue #110, opened 2021-04-30
Port the examples from packed_simd


workingjubilee:
`packed_simd` hosts various examples in its repo, featuring classics like
- [ ] Aobench
- [ ] Dot Product
- [ ] Fannkuch Redux
- [ ] Mandelbrot
- [x] Matrix Inverse
- [x] N-Body
- [ ] Options Pricing
- [ ] Slice Sum
- [x] Spectral Norm
- [ ] Stencil
- [ ] Triangle Xform

We've already had users come up with code (in spite of not even being published to crates.io), so obviously we're doing *something* right in terms of making a decently intuitive API, but we should port some of these to the current API to provide more practical working examples, and also tacitly a bench suite, in this repo or another. That or we should figure out why we're _not_ doing so.

calebzulawski:
Great idea, also a good way to ensure our API is actually usable.

miguelraz:
Should every example be packaged in a mini-crate? or would a single runnable for Rust Playground be a better idea?

calebzulawski:
Cargo supports examples in the `examples/` directory, which can be automatically built and tested: https://doc.rust-lang.org/cargo/guide/project-layout.html)

I would put them each in their own file in there.

PR #122, opened 2021-05-26
nbody example

miguelraz:
WIP #110 , nbody example

I can run this with `cargo run --examples` 🎉 .

### DIFF
--- a/crates/core_simd/examples/nbody.rs
+++ b/crates/core_simd/examples/nbody.rs
@@ -1,4 +1,5 @@
-use stdsimd::simd::*;
+#![feature(platform_intrinsics, repr_simd)]
+use core_simd::*;
 
 use std::f64::consts::PI;
 const SOLAR_MASS: f64 = 4.0 * PI * PI;
--- a/crates/core_simd/examples/nbody.rs
+++ b/crates/core_simd/examples/nbody.rs
@@ -1,0 +1,188 @@
+use stdsimd::simd::*;
+
+use std::f64::consts::PI;
+const SOLAR_MASS: f64 = 4.0 * PI * PI;
+const DAYS_PER_YEAR: f64 = 365.24;
+
+pub struct Body {
+    pub x: f64x4,
+    pub v: f64x4,
+    pub mass: f64,
+}
+const N_BODIES: usize = 5;
+#[allow(clippy::unreadable_literal)]
+const BODIES: [Body; N_BODIES] = [
+    // sun:
+    Body {
+        x: f64x4::new(0., 0., 0., 0.),
+        v: f64x4::new(0., 0., 0., 0.),
+        mass: SOLAR_MASS,
+    },
+    // jupiter:
+    Body {
+        x: f64x4::new(
+            4.84143144246472090e+00,
+            -1.16032004402742839e+00,
+            -1.03622044471123109e-01,
+            0.,
+        ),
+        v: f64x4::new(
+            1.66007664274403694e-03 * DAYS_PER_YEAR,
+            7.69901118419740425e-03 * DAYS_PER_YEAR,
+            -6.90460016972063023e-05 * DAYS_PER_YEAR,
+            0.,
+        ),
+        mass: 9.54791938424326609e-04 * SOLAR_MASS,
+    },
+    // saturn:
+    Body {
+        x: f64x4::new(
+            8.34336671824457987e+00,
+            4.12479856412430479e+00,
+            -4.03523417114321381e-01,
+            0.,
+        ),
+        v: f64x4::new(
+            -2.76742510726862411e-03 * DAYS_PER_YEAR,
+            4.99852801234917238e-03 * DAYS_PER_YEAR,
+            2.30417297573763929e-05 * DAYS_PER_YEAR,
+            0.,
+        ),
+        mass: 2.85885980666130812e-04 * SOLAR_MASS,
+    },
+    // uranus:
+    Body {
+        x: f64x4::new(
+            1.28943695621391310e+01,
+            -1.51111514016986312e+01,
+            -2.23307578892655734e-01,
+            0.,
+        ),
+        v: f64x4::new(
+            2.96460137564761618e-03 * DAYS_PER_YEAR,
+            2.37847173959480950e-03 * DAYS_PER_YEAR,
+            -2.96589568540237556e-05 * DAYS_PER_YEAR,
+            0.,
+        ),
+        mass: 4.36624404335156298e-05 * SOLAR_MASS,
+    },
+    // neptune:
+    Body {
+        x: f64x4::new(
+            1.53796971148509165e+01,
+            -2.59193146099879641e+01,
+            1.79258772950371181e-01,
+            0.,
+        ),
+        v: f64x4::new(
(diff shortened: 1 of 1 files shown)


crates/core_simd/examples/nbody.rs
```diff
@@ -0,0 +1,192 @@
...
+    }
+}
+fn main() {
+    //let n: usize = std::env::args()
+    //.nth(1)
+    //.expect("need one arg")
+    //.parse()
+    //.expect("argument should be a usize");
+    //run(&mut std::io::stdout(), n, alg);
+    println!("{:?}", run_k::<10>(10, 10));
```
workingjubilee, line 191: This turbofish causes a disagreement between this constant value and the previous definition of <K> as `Fn(&mut [Body; N_BODIES], f64)`, as it is a constant value and not a function. You may have meant to use just `run`?

It would also probably be fine if your function accepted a mutable slice, `&mut [Body]`, currently it's expecting specifically `Fn<(&mut [Body; 5], f64)>`.

crates/core_simd/examples/nbody.rs
```diff
@@ -0,0 +1,192 @@
...
+#[derive(Debug, Clone, Copy)]
+pub struct Body {
+    pub x: f64x4,
+    pub v: f64x4,
+    pub mass: f64,
+}
+
+// translation up to here
+const N_BODIES: usize = 5;
+#[allow(clippy::unreadable_literal)]
```
workingjubilee, line 17: We don't run clippy in CI, so you're safe here. :^)

crates/core_simd/examples/nbody.rs
```diff
@@ -0,0 +1,192 @@
+#![feature(platform_intrinsics, repr_simd)]
```
workingjubilee, line 1:
```suggestion
```
I believe these can be removed.

crates/core_simd/examples/nbody.rs
```diff
@@ -0,0 +1,192 @@
...
+    for _ in 0..n {
+        k(&mut bodies, 0.01);
+    }
+    let energy_after = energy(&bodies);
+
+    (energy_before, energy_after)
+}
+
+pub fn run(n: usize) -> (f64, f64) {
+    run_k(n, advance)
```
workingjubilee, line 167: The `advance` function suddenly leaping into context is slightly surprising here. By its name, I somewhat expected it to be a variable. Partially it's because there's both `run` and `run_k`. Collapsing `run` into `run_k` as a single function would probably be fine, since this is just a pass-through.

workingjubilee (commented):
We're getting there.

crates/core_simd/examples/nbody.rs
miguelraz: Oh right - `rustc` called me out on this too. I'll change it to `run(10)` and that should do it.

crates/core_simd/examples/nbody.rs
miguelraz: Gotcha. Got a bit punk'd with the globals, but this should be fine now.

crates/core_simd/examples/nbody.rs
```diff
@@ -0,0 +1,190 @@
...
+    let energy_before = energy(&bodies);
+    for _ in 0..n {
+        advance(&mut bodies, 0.01);
+    }
+    let energy_after = energy(&bodies);
+
+    (energy_before, energy_after)
+}
+
+fn approx_eq_f32(a: f32, b: f32) -> bool {
```
Lokathor, line 175: Note that when i used this as an example on zulip... this particular threshold is just "the smallest value where the intel rsqrt intrinsic seemed to stop complaining", so maybe we could change the threshold, but since this is just an example i think it's probably fine to pick a threshold out of thin air.

miguelraz: I demoted the answers to f32 to make it work, so I don't mind fudging it again for the purposes of demonstration.

miguelraz:
@workingjubilee this seems ready to merge.
Slight tweaks can be made in future PRs to address small fixes.

programmerjake:
maybe you could add `println!` calls to `run`, print the planets' location and the current time after each time-step, and add a call to `run` to `main`, allowing `cargo run` to do something interesting.

miguelraz:
@programmerjake that's a fair addition - I don't think it's a merge blocker for the time being though.

crates/core_simd/examples/nbody.rs
```diff
@@ -0,0 +1,186 @@
...
+    while i < N {
+        let d2s = f64x2::from_array([
+            (r[i] * r[i]).horizontal_sum(),
+            (r[i + 1] * r[i + 1]).horizontal_sum(),
+        ]);
+        let dmags = f64x2::splat(dt) / (d2s * d2s.sqrt());
+        mag[i] = dmags[0];
+        mag[i + 1] = dmags[1];
+        i += 2;
+    }
```
workingjubilee, line 139: This can use
```rust
for i in 0..N.step_by(2) {
```
instead of the mutation of `i`, here.

crates/core_simd/examples/nbody.rs
```diff
@@ -0,0 +1,186 @@
...
+
+    i = 0;
+    for j in 0..N_BODIES {
+        for k in j + 1..N_BODIES {
+            let f = r[i] * mag[i];
+            bodies[j].v -= f * bodies[k].mass;
+            bodies[k].v += f * bodies[j].mass;
+            i += 1
+        }
+    }
```
workingjubilee, line 149: Ideally, for cases like this, we would declare the value of `i` in a formulaic fashion here instead of mutating it, but using `let mut i = 0;` here would be fine, to confine the mutation to this passage of code instead of the... very "C" idiom of reusing variables from earlier.

crates/core_simd/examples/nbody.rs
```diff
@@ -0,0 +1,186 @@
...
+    let energy_after = energy(&bodies);
+
+    (energy_before, energy_after)
+}
+
+// Good enough for demonstration purposes, not going for strictness here.
+fn approx_eq_f64(a: f64, b: f64) -> bool {
+    (a - b).abs() < 0.00001
+}
```
workingjubilee, line 170: This can be moved inside `mod tests` so that it doesn't emit unused code warnings, as it's unused outside of it. This should also make the `use super::*;` unnecessary.

workingjubilee (commented):
Cleared some headspace to sit down and look at the implementation more closely.
I have a few relatively minor nits and then I can indeed merge this.

crates/core_simd/examples/nbody.rs
miguelraz: Rust says it wants `(0..N).step_by(2)` but yes, nice idiom!

crates/core_simd/examples/nbody.rs
miguelraz: Oh, that is a nice way of putting it. I will try to enclose those mutations to smaller scopes. Good catch.

crates/core_simd/examples/nbody.rs
miguelraz: Good point - will move.

workingjubilee:
Excellent!

PR #123, opened 2021-05-26
add simd_rsqrt

miguelraz:


crates/core_simd/src/vector/float.rs
```diff
@@ -43,6 +43,15 @@ macro_rules! impl_float_vector {
...
                 unsafe { crate::intrinsics::simd_fsqrt(self) }
             }
+
+            /// Produces a vector where every lane has the reciprocal square root value
+            /// of the equivalently-indexed lane in `self`. Very useful for chaining multiplications
+            /// instead of divisions.
+            #[inline]
+            #[cfg(feature = "std")]
+            pub fn rsqrt(self) -> Self {
+                unsafe { _mm_rsqrt_ps(self) } // I'm missing the _mm256_rsqrt_ps version
```
thomcc, line 53: I think this requires more discussion. It's also not portable

Lokathor: Yeah, we want this op, but that's not the way to do it.

programmerjake: Also, `_mm_rsqrt_ps` is a very inaccurate approximation to reciprocal sqrt (iirc just 12 mantissa bits of the result are correct), for nearly all purposes you'd need something other than just that.

thomcc: Right, that's why I said it needs more discussion. To even use that intrinsic would require some decision on ffast-math stuff, and whatnot.

crates/core_simd/src/intrinsics.rs
```diff
@@ -49,6 +49,9 @@ extern "platform-intrinsic" {
     /// fsqrt
     pub(crate) fn simd_fsqrt<T>(x: T) -> T;
 
+    /// rsqrt
+    pub(crate) fn simd_rsqrt<T>(x: T) -> T;
```
workingjubilee, line 53: Alas, `simd_rsqrt` needs to be added to the compiler before we can invoke it.

workingjubilee:
Recently I found these two reports:
https://robert.ocallahan.org/2021/09/rr-trace-portability-diverging-behavior.html
https://cookieplmonster.github.io/2020/07/19/silentpatch-mass-effect/

I believe we should omit "rough approximation" functions from our usage until we have Rust-wide consensus on their acceptability, and I believe that LLVM emits these microarch-dependent instructions, which are even more intimidating than just an approximation. With that in mind, I am closing this PR. Thank you though, Raz, for bringing this up!

PR #124, opened 2021-05-26
Add load/store functions

calebzulawski:
This came up in zulip, so figured I'd add them.

### DIFF
--- a/crates/core_simd/src/first.rs
+++ b/crates/core_simd/src/first.rs
@@ -70,20 +70,20 @@ macro_rules! impl_vector {
             /// Stores a vector to a slice, without doing bounds checking
             ///
             /// # Safety
-            /// The length of `src` must be at least `LANES`.
+            /// The length of `dst` must be at least `LANES`.
             #[inline]
-            pub unsafe fn store_to_slice_unchecked(self, src: &mut [$type]) {
-                (src.as_mut_ptr() as *mut Self).write_unaligned(self)
+            pub unsafe fn store_to_slice_unchecked(self, dst: &mut [$type]) {
+                (dst.as_mut_ptr() as *mut Self).write_unaligned(self)
             }
 
             /// Stores a vector to a slice.
             ///
             /// # Panics
-            /// This function will panic if the length of `src` is not at least `LANES`.
+            /// This function will panic if the length of `dst` is not at least `LANES`.
             #[inline]
-            pub fn store_to_slice(self, src: &mut [$type]) {
-                assert!(src.len() >= LANES, "slice must be at least as long as vector");
-                unsafe { self.store_to_slice_unchecked(src) }
+            pub fn store_to_slice(self, dst: &mut [$type]) {
+                assert!(dst.len() >= LANES, "slice must be at least as long as vector");
+                unsafe { self.store_to_slice_unchecked(dst) }
             }
         }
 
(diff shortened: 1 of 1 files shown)


miguelraz:
Needs tests!

crates/core_simd/src/first.rs
```diff
@@ -42,6 +47,44 @@ macro_rules! impl_vector {
...
+            pub unsafe fn store_to_slice_unchecked(self, dst: &mut [$type]) {
+                (dst.as_mut_ptr() as *mut Self).write_unaligned(self)
+            }
+
+            /// Stores a vector to a slice.
+            ///
+            /// # Panics
+            /// This function will panic if the length of `dst` is not at least `LANES`.
+            #[inline]
+            pub fn store_to_slice(self, dst: &mut [$type]) {
```
workingjubilee, line 84: I am not sure these names are the way to go: we don't use "load" and "store", we use "read" and "write". This risks being _especially_ silly if it optimizes to something other than `STR`/`ST1`/etc.

workingjubilee, line 84: I am not sure these names are the way to go: we don't use "load" and "store", we use "read" and "write". This risks being _especially_ silly if it optimizes to something other than `STR`/`ST1`/etc.

crates/core_simd/src/first.rs
```diff
@@ -42,6 +47,44 @@ macro_rules! impl_vector {
...
+
+            /// Loads a vector from a slice.
+            ///
+            /// # Panics
+            /// This function will panic if the length of `src` is not at least `LANES`.
+            #[inline]
+            pub fn load_from_slice(src: &[$type]) -> Self {
+                assert!(src.len() >= LANES, "slice must be at least as long as vector");
+                unsafe { Self::load_from_slice_unchecked(src) }
+            }
```
workingjubilee, line 68: I'm fine with offering an unaligned read as the "default" interface, but panic machinery can often fail to optimize out (and even result in double-retests... a clear compiler error, but one we should be aware of). So my question is, do we want to have the infallible version offer a panicking interface or one that masks and zero-fills?

workingjubilee (commented):
Missing all those `#[inline]`s was a mistake, but I am not sure the rest of this PR is the way to go. If you want to split those out, we can merge them independently.

crates/core_simd/src/first.rs
calebzulawski: I was on the fence about the names.  At first I was thinking I wanted to distinguish between pointer and slice access, but now I remember that the `Read` and `Write` traits use slices...  On the other hand, SIMD instruction sets seem to use "load" and "store" relatively universally.  Read and write are probably most consistent.

crates/core_simd/src/first.rs
calebzulawski: I think panicking is appropriate.  This function will probably rarely be optimal but it's probably necessary to provide and it's no less efficient than zero-filling, I think.  Either way it's just a branch.  In my opinion `from_array` is the infallible version.

calebzulawski:
I've been thinking a little more about this PR and perhaps the correct way to read from a slice is to obtain an array from the slice (using whatever functions are available--if they aren't they should be added to `std`).  @workingjubilee made a good point somewhere in zulip that an array basically is a vector, the semantics are different, but `from_array` should just be a move instruction at worst.

I still feel like it's possible there's something we should be adding to help reading vectors from slices, but maybe it's more about arrays and not so much vectors.  Should we close this?

workingjubilee:
Yeah, I think we should revisit this later with a different approach when we are more confident there's a special desirable here. We should already be emitting `vmovups` and friends.

PR #127, opened 2021-06-01
Add mask width conversion

### LINKED ISSUE #115: Mask types should implement From from all other mask types with the same lane count
Allow the following for all of `mask8xN`, `mask16xN`, `mask32xN`, ... for equal `N`:
```rust
fn f(v: mask32x4) -> mask64x4 {
    From::from(v)
}
```

calebzulawski:
Resolves #115


programmerjake:
wouldn't we want to use sign-extension/truncation for converting full-masks?

calebzulawski:
Eventually yes, but we don't have an intrinsic for that yet.

workingjubilee:
seems good to me!

PR #128, opened 2021-06-04
add dot_product example

miguelraz:
Scratching this off the #110 list.

Special thanks to @Lokathor, @workingjubilee and the Zulip chat for answering questions on this one.

### DIFF
--- a/crates/core_simd/examples/dot_product.rs
+++ b/crates/core_simd/examples/dot_product.rs
@@ -13,7 +13,7 @@ use core_simd::*;
 // go along the resulting array and add up the result.
 // In the next example we will see if there
 //  is any difference to adding and multiplying in tandem.
-pub fn dot_prod_0(a: &[f32], b: &[f32]) -> f32 {
+pub fn dot_prod_scalar_0(a: &[f32], b: &[f32]) -> f32 {
     assert_eq!(a.len(), b.len());
 
     a.iter().zip(b.iter()).map(|(a, b)| a * b).sum()
@@ -26,7 +26,7 @@ pub fn dot_prod_0(a: &[f32], b: &[f32]) -> f32 {
 // hypothesis and benchmarks - we will mention them later on.
 // With the use of `fold`, we're doing a multiplication,
 // and then adding it to the sum, one element from both vectors at a time.
-pub fn dot_prod_1(a: &[f32], b: &[f32]) -> f32 {
+pub fn dot_prod_scalar_1(a: &[f32], b: &[f32]) -> f32 {
     assert_eq!(a.len(), b.len());
     a.iter()
         .zip(b.iter())
@@ -154,8 +154,8 @@ mod tests {
         let y: Vec<f32> = [2.0; 1003].to_vec();
 
         // Basic check
-        assert_eq!(0.0, dot_prod_0(&a, &b));
-        assert_eq!(0.0, dot_prod_1(&a, &b));
+        assert_eq!(0.0, dot_prod_scalar_0(&a, &b));
+        assert_eq!(0.0, dot_prod_scalar_1(&a, &b));
         assert_eq!(0.0, dot_prod_simd_0(&a, &b));
         assert_eq!(0.0, dot_prod_simd_1(&a, &b));
         assert_eq!(0.0, dot_prod_simd_2(&a, &b));
--- a/crates/core_simd/examples/dot_product.rs
+++ b/crates/core_simd/examples/dot_product.rs
@@ -1,39 +1,113 @@
 // Code taken from the `packed_simd` crate
 // Run this code with `cargo test --example dot_product`
+//use std::iter::zip;
+
 #![feature(array_chunks)]
+#![feature(slice_as_chunks)]
+// Add these imports to use the stdsimd library
+#![feature(portable_simd)]
 use core_simd::*;
 
-/// This is your barebones dot product implementation: 
-/// Take 2 vectors, multiply them element wise and *then*
-/// add up the result. In the next example we will see if there
-///  is any difference to adding as we go along multiplying.
+// This is your barebones dot product implementation:
+// Take 2 vectors, multiply them element wise and *then*
+// go along the resulting array and add up the result.
+// In the next example we will see if there
+//  is any difference to adding and multiplying in tandem.
 pub fn dot_prod_0(a: &[f32], b: &[f32]) -> f32 {
     assert_eq!(a.len(), b.len());
 
-    a.iter()
-    .zip(b.iter())
-    .map(|a, b| a * b)
-    .sum()
+    a.iter().zip(b.iter()).map(|(a, b)| a * b).sum()
 }
 
+// When dealing with SIMD, it is very important to think about the amount
+// of data movement and when it happens. We're going over simple computation examples here, and yet
+// it is not trivial to understand what may or may not contribute to performance
+// changes. Eventually, you will need tools to inspect the generated assembly and confirm your
+// hypothesis and benchmarks - we will mention them later on.
+// With the use of `fold`, we're doing a multiplication,
+// and then adding it to the sum, one element from both vectors at a time.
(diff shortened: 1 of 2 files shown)


miguelraz:
Failure on `riscv64gc` seems unrelated because something couldn't pull the `proptest` lib.

miguelraz:
I think this is good to merge. Any objections @workingjubilee ?

crates/core_simd/examples/dot_product.rs
```diff
@@ -0,0 +1,31 @@
...
+
+pub fn dot_prod(a: &[f32], b: &[f32]) -> f32 {
+    assert_eq!(a.len(), b.len());
+
+    // TODO handle remainder when a.len() % 4 != 0
+    a.array_chunks::<4>()
+        .map(|&a| f32x4::from_array(a))
+        .zip(b.array_chunks::<4>().map(|&b| f32x4::from_array(b)))
+        .map(|(a, b)| (a * b).horizontal_sum())
+        .sum()
```
calebzulawski, line 14: Horizontal sums are usually slow so this is probably a better way to do a dot product:

```rust
let mut sum = a.array_chunks::<4>()
    .map(|&a| f32x4::from_array(a))
    .zip(b.array_chunks::<4>().map(|&b| f32x4::from_array(b)))
    .map(|(a, b) a * b)
    .fold(f32x4::splat(0.), std::ops::Add)
    .horizontal_sum();
let remain = a.len() - (a.len() % 4);
sum += a[remain..].zip(&b[remain..]).map((a, b) a * b).sum();
```

The `fold` can be replaced with `sum` once that trait is implemented.

calebzulawski (changes requested):

crates/core_simd/examples/dot_product.rs
miguelraz: Oh that's neat - I think it would be good to document both cases and show performance differences.

miguelraz:
Todo:
- [ ] Rebase to master to exploit #138 
- [ ] Add Caleb's method, discuss tradeoffs
- [ ] Add a mul_add method from #138 
- [ ] Benchmark, at some point.

workingjubilee:
With #138 that is now unblocked~
( it should be zero-cost, tbh )

crates/core_simd/examples/dot_product.rs
```diff
@@ -0,0 +1,31 @@
...
+use core_simd::*;
+
+pub fn dot_prod(a: &[f32], b: &[f32]) -> f32 {
+    assert_eq!(a.len(), b.len());
+
+    // TODO handle remainder when a.len() % 4 != 0
+    a.array_chunks::<4>()
+        .map(|&a| f32x4::from_array(a))
+        .zip(b.array_chunks::<4>().map(|&b| f32x4::from_array(b)))
+        .map(|(a, b)| (a * b).horizontal_sum())
```
thomcc, line 13: This is a pretty bad dot-product example, honestly. You want to use something closer to the algorithm [here](https://gist.github.com/thomcc/4aceff7f4482dd6ab1eabe391e843c49#file-dot-bench-rs-L256-L265) to defer the hsum until the end, as opposed to doing one hsum per chunk (which is going to be terrible for perf).

That said, I guess it's just ported from packed_simd, and we can probably improve it later...

thomcc: Oh, this was already brought up.

workingjubilee:
I believe this can proceed if it is rebased/updated, with a minimal effort. @miguelraz Would you like to do so?

miguelraz:
Alright, I think I have some bandwidth now... once more unto the breach my friends! 🎉

crates/core_simd/examples/dot_product.rs
y0ast: I might be out of place here, but this seems to allocate more memory than necessary:

`.map(|(a, b) a * b)` will need ~len(a) many allocations, such that the results can be used in the subsequent fold.

Instead you could avoid that by using the `fold` directly:
```rust
let mut sum = a.array_chunks::<4>()
    .map(|&a| f32x4::from_array(a))
    .zip(b.array_chunks::<4>().map(|&b| f32x4::from_array(b)))
    .fold(f32x4::splat(0.), |acc (a, b)| acc + a * b)
    .horizontal_sum();
```
which allocates just 1 XMM register for accumulation (f32x4) and and the op inside the fold can be replaced by a `mul_add` (or some other fused multiply-add) as mentioned in the todo below.

crates/core_simd/examples/dot_product.rs
```diff
@@ -0,0 +1,54 @@
...
+    assert_eq!(a.len(), b.len());
+    a.iter()
+    .zip(b.iter())
+    .fold(0.0, |a, b| a * b)
+}
+
+pub fn dot_prod_simd_0(a: &[f32], b: &[f32]) -> f32 {
+    assert_eq!(a.len(), b.len());
+
+    // TODO handle remainder when a.len() % 4 != 0
```
scottmcm, line 29: I've put together an example that handles remainder, if you'd like to use it: <https://rust.godbolt.org/z/eMPEnc9or>

```rust
const LANES: usize = 8;
pub fn dot_prod_simd(a: &[f32], b: &[f32]) -> f32 {
    assert_eq!(a.len(), b.len());

    let (a_extra, a_chunks) = a.as_rchunks();
    let (b_extra, b_chunks) = b.as_rchunks();

    // These are always true, but for emphasis:
    assert_eq!(a_chunks.len(), b_chunks.len());
    assert_eq!(a_extra.len(), b_extra.len());

    let mut sums = [0.0; LANES];
    for ((x, y), d) in zip(a_extra, b_extra).zip(&mut sums) {
        *d = x * y;
    }

    let mut sums = Simd::from_array(sums);
    zip(a_chunks, b_chunks)
        .for_each(|(x, y)| {
            sums += Simd::from_array(*x) * Simd::from_array(*y);
        });

    sums.reduce_sum()
}
```

miguelraz: Awesome! I'll add it to the batch.

crates/core_simd/examples/dot_product.rs
miguelraz: I've added both of these versions to the recent PR. Marking as resolved.

miguelraz:
I've addressed all of my `TODO`s from above, and I'm punting the benchmarking to a future PR once more examples take shape and the proper benchmarking infra makes sense.

miguelraz:
@calebzulawski I believe your review is required.

miguelraz:
@calebzulawski This should be good to go now.

workingjubilee:
Would you kindly rebase commit https://github.com/miguelraz/stdsimd/commit/340491b2ce5dbc70ea4ebf75e912dd396901678e out of existence?

crates/core_simd/examples/dot_product.rs
```diff
@@ -0,0 +1,169 @@
...
+// `(a * b).reduce_sum()`: Multiply both f32x4 vectors together, and then reduce them.
+// This approach essentially uses SIMD to produce a vector of length N/4 of all the products,
+// and then add those with `sum()`. This is suboptimal.
+// TODO: ASCII diagrams
+pub fn dot_prod_simd_0(a: &[f32], b: &[f32]) -> f32 {
+    assert_eq!(a.len(), b.len());
+    // TODO handle remainder when a.len() % 4 != 0
+    a.array_chunks::<4>()
+        .map(|&a| f32x4::from_array(a))
+        .zip(b.array_chunks::<4>().map(|&b| f32x4::from_array(b)))
```
programmerjake, line 48: this could use `slice::as_simd`

workingjubilee: Is that better codegen?

programmerjake: probably, but i was more thinking that it's the function designed to split a slice into simd chunks, so we should use it instead of going through arrays.

PR #132, opened 2021-06-05
add more doctests

miguelraz:
DO NOT MERGE - WIP
As well as
- clarify that rounding ties away from zero, as is usual in Rust
- comment on wrapping addition for `horizontal_sum()` on floats


programmerjake:
See also: The debate on if we should use the slow C `round` semantics or use the ties-to-even semantics that are generally faster due to better cpu support:
https://github.com/rust-lang/rust/issues/55107

miguelraz:
@programmerjake ooooh, thanks for the link!

workingjubilee:
This PR was unfortunately disrupted by many internal rearrangements.  My apologies.
I think it would be good if you wanted to reopen with an updated version but I am closing this for now.

PR #133, opened 2021-06-07
add hello world example to beginnersguide.md

miguelraz:
I ran into this, so why not spare others the trouble as well.


miguelraz:
Disregard in favor of #134 - derp derp

PR #134, opened 2021-06-07
add helloworld to README

miguelraz:
Might as well have a hello world

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,39 @@ We can also be found on [Zulip][zulip-project-portable-simd].
 
 If you are interested in support for a specific architecture, you may want [stdarch] instead.
 
+## Hello World
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -12,6 +12,39 @@ We can also be found on [Zulip][zulip-project-portable-simd].
 
 If you are interested in support for a specific architecture, you may want [stdarch] instead.
 
+## Hello World
+
+Now we're gonna dip our toes into this world with a small SIMD "Hello, World!" example. Make sure your compiler is up to date by running 
+
+```bash
+rustup update
+```
```
workingjubilee, line 21: We might as well also include a note that says users need to use nightly, because this isn't available on stable. 😅 
Either by changing the command here to `rustup update -- nightly` or just mentioning that they have to use `rustup default nightly` or else `cargo +nightly {build,test,run}`.

Issue #14, opened 2020-09-30
Impl special functions for SIMD


programmerjake:
Need all of:
- [ ] trig./hyperbolic functions: #6
- [ ] floor/ceil/round/trunc/fract
- [ ] abs
- [ ] signum
- [ ] copysign
- [ ] mul_add
- [ ] div_euclid/rem_euclid
- [ ] powi/powf
- [ ] sqrt
- [ ] exp/exp2/ln/log/log2/log10
- [ ] cbrt
- [ ] hypot
- [ ] exp_m1/ln_1p
- [ ] clamp
- [ ] recip
- [ ] to_degrees/to_radians
- [ ] max/min
- [ ] to_int_unchecked
- [ ] rem
- [ ] count_ones/count_zeros
- [ ] leading_zeros/trailing_zeros
- [ ] leading_ones/trailing_ones
- [ ] rotate_left/rotate_right
- [ ] swap_bytes/reverse_bits
- [ ] from_be/from_le/to_be/to_le
- [ ] saturating_add/saturating_sub/saturating_neg/saturating_abs
- [ ] saturating_mul/saturating_pow
- [ ] wrapping_add/wrapping_sub/wrapping_mul/wrapping_pow
- [ ] wrapping_div/wrapping_rem/wrapping_div_euclid/wrapping_rem_euclid
- [ ] wrapping_neg/wrapping_abs
- [ ] wrapping_shl/wrapping_shr
- [ ] overflowing_add/overflowing_sub
- [ ] overflowing_mul/overflowing_pow
- [ ] overflowing_div/overflowing_div_euclid
- [ ] overflowing_rem/overflowing_rem_euclid
- [ ] overflowing_neg/overflowing_abs
- [ ] overflowing_shl/overflowing_shr
- [ ] pow
- [ ] div_euclid/rem_euclid
- [ ] is_positive/is_negative
- [ ] to_be_bytes/to_le_bytes/to_ne_bytes/from_be_bytes/from_le_bytes/from_ne_bytes

Lokathor:
I don't believe we have non-overflowing/non-wrapping ops actually.

programmerjake:
Having ops that panic on overflow (like Rust's standard integer ops in debug mode) seems like something that would be useful for debugging, even if it has a runtime penalty. It could be disabled by Release mode, like usual.

calebzulawski:
I would like to add `as_slice` and `as_array` functions to this list.

[...]

PR #138, opened 2021-06-13
Add various fns

### LINKED ISSUE #102: Support for Fused Multiply-Add (FMA)
From my limited testing, multiplying and then adding `SimdF32`/`SimdF64` does not result in FMA instructions, and rightly so, however it would be useful to have a method on these types and a corresponding platform intrinsic to generate LLVM calls to `llvm.fma.*` intrinsics, which support vector types.

Edit: In fact it looks like the `simd_fma` platform intrinsic is already available, just not used by the crate?

calebzulawski:
Adds
* `Sum`/`Product` traits
* `recip`/`to_degrees`/`to_radians`/`min`/`max`/`clamp`/`signum`/`copysign` (#14)
* `mul_add` (#14, fixes #102)

### DIFF
--- a/crates/core_simd/src/vector/float.rs
+++ b/crates/core_simd/src/vector/float.rs
@@ -136,6 +136,47 @@ macro_rules! impl_float_vector {
                 let magnitude = self.to_bits() & !Self::splat(-0.).to_bits();
                 Self::from_bits(sign_bit | magnitude)
             }
+
+            /// Returns the minimum of each lane.
+            ///
+            /// If one of the values is `NAN`, then the other value is returned.
+            #[inline]
+            pub fn min(self, other: Self) -> Self {
+                // TODO consider using an intrinsic
+                self.is_nan().select(
+                    other,
+                    self.lanes_ge(other).select(other, self)
+                )
+            }
+
+            /// Returns the maximum of each lane.
+            ///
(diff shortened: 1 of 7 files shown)


crates/core_simd/src/vector/float.rs
```diff
@@ -36,13 +36,44 @@ macro_rules! impl_float_vector {
...
+            #[inline]
+            pub fn to_degrees(self) -> Self {
+                // to_degrees uses a special constant for better precision, so extract that constant
+                self * Self::splat($type::to_degrees(1.))
+            }
+
+            /// Converts each lane from degrees to radians.
+            #[inline]
+            pub fn to_radians(self) -> Self {
+                self * Self::splat(core::$type::consts::PI / 180.)
```
programmerjake, line 75: this could also use a special constant for more precision

miguelraz: I'd hope the division gets optimized to a fmul here, otherwise that would be bad.

calebzulawski: It should be optimized to a constant.

workingjubilee:
Hello! The question arose in the conversation of: "Is this ready to merge?"

calebzulawski:
Yep!

PR #139, opened 2021-06-15
Add SimdArray trait and safe gather/scatter API

workingjubilee:
This PR has four parts, without which it doesn't make a lot of sense:

0. The introduction of the `SimdArray` trait for abstraction over vectors.
1. The implementation of private vector-of-pointers types.
2. Using these to allow constructing vectors with `SimdArray::gather_{or, or_default, select}`.
3. Using these to allow writing vectors using `SimdArray::scatter{,_select}`.

### DIFF
--- a/crates/core_simd/src/array.rs
+++ b/crates/core_simd/src/array.rs
@@ -1,0 +1,120 @@
+use crate::masks::*;
+use crate::vector::*;
+
+/// A representation of a vector as an "array" with indices, implementing
+/// operations applicable to any vector type based solely on "having lanes",
+/// and describing relationships between vector and scalar types.
+pub trait SimdArray<const LANES: usize>: crate::LanesAtMost32
+where
+    SimdUsize<LANES>: crate::LanesAtMost32,
+    SimdIsize<LANES>: crate::LanesAtMost32,
+    MaskSize<LANES>: crate::Mask,
+    Self: Sized,
+{
+    /// The scalar type in every lane of this vector type.
+    type Scalar: Copy + Sized;
+
+    /// Generates a SIMD vector with the same value in every lane.
+    #[must_use]
+    fn splat(val: Self::Scalar) -> Self;
+}
+
+macro_rules! impl_simdarray_for {
+    ($simd:ident {type Scalar = $scalar:ident;}) => {
+        impl<const LANES: usize> SimdArray<LANES> for $simd<LANES>
+            where SimdUsize<LANES>: crate::LanesAtMost32,
+            SimdIsize<LANES>: crate::LanesAtMost32,
+            MaskSize<LANES>: crate::Mask,
+            Self: crate::LanesAtMost32,
+        {
+            type Scalar = $scalar;
+
+            #[must_use]
+            #[inline]
+            fn splat(val: Self::Scalar) -> Self {
+                [val; LANES].into()
+            }
+        }
+    };
+
+    ($simd:ident $impl:tt) => {
+        impl<const LANES: usize> SimdArray<LANES> for $simd<LANES>
+            where SimdUsize<LANES>: crate::LanesAtMost32,
+            SimdIsize<LANES>: crate::LanesAtMost32,
+            MaskSize<LANES>: crate::Mask,
+            Self: crate::LanesAtMost32,
+        $impl
+    }
+}
+
+impl_simdarray_for! {
+    SimdUsize {
+        type Scalar = usize;
+    }
+}
+
+impl_simdarray_for! {
+    SimdIsize {
+        type Scalar = isize;
+    }
+}
+
+impl_simdarray_for! {
+    SimdI8 {
+        type Scalar = i8;
+    }
+}
+
+impl_simdarray_for! {
+    SimdI16 {
+        type Scalar = i16;
+    }
+}
+
+impl_simdarray_for! {
+    SimdI32 {
+        type Scalar = i32;
+    }
+}
+
+impl_simdarray_for! {
+    SimdI64 {
+        type Scalar = i64;
+    }
+}
+
+impl_simdarray_for! {
+    SimdU8 {
+        type Scalar = u8;
+    }
+}
+
+impl_simdarray_for! {
+    SimdU16 {
+        type Scalar = u16;
+    }
+}
+
+impl_simdarray_for! {
+    SimdU32 {
+        type Scalar = u32;
+    }
+}
+
+impl_simdarray_for! {
+    SimdU64 {
+        type Scalar = u64;
+    }
+}
+
+impl_simdarray_for! {
+    SimdF32 {
+        type Scalar = f32;
+    }
+}
+
+impl_simdarray_for! {
+    SimdF64 {
+        type Scalar = f64;
+    }
+}
--- a/crates/core_simd/src/array.rs
+++ b/crates/core_simd/src/array.rs
@@ -1,6 +1,6 @@
 use crate::intrinsics;
 use crate::masks::*;
-use crate::vector::ptr::SimdConstPtr;
+use crate::vector::ptr::{SimdConstPtr, SimdMutPtr};
 use crate::vector::*;
 
 /// A representation of a vector as an "array" with indices, implementing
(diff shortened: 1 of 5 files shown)


crates/core_simd/src/array.rs
```diff
@@ -0,0 +1,165 @@
...
+    /// let result = SimdI32::<4>::gather_or(&vec, indices, alt);
+    /// assert!(result == SimdI32::from_array([-5, 13, 10, 15]), "oh no, got {:?}", result);
+    /// ```
+    #[must_use]
+    #[inline]
+    fn gather_or(slice: &[Self::Scalar], idxs: SimdUsize<LANES>, or: Self) -> Self {
+        // IMPORTANT: We MUST construct our gather mask before we derive a pointer!
+        let mask = idxs.lanes_lt(SimdUsize::splat(slice.len())).to_int();
+        // IMPORTANT: We MUST obtain a pointer via as_ptr or an equivalent method!
+        let base_ptr = SimdConst::splat(slice.as_ptr());
```
Lokathor, line 59: why is it important that these steps happen in this order?

workingjubilee: Oh, good point.

If you don't do it precisely in that order it makes the following calls potentially UB, I should make a note of that. :D

Lokathor: ah, because len is &self, which invalidates the pointer if it's after?

incidentally, shouldn't this be as_mut_ptr?

workingjubilee: Yeah, re: &self.

Why would it be `as_mut_ptr`? This one is a gather, not a scatter. I am pretty sure obtaining `*mut T` from `&[T]` would be incorrect, unless we do in fact need to take `&mut [T]` to only do a gather. ^^;

Lokathor: oh this is gather, right

Lokathor: wait if this is gather and they're both &self methods then the ordering doesn't matter

workingjubilee: Oh OK~ Then I'll add those notes to the scatter. :D

Lokathor: some more background: &mut T asserts it's unique, but &T does not, and so making more than one &T on the same value won't invalidate previous *const T you have to the same object.

This is not true for &mut T, with &mut T you *do* want to have the appropriate level of paranoia.

run it all through miri some day.

crates/core_simd/src/array.rs
```diff
@@ -0,0 +1,165 @@
...
+            SimdIsize<LANES>: crate::LanesAtMost32,
+            MaskSize<LANES>: crate::Mask,
+            Self: crate::LanesAtMost32,
+        {
+            type Scalar = $scalar;
+
+            #[must_use]
+            #[inline]
+            fn splat(val: Self::Scalar) -> Self {
+                [val; LANES].into()
```
miguelraz, line 80: This is neat!
Would I also be able to do
```rust
[1,2,3,4].into(i32x4)
```
?

Lokathor: you don't pass a type into into as an argument. If you need to pick the target type, i think you can call .into::<T>() if you have to, but usually it's either inferred or you'd call T::from(val) instead.

miguelraz: Oh derp, that's right.
It'd be nice to have `[1,2,3,4].into::i32x4::()` I think.

workingjubilee: No, `into` doesn't take an argument like that. It's relying on type inference here because Rust knows the return type is Self, so it has to be using the `impl Into<${Self}> for [T; N]`.

workingjubilee: And unfortunately you can't turbofish it because the generic is actually on `Into<A> for Z`, not the function, whereas `collect` can be turbofished because collect is written `fn collect<B>(self) -> B`.

crates/core_simd/src/vector/ptr.rs
```diff
@@ -0,0 +1,55 @@
+//! Private implementation details of public gather/scatter APIs.
+use crate::SimdUsize;
+use core::mem;
+
+/// A vector of *const T.
+#[derive(Debug, Copy, Clone)]
+#[repr(simd)]
+pub(crate) struct SimdConst<T, const LANES: usize>([*const T; LANES]);
+
+impl<T, const LANES: usize> SimdConst<T, LANES>
```
calebzulawski, line 10: I would call this `SimdConstPtr`, I know it's only internal for now but I was pretty confused what it meant.

crates/core_simd/src/array.rs
calebzulawski, line 80: I would probably just write `Self::splat(val)` even though I know it's the same

crates/core_simd/src/array.rs
```diff
@@ -0,0 +1,165 @@
...
+    /// # use core_simd::*;
+    /// let vec: Vec<i32> = vec![10, 11, 12, 13, 14, 15, 16, 17, 18];
+    /// let indices = SimdUsize::<4>::from_array([9, 3, 0, 5]);
+    ///
+    /// let result = SimdI32::<4>::gather(&vec, indices);
+    /// assert!(result == SimdI32::from_array([0, 13, 10, 15]));
+    /// ```
+    #[must_use]
+    #[inline]
+    fn gather(slice: &[Self::Scalar], idxs: SimdUsize<LANES>) -> Self
```
calebzulawski, line 35: I'm not opposed to this in the trait necessarily, but I think there should still be an implementation on the type, too.

crates/core_simd/src/array.rs
```diff
@@ -0,0 +1,165 @@
+use crate::intrinsics;
+use crate::masks::*;
+use crate::vector::ptr::SimdConst;
+use crate::vector::*;
+
+/// A representation of a vector as an "array" with indices,
+/// implementing operations applicable to any vector type and
+/// describing relationships between vector and scalar types.
+pub trait SimdArray<const LANES: usize>: crate::LanesAtMost32
```
calebzulawski, line 9: This trait should be sealed, I think, based on the function implementations.

Do we need to be concerned about backwards compatibility?  What happens if we want to add more functions to this trait in the future?

calebzulawski (changes requested):

crates/core_simd/src/array.rs
workingjubilee: How does anyone implement this for their own type when it requires them to implement LanesAtMost32?

calebzulawski: Hmm, that trait is sealed, so I guess it effectively is sealed already.

calebzulawski: Looking over https://rust-lang.github.io/rfcs/1105-api-evolution.html to be sure, it looks like we should be good with backwards compatibility as long as the trait is sealed.

crates/core_simd/src/array.rs
Lokathor: <x as Into<f32x4>>.into()

...maybe? i forget if that works

crates/core_simd/src/array.rs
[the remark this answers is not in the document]
workingjubilee: My operating theory is that "array" generalizes here in a few important ways:

0. By default: arrays go on the stack, Vec goes to the heap. The name-clash with Vector might suggest the wrong things about data locality, which I've noticed Rustaceans care about a lot.
1. Assuming people successfully "deconflict" Vec and Vector, arrays are an existing type, so "this is an array, but SIMD" is only one bit of new data to absorb whereas "this is a vector, and SIMD" is two bits (three if counting the disambiguation bit).
2. The type is in fact more or less an array in the sense of readily admitting conversions to such (to the point of allowing some safe transmutes!), even if back again is slightly trickier.
3. People who use APL derivatives, Fortran, or ndarray libs seem to have no problem calling them arrays, even if their code  does vectorize quite nicely, and while we're from a slightly different "tradition", programming-wise, I think it's useful as a signpost that it's fine to embrace a level of abstraction from the vector registers... even if we still have to talk about them anyways, IMO.

But I'm not like... married to this? I just felt like experimenting with this alternative would be useful, especially after noticing that in conversations with other programmers I have to say "vectors... n-no, not Vec..." Even formal papers on portable SIMD APIs for C++ or Java have to add that caveat.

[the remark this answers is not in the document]
Lokathor: I call simd values "registers", which is slightly wrong also, but not too much so. Maybe that applies more to `safe_arch` than stdsimd

workingjubilee: Yeah, I was thinking while I was writing this that we might want to move the actual sealing trait around, but that we can pick any sealed trait.

programmerjake:
Naming question: Do we want the `gather`/`gather_or` methods to take a mask? or will the masked variant be named `gather_masked`?

Maybe like:
```rust
// excuse my syntax
impl ... for SimdUsize<...> {
    fn gather_or<T>(self, slice: &[T], mask: ..., fallback: Simd<[T; LANES]>) -> Simd<[T; LANES]>;
    fn gather_or_default<T>(self, slice: &[T], mask: ...) -> ...;
}
```

crates/core_simd/src/array.rs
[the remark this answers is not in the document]
calebzulawski: At that point wouldn't we want to strike "vector" from our terminology everywhere?

My concern is that as unfortunate the name clash is, we probably should still call them vectors simply because that's what they're called everywhere else.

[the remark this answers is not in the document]
Lokathor: I think it's only confusing when you say it out loud and also follow the "it's spelled Vec but pronounced Vector" line. Everyone knows that you should spell it Vec and *also* pronounce it Vec.

In text form you wouldn't even have a problem.

Though I'll say that SimdArray doesn't really tell me what ops it supports. Can we just have these ops be on two separate traits called SimdScatter and SimdGather?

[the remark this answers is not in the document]
workingjubilee: Hmm. My thought is that having a single trait that defines the operations on the SIMD-vector-qua-SIMD-vector (or SIMD-array-qua-SIMD-array) will help understanding the API a fair amount. But I think we should discuss this while sync.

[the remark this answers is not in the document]
workingjubilee: > In text form you wouldn't even have a problem.
Unfortunately, this has not been my experience.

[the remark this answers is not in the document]
thomcc: > Can we just have these ops be on two separate traits called SimdScatter and SimdGather?

I think this line of reasoning leads to dozens of traits, and I'm fairly strongly opposed to following that route. I think the stdlib has been fairly conservative about trait definitions, which is a *good* thing.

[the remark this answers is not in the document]
Lokathor: Ehhh, core::ops has one trait per op and it works fine.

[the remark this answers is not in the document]
calebzulawski: I think it would be fair to say that each op trait inherently needs to be separate since the ops are separate, this is a little more like Iterator, lots of functionality on one type.

workingjubilee:
discussed in meeting, consensus was largely:

0. `SimdArray` is an okay name, but we may want to rename `SimdArray` to `Vector` in the very near future (like, right before debut to nightly), just to find out the Rust community's opinion on it. :^)
1. Having methods on a trait is fine, but we probably want to max out on a very small number of "big" crate-defined expected-to-be-imported traits for users though. `{Vector,SimdArray}` and perhaps `{Mask,SimdMask}`?
2. We'd like inherent implemented traits in a bright and shiny future but it's fine if they're not, as long as we're **very** disciplined about limiting required traits to use our API, see previous remark.
3. Maybe someday we'll have `SimdInt` or `SimdFloat` trats and the like, but definitely not in this crate for now.

No "default, unmasked" `fn gather`, call it `fn gather_or_default` instead.
And it's fine if `fn scatter` has no exact parallel.
`fn gather_select` and `fn scatter_select` will be the masked versions.

crates/core_simd/src/array.rs
```diff
@@ -0,0 +1,244 @@
+use crate::intrinsics;
+use crate::masks::*;
+use crate::vector::ptr::{SimdConstPtr, SimdMutPtr};
+use crate::vector::*;
+
+/// A representation of a vector as an "array" with indices, implementing
+/// operations applicable to any vector type based solely on "having lanes",
+/// and describing relationships between vector and scalar types.
+pub trait SimdArray<const LANES: usize>: crate::LanesAtMost32
```
calebzulawski, line 9: Looking this over again, shouldn't the lane count be an associated const?

```rust
pub trait SimdArray {
    const LANES: usize;
}
```

It's impossible to implement `SimdArray` with different lengths for a single type.

workingjubilee: Hm... is that a bug?

workingjubilee: I am OK with using an assoc const fwiw, I just genuinely don't know which one would actually be best here.

calebzulawski: The associated const would let you do something like `<T as SimdArray>::LANES` which I could see being useful.  No easy way to extract the lane count otherwise, I think.

thomcc: Also it follows the best practice of "if you want to allow this type to implement the trait multiple times, use an associated ~~type~~ const, otherwise use a ~~type~~ const parameter".

workingjubilee: I can add the assoc const on its own but removing the const from the trait definition induces a cycle in evaluating the well-formedness of the const generic, so I think it can't be done.

calebzulawski: I'll check it out in a bit, just to see if I can convince it.  I don't see why it shouldn't work.

crates/core_simd/src/array.rs
```diff
@@ -0,0 +1,246 @@
...
+    /// let mut vec: Vec<i32> = vec![10, 11, 12, 13, 14, 15, 16, 17, 18];
+    /// let idxs = SimdUsize::<4>::from_array([9, 3, 0, 5]);
+    /// let vals = SimdI32::from_array([-5, -4, -3, -2]);
+    /// let mask = MaskSize::from_array([true, true, true, false]); // Note the mask of the last lane.
+    ///
+    /// vals.scatter_select(&mut vec, mask, idxs);
+    /// assert_eq!(vec, vec![-3, 11, 12, -4, 14, 15, 16, 17, 18]);
+    /// ```
+    #[inline]
+    fn scatter_select(
```
programmerjake, line 118: We should document what happens when multiple lanes write to the same location, iirc later lanes overwrite earlier lanes according to LLVM IR semantics.

workingjubilee: This appears to be correct: added those to the doctests.

crates/core_simd/src/array.rs
workingjubilee: cc @rust-lang/project-const-generics since we seem to be running into an amusing niche case in our expectations of how const things work and interact with Rust programming idioms, vs. how they do in practice.

for note: my attempt was to bound everything currently described by LANES on `Self::LANES` and instead write stuff like
```rust
impl SimdArray for Simd{Ty}<LANES> {
    const LANES: usize = LANES;
}
```
and I experimented with a few different configurations of where-bounds and the like, but I kept hitting cycles or outright failures to type check.

calebzulawski (approved):

crates/core_simd/src/array.rs
workingjubilee: After finding that the problem is probably in the bounds not being moved mostly into the implementations of the traits, we will probably catch up to refactoring this in after a bit more tinkering around.

PR #142, opened 2021-06-28
Combine LanesAtMost32 and SimdArray into a single trait "Vector"

### LINKED ISSUE #51: Improve wasm32 to_array performance

calebzulawski:
Attempts to fix some unresolved questions in #139 regarding `SimdArray` having a generic parameter.

In particular, this made it not appropriate for replacing `LanesAtMost32`.  Additionally, it made it impossible to use in a context where you otherwise don't know the lane count, e.g. `impl Vector`.

An unfortunate side effect of this change is that scatter/gather no longer work in the trait (nor does anything else that references the lane count in a type.  This requires the super-unstable `const_evaluatable_checked` feature).

I also threw in the change from `as_slice` to `as_array` as discussed in zulip, and fixes #51.


crates/core_simd/src/vector/mod.rs
```diff
@@ -0,0 +1,131 @@
...
+pub(crate) mod ptr;
+
+mod sealed {
+    pub trait Sealed {}
+}
+
+/// A representation of a vector as an "array" with indices, implementing
+/// operations applicable to any vector type based solely on "having lanes",
+/// and describing relationships between vector and scalar types.
+pub trait Vector: sealed::Sealed {
```
workingjubilee, line 19: So, I am fine with moving the trait into this module, but if we go to mod.rs it makes it harder to navigate because every mod.rs is the same as every other mod.rs in terms of terminal names (which isn't a problem now but may become one soon). I would prefer to retain the vector.rs style, even if it is just a "foyer" without much content of its own.

crates/core_simd/src/first.rs
```diff
@@ -24,51 +24,166 @@ macro_rules! impl_vector {
...
+            }
+        }
+
+        impl<const LANES: usize> $name<LANES>
+        where
+            Self: crate::Vector,
+            crate::MaskSize<LANES>: crate::Mask,
+            crate::SimdIsize<LANES>: crate::Vector,
+            crate::SimdUsize<LANES>: crate::Vector,
+        {
```
workingjubilee, line 37: I think we may be capable of getting rid of this file, which is an undesirable aberration in our implementation, soon, so I would prefer these remain factored out of this file, regardless of how this is achieved.

workingjubilee (changes requested):
the first line of code
greets the cutting room floor
soon, the raking begins

crates/core_simd/src/first.rs
calebzulawski: I originally tried combining it into the root of the vector module, but it wasn't obvious that there was an order between the submodules and macros.  I moved the file into `vector/vector_impl.rs` and was able to move the `Vector` trait implementation into it, too.

crates/core_simd/src/comparisons.rs
```diff
@@ -1,13 +1,11 @@
...
 
 macro_rules! implement_mask_ops {
     { $($vector:ident => $mask:ident ($inner_ty:ident),)* } => {
         $(
             impl<const LANES: usize> crate::$vector<LANES>
             where
-                crate::$vector<LANES>: LanesAtMost32,
-                crate::$inner_ty<LANES>: LanesAtMost32,
-                crate::$mask<LANES>: crate::Mask,
+                LaneCount<LANES>: SupportedLaneCount,
```
workingjubilee, line 8: Well, this kind of thing is a huge improvement! And hopefully will make some other simplifications easier.

crates/core_simd/src/lib.rs
```diff
@@ -29,20 +27,16 @@ mod comparisons;
...
 mod ops;
 mod round;
 mod vendor;
 
-mod math;
-
-mod lanes_at_most_32;
-pub use lanes_at_most_32::LanesAtMost32;
+mod lane_count;
+pub use lane_count::*;
```
workingjubilee, line 36: Return of LanesAtMost: General Trait Bound's Revenge

workingjubilee (approved):
This is really good! Hopefully we can work a little magic with the traits Soon™ but this is exciting and anything we do will almost certainly be predicated on this changeset.

PR #147, opened 2021-07-20
Add fmt and clippy to CI

calebzulawski:


workingjubilee:
If we are going to use clippy I would prefer a deliberately conservative subset, like limiting it to e.g. correctness lints.

calebzulawski:
We pass clippy right now so I think it would be fine to leave it and then disable anything that becomes problematic.  I didn't enable failure on warnings, though we nearly pass without warnings too.

workingjubilee (approved):
Oh, I guess that's fair then~ LGTM

PR #152, opened 2021-07-28
Feature/const eval checked

calebzulawski:
Adds the `const_evaluatable_checked` cargo feature for some experimental (but very useful) function implementations.  Adds testing all cargo features to CI.


workingjubilee:
Looks good to me!

PR #154, opened 2021-08-07
Change vectors to be generic over element type.

calebzulawski:
Most important thing to note upfront: despite the pretty significant implementation changes, the API is basically unchanged.

This PR results in changing `SimdI32<4>` to `Simd<i32, 4>`, `Mask8<2>` to `Mask<i8, 2>`, etc.

This results in a pretty substantial reduction of the API, which now only contains 6 things (not counting type aliases): `Simd`, `Mask`, `SimdElement`, `MaskElement`, `LaneCount`, and `SupportedLaneCount`.

This allows functions such as `splat`, `shuffle`, `gather_or`, `lanes_eq`, `select`, etc to be implemented generically once for all vectors, rather than once for each element type (with tons of macros 🤢).  Users can still call these functions in the same way as before this PR--only the implementation is different.

There is one notable downside: rustdoc really does not know how to handle this situation and the page for the `Simd` type is massive.  This PR _does not_ attempt to unify implementations over similar types, such as all float types or all unsigned integers.  A future PR changing some of the functions to trait functions could reduce this and make those functions a little more discoverable.  Alternatively, there might be some improvements we can suggest/add to rustdoc to help wrangle the extra functions.


crates/core_simd/src/vector.rs
```diff
@@ -12,21 +9,394 @@ pub use uint::*;
...
 mod sealed {
     pub trait Sealed {}
 }
+use sealed::Sealed;
+
+/// Marker trait for types that may be used as SIMD vector elements.
+pub unsafe trait SimdElement: Sealed + Copy {
+    /// The mask element type corresponding to this element type.
+    type Mask: MaskElement;
+}
```
workingjubilee, line 342: Why is this unsafe to implement?

workingjubilee (commented):
> Most important thing to note upfront: despite the pretty significant implementation changes, the API is basically unchanged.

The API has undergone a significant change, IMO, in that it is a lot easier to use generically in most ways. Admittedly that's not what everyone might typically be thinking by API, but because SIMD often involves type shenanigans, I think it is relevant, eh?

Which is to say I think this will probably be a vast improvement for actual use also.
I have very small legibility remarks but I think this is good to merge.

crates/core_simd/src/vector.rs
Lokathor: probably makes rustc flip out and ICE if you try to do simd over a non-primitive, if i had to guess.

Lokathor:
i think if you make a type alias and then impl methods on that type alias, rustdoc will collect the methods there.

However, that would halfway undo the PR

crates/core_simd/src/masks/bitmask.rs
```diff
@@ -1,76 +1,73 @@
...
+use crate::{LaneCount, MaskElement, Simd, SupportedLaneCount};
+use core::marker::PhantomData;
 
 /// A mask where each lane is represented by a single bit.
 #[repr(transparent)]
-pub struct BitMask<const LANES: usize>(<LaneCount<LANES> as SupportedLaneCount>::BitMask)
+pub struct Mask<Element, const LANES: usize>(
+    <LaneCount<LANES> as SupportedLaneCount>::BitMask,
+    PhantomData<Element>,
+)
```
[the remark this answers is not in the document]
calebzulawski: That's a good point.  The standard library tends to use a single letter.  I would probably just opt for `T` if we go that route.

crates/core_simd/src/vector.rs
calebzulawski: Yeah, it's unsafe for two reasons--one is that rustc will produce unfortunate monomorphisation errors for non-scalars.  But another reason to make it unsafe is in the future `repr(simd)` could be expanded to some arbitrary type that we haven't considered, and it may suddenly allow some broken things.

calebzulawski:
Even if that is true, is that desirable?  It may be a temporary "solution" but in some ways I think it's more confusing because we want people to use `Simd`, not `SimdF32`.

crates/core_simd/src/vector.rs
workingjubilee: This is fair, I basically just wanted to get "presumed unsafe to implement because it relies on compiler-backed monomorphization guarantees" on the record. :^)
```suggestion
/// Marker trait for types that may be used as SIMD vector elements.
/// SAFETY: This trait implies the compiler knows how to monomorphize the element in question.
/// Implementing it for a type asserts the compiler won't miscompile it in a `#[repr(simd)]` struct.
pub unsafe trait SimdElement: Sealed + Copy {
    /// The mask element type corresponding to this element type.
    type Mask: MaskElement;
}
```

Lokathor:
In practice many things still can't be generic over your element type anyway. Slightly jiggering things to make rustdoc behave a little better doesn't make it any worse.

workingjubilee:
Given there are multiple routes of exploration for improving the docs I am okay with deferring them somewhat to a later PR, even if they matter.

crates/core_simd/src/masks/bitmask.rs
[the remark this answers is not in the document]
calebzulawski: Decided to go with `T`, like arrays and `Vec`.

PR #159, opened 2021-09-13
Use new bitmask intrinsics with byte arrays

calebzulawski:
Pending rust-lang/rust#88868.


workingjubilee:
Deleted code. It's beautiful. 🥲

PR #161, opened 2021-09-20
Swizzle

calebzulawski:
This replaces the "shuffle" API with a more generic implementation.  With these changes, you can now create any length vector from any other length vector (a "swizzle").

The API isn't _great_ for creating arbitrary, one-off shuffles (e.g. `_mm_shuffle_*`), but it's great for algorithmically generating shuffles.  We could consider something like a macro for one-off shuffles, but not sure if that's a great idea, I left that for later.

One thing I don't love in this PR is the `Which` enum, at minimum I think it could use a better name.


thomcc:
So you have to define a dummy struct and implement a trait for each swizzle? That doesn't seem great...

calebzulawski:
Well, it kind of is a dummy struct I suppose, but traits are the only way right now to have a dependency on a const that you can parameterize.

calebzulawski:
To clarify, there is no way to do `<T, const V: T>`

Lokathor:
This is... a remarkably poor direction to take.

calebzulawski:
Well, I am open to any other ideas, but I don't think there is _any_ other way to do it, other than with a trait const.

It's worth noting that 1) this allowed implementation of `rotate_{left,right}` which was previously not possible, and 2) you can still do something like this for convenience:

```
pub struct Shuffle4<const I: [usize; 4]>;

impl<const I: [usize; 4]> Swizzle<4, 4> for Shuffle4<I> {
    const INDEX: [usize; 4] = I;
}
```

workingjubilee:
> Well, I am open to any other ideas, but I don't think there is any other way to do it, without a trait const.

A "trait const"? Do you mean a trait-level const generic parameter?
As I have mentioned before, I do not think that is the worst sin.
The code using a trait could ideally, for a user, look something like

```rust
use core::simd::{f32x4, f32x8, Swizzle};

fn get_prime_indices(floats: f32x8) -> f32x4 {
  Swizzle::swizzle<{ [2, 3, 5, 7] }>(floats)
}
```

That's fine? I haven't reviewed this code as closely as I would like but it does seem... slightly heavier weight than that.

calebzulawski:
I meant associated const.  Unfortunately I don't think there's any way to do what you're suggesting, because there is a const panic to catch invalid indices, as well as an index conversion in the case of the 2-argument swizzle.

Even if you forgo avoiding monomorphisation errors, you still can't pass the array in via a generic parameter, because the index type can't depend on outer parameters.

Lokathor:
isn't the index type always usize? or, is it sometimes u32 instead or something?

can't we make it always u32? i mean it's a const so the actual thing is gonna usually be literals anyway.

programmerjake:
> isn't the index type always usize? or, is it sometimes u32 instead or something?
> 
> can't we make it always u32? i mean it's a const so the actual thing is gonna usually be literals anyway.

I think what @calebzulawski meant was that the array length of the index type was the problem, not `u32` vs. `usize`.

calebzulawski:
The usize isn't the issue--I meant the entire type, including the length.  You cannot have generic parameters depend on other generic parameters, which prevents doing anything like `Swizzle::<N>::swizzle<[0; N]>()`.  There is no way to create a function that has the array as a const generic parameter, since the length of the array is generic as well.

The reason this implementation works is because the array type is _effectively concrete_ once the trait is monomorphised, and we can then create regular consts (such as `Swizzle::INDEX`) with that concrete type.

crates/core_simd/src/swizzle.rs
```diff
@@ -0,0 +1,297 @@
...
+}
+
+/// Create a vector from the elements of another vector.
+pub trait Swizzle<const INPUT_LANES: usize, const OUTPUT_LANES: usize> {
+    /// Map from the lanes of the input vector to the output vector.
+    const INDEX: [usize; OUTPUT_LANES];
+
+    #[doc(hidden)]
+    /// The `simd_shuffle` intrinsic expects `u32`, so do error checking and conversion here.
+    const INDEX_IMPL: [u32; OUTPUT_LANES] = check::<INPUT_LANES, OUTPUT_LANES>(Self::INDEX);
```
programmerjake, line 20: this should be somehow protected, since a user can currently cause UB without `unsafe` by overriding `INDEX_IMPL`.

crates/core_simd/src/swizzle.rs
```diff
@@ -0,0 +1,297 @@
...
+}
+
+/// Create a vector from the elements of two other vectors.
+pub trait Swizzle2<const INPUT_LANES: usize, const OUTPUT_LANES: usize> {
+    /// Map from the lanes of the input vectors to the output vector
+    const INDEX: [Which; OUTPUT_LANES];
+
+    #[doc(hidden)]
+    /// The `simd_shuffle` intrinsic expects `u32`, so do error checking and conversion here.
+    const INDEX_IMPL: [u32; OUTPUT_LANES] = check2::<INPUT_LANES, OUTPUT_LANES>(Self::INDEX);
```
programmerjake, line 42: this also needs protection

crates/core_simd/src/swizzle.rs
calebzulawski: Actually, it wouldn't cause UB, the compiler would bail with a monomorphisation error if we'rw ok with that.

programmerjake: we could do something like:
```rust
mod private {
pub trait SwizzleConsts<const A: usize, const B: usize>: Swizzle<A, B> {
    const INDEX_IMPL: [...; ...];
}
}

pub trait Swizzle<const A: usize, const B: usize> {
    const INDEX: [...; ...];
    fn swizzle(self) -> ...
    where
        Self: private::SwizzleConsts<A, B>,
    {
        ...
    }
}

impl<T, const A: usize, const B: usize> private::SwizzleConsts<A, B> for T
where
    T: Swizzle<A, B>,
{
    ...
}
```

thomcc: Inside the stdlib this stuff often can be protected using `#[unstable]`, even if its a bit of a hack.

calebzulawski:
Thinking a little more about this--I still think this trait is necessary.  The concern about the ergonomics of a dummy struct is valid, but I think the ergonomics it provides are absolutely necessary:
* Error check lengths, panicking on overflowing the input vector, much more desirable than the modulo arithmetic LLVM provides.  It's really hard to debug your index calculation when your off-by-one error just turns into a rotation instead of a hard error.
* Using an enum to select between your input vectors in the case of a 2-vector swizzle. Manually calculating the offset into the second vector is a terrible API.

This simply isn't possible with fn generics.  Even if we extended const generics, you still won't be able to do:
```rust
fn shuffle<T, const LANES: usize, const INDEX: [usize; LANES]>(self) -> Self {
    const _ASSERT: () = our_assert_fn(INDEX);
}
```
for the same reason that you can't do:
```rust
fn foo<T>() {
    const FOO: Option<T> = None;
}
```
It's not a limitation of const generics, it's just generics and monomorphization in general.

I propose that we include this trait for now, which isn't particularly convenient to use (but really not that bad, either).  It's usable _today_, even on stable.  In the future, _if_ const generics are extended this way, we can then provide the following if we deem it necessary:
```rust
struct Shuffle<const LANES: usize, const INDEX: [usize; LANES]>;
struct Shuffle2<const LANES: usize, const INDEX: [Which; LANES]>;
```
This would allow you to write something like `Shuffle::<4, [4, 3, 2, 1]>::swizzle(x)`, which would cover the common cases.

thomcc:
> It's not a limitation of const generics, it's just generics and monomorphization in general.

I think there's a workaround for this that gets used in e.g. stdarch, see: https://github.com/rust-lang/stdarch/blob/master/crates/core_arch/src/macros.rs, which is used to statically assert things about const generic parameters.

crates/core_simd/src/swizzle.rs
calebzulawski: I was able to hide the impls with a helper trait like this

calebzulawski:
Hmm, that does look like it might be a good trick to get past the monomorphization problems, I'd amend my plan to say if we get the hypothetical const generics extension developed and stabilized before we need shuffle stabilized, we try that out.

programmerjake:
> ```rust
> struct Shuffle<const LANES: usize, const INDEX: [usize; LANES]>;
> struct Shuffle2<const LANES: usize, const INDEX: [Which; LANES]>;
> ```

Well, now that I think about it, wouldn't it be better to do something like:
```rust
pub struct Shuffle<const V: &[usize]>;
pub struct Shuffle2<const V: &[Which]>;
```
Which allows doing something like `Shuffle::<&[4, 3, 2, 1]>::swizzle(x)`

calebzulawski:
Putting aside that you can't use slices as const generic parameters right now, one of the downsides of that is it makes the type deduction on the return type pretty nasty.  Definitely worth considering as const generics progress, though.

calebzulawski:
I just added a `simd_shuffle` macro, which I believe handles the API everyone seems to be looking for in the common case.  It's a macro, not a function, but again it works on current rust.  I was thinking we could address this in a later PR but it seems to be a big enough priority that I just took care of it.  It has all of the ergonomics benefits this PR already had, but minimizes the boilerplate in the common case of a regular shuffle.  I still think we need to provide the full shuffle interface for more complex things (see the implementation of `rotate_{left,right}` as an example...), but this is much less verbose for common usage.

calebzulawski:
I think the 2-input-shuffle in the matrix inversion example shows how much more readable it is specifying which vector you are choosing from, especially with all the noise gone when using the macro.

crates/core_simd/src/swizzle.rs
```diff
@@ -0,0 +1,364 @@
...
+#[macro_export]
+macro_rules! simd_shuffle {
+    {
+        $vector:expr, $index:expr $(,)?
+    } => {
+        {
+            // FIXME this won't work when we are in `core`!
+            use $crate::Swizzle;
+            struct Shuffle;
+            impl Swizzle<{$index.len()}, {$index.len()}> for Shuffle {
```
programmerjake, line 42: this should be generic over input vector length

crates/core_simd/src/swizzle.rs
```diff
@@ -0,0 +1,364 @@
...
+        }
+    };
+    {
+        $first:expr, $second:expr, $index:expr $(,)?
+    } => {
+        {
+            // FIXME this won't work when we are in `core`!
+            use $crate::{Which, Swizzle2};
+            struct Shuffle;
+            impl Swizzle2<{$index.len()}, {$index.len()}> for Shuffle {
```
programmerjake, line 55: this should also be generic over input vector length

crates/core_simd/src/swizzle.rs
```diff
@@ -0,0 +1,364 @@
...
+trait SwizzleImpl<const INPUT_LANES: usize, const OUTPUT_LANES: usize> {
+    const INDEX_IMPL: [u32; OUTPUT_LANES];
+}
+
+impl<T, const INPUT_LANES: usize, const OUTPUT_LANES: usize> SwizzleImpl<INPUT_LANES, OUTPUT_LANES>
+    for T
+where
+    T: Swizzle<INPUT_LANES, OUTPUT_LANES> + ?Sized,
+{
+    const INDEX_IMPL: [u32; OUTPUT_LANES] = {
```
programmerjake, line 122: looks good!

workingjubilee:
cc @miguelraz, I am genuinely curious what he thinks of this sample since he ported the Matrix Inversion code.
both the overall diff:
https://github.com/rust-lang/portable-simd/pull/161/files/109b908c071dd082a1c23706f5951a6c2330e8b1
and especially the diff here, between the pre/post macro version:
https://github.com/rust-lang/portable-simd/pull/161/commits/109b908c071dd082a1c23706f5951a6c2330e8b1

Lokathor:
> const SHUFFLE01: [Which; 4] = [First(0), First(1), Second(0), Second(1)];

While I maybe understand, based on crusty knowledge about how the sse shuffles work, why this looks like this, it's very opaque.

calebzulawski:
> > const SHUFFLE01: [Which; 4] = [First(0), First(1), Second(0), Second(1)];
> 
> While I maybe understand, based on crusty knowledge about how the sse shuffles work, why this looks like this, it's very opaque.

Maybe I'm forgetting something, but SSE/AVX doesn't have anything like this interface.  I think you would have to do two shuffles and a blend.

I'm certainly opening to rewording etc, but `[0, 1, 5, 6]` seems much worse to me.  First of all, it's limited to use only with length-4 inputs, and secondly you need to _remember_ that the input is length 4 to even read it, and finally you need to actually do the modulo correctly in your head when reading it.  I definitely prefer if both vectors are 0-indexed and the length isn't encoded into the second vector index.

calebzulawski:
Proof to my point above, the equivalent would be `[0, 1, 4, 5]` and I screwed up the math 🤦

Lokathor:
i presumed it was based on https://doc.rust-lang.org/nightly/core/arch/x86/fn._mm_shuffle_ps.html, which shuffles two vectors into one output vector.

And the fact that you also kinda messed up the math i think shows that it's maybe very weird.

it might be fixable with sufficient macro power though

calebzulawski:
I suppose I'm confused what the problem is. I messed up the math because I screwed up the modulo in my head, which is why I'm suggesting we don't use the modulo representation.  It's very simple, `First(i)` takes `first[i]` and `Second(i)` takes `second[i]`.  Rather than the LLVM representation of `i` takes `if i < first.len() { first[I] } else { second[i - first.len()] }`

calebzulawski:
This is very different from the SSE instruction that takes half from the first vector and half from the second vector.  With this API you can do `[First(0), Second(0), First(0), Second(0)]` which can't be done with that SSE instruction.

Lokathor:
Ah, see that's the confusion I guess. It looks similar enough to sse shuffle that i tried to decode it with my sse brain.

calebzulawski:
I think it will look similar to all shuffles, which have their own quirks (neon's vtbl/vtbx are a little weird too).  I think this is the API with the least magic, you get exactly the lanes you ask for, and anything else will error.

crates/core_simd/src/swizzle.rs
calebzulawski: That would be desirable, but I don't see a way of extracting the const length from a non-const vector expression.  It's probably possible, so I'll have to think about it a bit more.

crates/core_simd/src/swizzle.rs
```diff
@@ -0,0 +1,364 @@
...
+            }
+            Shuffle::swizzle($vector)
+        }
+    };
+    {
+        $first:expr, $second:expr, $index:expr $(,)?
+    } => {
+        {
+            // FIXME this won't work when we are in `core`!
+            use $crate::{Which, Swizzle2};
```
workingjubilee, line 53:
```suggestion
            use $crate::simd::Swizzle;
            struct Shuffle;
            impl Swizzle<{$index.len()}, {$index.len()}> for Shuffle {
                const INDEX: [usize; {$index.len()}] = $index;
            }
            Shuffle::swizzle($vector)
        }
    };
    {
        $first:expr, $second:expr, $index:expr $(,)?
    } => {
        {
            use $crate::simd::{Which, Swizzle2};
```

calebzulawski: I just remembered my original issue.  `core_simd::simd` isn't public, so a user-accessible macro can't access it.  I'm just going to make it public.

crates/core_simd/src/swizzle.rs
calebzulawski: I realized I was overthinking it--the implementation was easy.  I changed the macro to `simd_swizzle` to reflect that, too.

programmerjake:
maybe it'd also be a good idea to rename the rustc `simd_shuffle` intrinsics to `simd_swizzle` for consistency's sake

programmerjake (approved):
lgtm

calebzulawski:
To note something that came up in rust-lang/rust#89167, this change also removes the requirement to use the `adt_const_params` feature, which is incomplete.

workingjubilee:
Finally properly reviewed this. There's a lot of duplication in our API and its appearance in the documentation but for some reason the `simd_swizzle!` macro looks like it's in the namespace twice. Do you have any idea why that is, and if it can be removed?

crates/core_simd/src/swizzle.rs
```diff
@@ -0,0 +1,374 @@
...
+}
+
+/// An index into one of two vectors.
+#[derive(Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum Which {
+    /// Indexes the first vector.
+    First(usize),
+    /// Indexes the second vector.
+    Second(usize),
+}
```
workingjubilee, line 80: It is traditional to name this type "Either", and its variants are named "Left" and "Right".
I am not a particularly devout adherent of tradition, however, so I am fine with whatever, at the moment.

crates/core_simd/src/swizzle.rs
```diff
@@ -0,0 +1,374 @@
+use crate::simd::intrinsics;
+use crate::{LaneCount, Simd, SimdElement, SupportedLaneCount};
+
+/// Rearrange vector elements.
+///
+/// A new vector is constructed by specifying the the lanes of the source vector or vectors to use.
```
workingjubilee, line 6: This should prefer active voice and possibly less repetition, e.g.
```suggestion
/// Constructs a new vector by selecting values from the lanes of the source vector or vectors to use.
```

crates/core_simd/src/swizzle.rs
```diff
@@ -0,0 +1,374 @@
...
+        T: SimdElement,
+        LaneCount<INPUT_LANES>: SupportedLaneCount,
+        LaneCount<OUTPUT_LANES>: SupportedLaneCount,
+    {
+        unsafe { intrinsics::simd_shuffle(first, second, Self::INDEX_IMPL) }
+    }
+}
+
+/// The `simd_shuffle` intrinsic expects `u32`, so do error checking and conversion here.
+trait SwizzleImpl<const INPUT_LANES: usize, const OUTPUT_LANES: usize> {
```
workingjubilee, line 123: Okay, I am actually kinda confused why we have both these and the other two Swizzle traits? "SwizzleImpl" isn't telling me much, as both traits are implemented. This is trying to say something about being used to define the indexing behavior.

crates/core_simd/src/swizzle.rs
```diff
@@ -0,0 +1,374 @@
...
+
+    /// Deinterleave two vectors.
+    ///
+    /// The first result takes every other lane of `self` and then `other`, starting with
+    /// the first lane.
+    ///
+    /// The second result takes every other lane of `self` and then `other`, starting with
+    /// the second lane.
+    ///
+    /// This particular permutation is efficient on many architectures.
```
workingjubilee, line 327: This is very vague, now that I look at it.
Is this about the architecture itself (as in x86 or aarch64) or the microarchitecture (as in Haswell vs. Nehalem)?

crates/core_simd/src/swizzle.rs
```diff
@@ -0,0 +1,374 @@
...
+            }
+            SwizzleImpl::swizzle($vector)
+        }
+    };
+    {
+        $first:expr, $second:expr, $index:expr $(,)?
+    } => {
+        {
+            use $crate::simd::{Which, Swizzle2};
+            struct SwizzleImpl;
```
workingjubilee, line 64: Here we define a struct that shares a name with a trait later...

workingjubilee (commented):
I have reviewed the code itself for correctness and I am relatively confident of it but it is hard to follow what is going on here in a few places. Additional commentary for *ourselves* would be useful, even if it might just seem like nattering on to users.

crates/core_simd/src/swizzle.rs
calebzulawski: Agreed, now that I look at it, it seems like this really should just be an `Either`.  Which isn't really the place of `core::simd` to add, but it's by far a superior API in this case.

crates/core_simd/src/swizzle.rs
calebzulawski: I'll try to improve the comments.  This trait is automatically implemented, it exists purely to create `INDEX_IMPL`, which is `u32` instead of `usize`, and may panic upon evaluation.

calebzulawski: (it's a separate trait to keep `INDEX_IMPL` completely hidden from the user)

crates/core_simd/src/swizzle.rs
calebzulawski: I think it's architecture--x86 has both generic shuffle and unpack, and aarch64 has unzip.  I think it's intentionally vague, as it's trying to just say "this shuffle is probably okay to use frequently".

crates/core_simd/src/swizzle.rs
workingjubilee: Normally in Rust we use Result for what is called Either, because in languages that have Either, they implicitly use it as a Result, but sometimes it really is Either. So I think it is harmless if we have `core::simd::Either`: it's specifically for this library, after all.

workingjubilee:
> Finally properly reviewed this. There's a lot of duplication in our API and its appearance in the documentation but for some reason the `simd_swizzle!` macro looks like it's in the namespace twice. Do you have any idea why that is, and if it can be removed?

Might be a `macro_use` combined with a `use module::*;` or something like that.
I don't think this is essential to fix, heh, so it's not at all a blocker, it's just something I have started paying attention to.

crates/core_simd/src/swizzle.rs
workingjubilee: Okay, we should probably say "architectures have specialized instructions for this" then.

workingjubilee:
Also re-reviewing again and I just want to affirm I will be happy with this when the internal documentation/commenting concerns are addressed so I can check things against what I thought things were for. ^^;

crates/core_simd/src/swizzle.rs
calebzulawski: Changed the name to make it a little less confusing

crates/core_simd/src/swizzle.rs
calebzulawski: I think maybe it's not quite correct to call it `Either`, since that would imply that it should be `Either<usize>`, I think.

calebzulawski:
Worth noting that this is unique to `core_simd`, and won't occur in `core` itself.

crates/core_simd/src/swizzle.rs
workingjubilee: That would be `Maybe<usize>`, no?
I am not of a strong opinion however, here.

PR #163, opened 2021-09-22
Feature-flag doc tests so they run for core

workingjubilee:
Checking to see if I can fix these "properly".


workingjubilee:
Everything but the known x86 bug passes, so I am calling it good.

PR #165, opened 2021-09-22
Hide mask impl details in sealed trait.

calebzulawski:


crates/core_simd/src/masks.rs
```diff
@@ -16,26 +16,36 @@ use crate::simd::{LaneCount, Simd, SimdElement, SupportedLaneCount};
...
 
-    #[doc(hidden)]
-    const FALSE: Self;
+        const FALSE: Self;
+    }
 }
+use sealed::Sealed;
+
+/// Marker trait for types that may be used as SIMD mask elements.
+pub unsafe trait MaskElement: SimdElement + Sealed {}
```
programmerjake, line 44: this doesn't stop users from accessing `Sealed`'s members

programmerjake:
Example:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=d86ff08097c08b734042d27f3d5a8279
```rust
pub mod api {
    pub trait PubAPI: Sealed {}
    use private::Sealed;
    mod private {
        pub trait Sealed {
            fn supposed_to_be_private(&self);
        }
    }
}

pub fn break_privacy<T: api::PubAPI>(v: T) {
    v.supposed_to_be_private();
}
```

```
Compiling playground v0.0.1 (/playground)
    Finished dev [unoptimized + debuginfo] target(s) in 1.57s
```

calebzulawski:
I think that's fine, it doesn't show up in docs and the trait will never be stabilized.

programmerjake:
> I think that's fine, it doesn't show up in docs and the trait will never be stabilized.

i'd expect it to probably still show up on ide auto-complete...we really need `pub(self)` trait members. that said, if it's marked perma-unstable and doc hidden (not just under the main feature gate), it'll probably be fine-ish.

programmerjake (approved):
assuming all `Sealed` members are perma-unstable, lgtm

workingjubilee:
Going to test this works against rust-lang/rust CI before I accept this.

PR #167, opened 2021-09-24
Add {gather,scatter}_select_unchecked

workingjubilee:
This unsafe variant allows the thinnest API, in case LLVM cannot
perform loop-invariant code motion on a hot loop when the safe
form is used.

An unchecked variant could be added to other forms, but doesn't
seem likely to improve anything, since it would just add heavier
codegen.

### DIFF
--- a/crates/core_simd/src/vector.rs
+++ b/crates/core_simd/src/vector.rs
@@ -210,7 +210,7 @@ where
     /// // If this mask was used to scatter, it would be unsound. Let's fix that.
     /// let mask = mask & idxs.lanes_lt(Simd::splat(vec.len()));
     ///
-    /// // We have masked the OOB lane, so it's safe to gather now.
+    /// // We have masked the OOB lane, so it's safe to scatter now.
     /// unsafe { vals.scatter_select_unchecked(&mut vec, mask, idxs); }
     /// // index 0's second write is masked, thus was omitted.
     /// assert_eq!(vec, vec![-41, 11, 12, 82, 14, 15, 16, 17, 18]);
--- a/crates/core_simd/src/vector.rs
+++ b/crates/core_simd/src/vector.rs
@@ -111,12 +111,42 @@ where
         idxs: Simd<usize, LANES>,
         or: Self,
     ) -> Self {
-        let mask = (mask & idxs.lanes_lt(Simd::splat(slice.len()))).to_int();
+        let mask: Mask<isize, LANES> = mask & idxs.lanes_lt(Simd::splat(slice.len()));
+        // SAFETY: We have masked-off out-of-bounds lanes.
+        unsafe { Self::gather_select_unchecked(slice, mask, idxs, or) }
+    }
+
+    /// Unsafe SIMD gather: construct a SIMD vector by reading from a slice, using potentially discontiguous indices.
+    /// Masked indices instead select the value from the "or" vector.
+    /// `gather_select_unchecked` is unsound if any unmasked index is out-of-bounds of the slice.
+    /// ```
+    /// # #![feature(portable_simd)]
+    /// # #[cfg(feature = "std")] use core_simd::{Simd, Mask};
+    /// # #[cfg(not(feature = "std"))] use core::simd::{Simd, Mask};
+    /// let vec: Vec<i32> = vec![10, 11, 12, 13, 14, 15, 16, 17, 18];
+    /// let idxs = Simd::from_array([9, 3, 0, 5]);
+    /// let alt = Simd::from_array([-5, -4, -3, -2]);
+    /// let mask = Mask::from_array([true, true, true, false]); // Note the final mask lane.
+    /// // If this mask was used to gather, it would be unsound. Let's fix that.
+    /// let mask = mask & idxs.lanes_lt(Simd::splat(vec.len()));
+    ///
+    /// // We have masked the OOB lane, so it's safe to gather now.
+    /// let result = unsafe { Simd::gather_select_unchecked(&vec, mask, idxs, alt) };
+    /// assert_eq!(result, Simd::from_array([-5, 13, 10, -2]));
+    /// ```
+    #[must_use]
+    #[inline]
+    pub unsafe fn gather_select_unchecked(
+        slice: &[T],
+        mask: Mask<isize, LANES>,
+        idxs: Simd<usize, LANES>,
+        or: Self,
+    ) -> Self {
         let base_ptr = crate::simd::ptr::SimdConstPtr::splat(slice.as_ptr());
         // Ferris forgive me, I have done pointer arithmetic here.
         let ptrs = base_ptr.wrapping_add(idxs);
         // SAFETY: The ptrs have been bounds-masked to prevent memory-unsafe reads insha'allah
-        unsafe { intrinsics::simd_gather(or, ptrs, mask) }
+        unsafe { intrinsics::simd_gather(or, ptrs, mask.to_int()) }
     }
 
     /// SIMD scatter: write a SIMD vector's values into a slice, using potentially discontiguous indices.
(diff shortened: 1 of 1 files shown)


crates/core_simd/src/vector.rs
```diff
@@ -111,12 +111,42 @@ where
         idxs: Simd<usize, LANES>,
         or: Self,
     ) -> Self {
-        let mask = (mask & idxs.lanes_lt(Simd::splat(slice.len()))).to_int();
+        let mask: Mask<isize, LANES> = mask & idxs.lanes_lt(Simd::splat(slice.len()));
+        // SAFETY: We have masked all the out-of-bounds lanes off.
+        unsafe { Self::gather_select_unchecked(slice, mask, idxs, or) }
+    }
+
+    /// Unsafe SIMD gather: construct a SIMD vector by reading from a slice, using potentially discontiguous indices.
```
calebzulawski, line 119: I would change this doc to something more like the rest of std (with a regular description of the function, and a Safety heading with the invariants below)

workingjubilee: Can you be more specific what function you are looking at as an example? std has... diverse standards of documentation in actuality.

calebzulawski: Just a random example: https://doc.rust-lang.org/std/boxed/struct.Box.html#method.from_raw

The beginning "Unsafe SIMD gather" seems repetitive, since that's pretty much what the function signature tells you.  I think the first line should basically say _what_ a gather does, and then go from there.  I think this and `gather_select` should basically have the same one line description, but this should then describe what must be done to use this correctly, but `gather_select` should instead say what causes it to panic.

calebzulawski: > Collect discontiguous slice elements into a vector.
>
> Copy elements from `slice`, indexed by `idxs`, into a vector, skipping any lanes disabled by `mask`.
>
> # Safety
> Calling this function with out-of-bounds, enabled indices is undefined behavior.

I'm thinking something like this format, but written more clearly

calebzulawski: we should probably go over our existing docs at some point and make sure they're consistent and make sense...

calebzulawski: fwiw, if you want to commit this now and deal with docs later, I'm fine with that too. the implementation looks good.

workingjubilee: I want to commit but I also want stuff in good condition. 😅
Docs should probably be a bar, going forward, and **especially** for `unsafe fn`.

crates/core_simd/src/vector.rs
```diff
@@ -160,12 +190,41 @@ where
...
+    /// # #[cfg(not(feature = "std"))] use core::simd::{Simd, Mask};
+    /// let mut vec: Vec<i32> = vec![10, 11, 12, 13, 14, 15, 16, 17, 18];
+    /// let idxs = Simd::from_array([9, 3, 0, 0]);
+    /// let vals = Simd::from_array([-27, 82, -41, 124]);
+    /// let mask = Mask::from_array([true, true, true, false]); // Note the mask of the last lane.
+    /// // If this mask was used to scatter, it would be unsound. Let's fix that.
+    /// let mask = mask & idxs.lanes_lt(Simd::splat(vec.len()));
+    ///
+    /// // We have masked the OOB lane, so it's safe to gather now.
+    /// vals.scatter_select(&mut vec, mask, idxs); // index 0's second write is masked, thus omitted.
```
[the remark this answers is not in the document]
workingjubilee: Whoops.

crates/core_simd/src/vector.rs
```diff
@@ -160,12 +190,42 @@ where
...
+    /// # #[cfg(feature = "std")] use core_simd::{Simd, Mask};
+    /// # #[cfg(not(feature = "std"))] use core::simd::{Simd, Mask};
+    /// let mut vec: Vec<i32> = vec![10, 11, 12, 13, 14, 15, 16, 17, 18];
+    /// let idxs = Simd::from_array([9, 3, 0, 0]);
+    /// let vals = Simd::from_array([-27, 82, -41, 124]);
+    /// let mask = Mask::from_array([true, true, true, false]); // Note the mask of the last lane.
+    /// // If this mask was used to scatter, it would be unsound. Let's fix that.
+    /// let mask = mask & idxs.lanes_lt(Simd::splat(vec.len()));
+    ///
+    /// // We have masked the OOB lane, so it's safe to gather now.
```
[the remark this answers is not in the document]
workingjubilee: gdi.

crates/core_simd/src/vector.rs
workingjubilee: Applied an attempt at this.

crates/core_simd/src/vector.rs
```diff
@@ -111,17 +117,57 @@ where
...
         // SAFETY: The ptrs have been bounds-masked to prevent memory-unsafe reads insha'allah
-        unsafe { intrinsics::simd_gather(or, ptrs, mask) }
+        unsafe { intrinsics::simd_gather(or, ptrs, mask.to_int()) }
     }
 
-    /// SIMD scatter: write a SIMD vector's values into a slice, using potentially discontiguous indices.
-    /// Out-of-bounds indices are not written.
-    /// `scatter` writes "in order", so if an index receives two writes, only the last is guaranteed.
+    /// Writes the values in a SIMD vector to  potentially discontiguous indices in `slice`.
+    /// Lanes masked `false` or given an out-of-bounds index are not written.
```
calebzulawski, line 166:
```suggestion
    /// If an index is out-of-bounds or the mask is `false`, the lane is not written.
```

crates/core_simd/src/vector.rs
```diff
@@ -111,17 +117,57 @@ where
...
+    /// If two unmasked lanes in the scattered vector would write to the same index
```
calebzulawski, line 167:
```suggestion
    /// If two lanes in the scattered vector would write to the same index
```
I don't think it's necessary to specify unmasked, since that's the only way it could write, anyway.

crates/core_simd/src/vector.rs
```diff
@@ -138,9 +184,12 @@ where
         self.scatter_select(slice, Mask::splat(true), idxs)
     }
 
-    /// SIMD scatter: write a SIMD vector's values into a slice, using potentially discontiguous indices.
-    /// Out-of-bounds or masked indices are not written.
-    /// `scatter_select` writes "in order", so if an index receives two writes, only the last is guaranteed.
+    /// Writes the values in a SIMD vector to multiple potentially discontiguous indices in `slice`.
+    /// Lanes masked `false` or given an out-of-bounds index are not written.
```
calebzulawski, line 188:
```suggestion
    /// If an index is out-of-bounds or the mask is `false`, the lane is not written.
```

crates/core_simd/src/vector.rs
```diff
@@ -138,9 +184,12 @@ where
...
+    /// If two unmasked lanes in the scattered vector would write to the same index,
```
calebzulawski, line 189:
```suggestion
    /// If two lanes in the scattered vector would write to the same index,
```
Ditto above

crates/core_simd/src/vector.rs
```diff
@@ -160,12 +209,50 @@ where
...
     ) {
-        // We must construct our scatter mask before we derive a pointer!
-        let mask = (mask & idxs.lanes_lt(Simd::splat(slice.len()))).to_int();
+        let mask: Mask<isize, LANES> = mask & idxs.lanes_lt(Simd::splat(slice.len()));
+        // SAFETY: We have masked-off out-of-bounds lanes.
+        unsafe { self.scatter_select_unchecked(slice, mask, idxs) }
+    }
+
+    /// Writes the values in a SIMD vector to multiple potentially discontiguous indices in `slice`.
+    /// Lanes masked `false` are not written.
```
calebzulawski, line 218:
```suggestion
    /// If a mask is `false`, the lane is not written.
```

crates/core_simd/src/vector.rs
```diff
@@ -160,12 +209,50 @@ where
...
+    /// If two unmasked lanes in the scattered vector would write to the same index,
```
calebzulawski, line 219:
```suggestion
    /// If two lanes in the scattered vector would write to the same index,
```
Ditto above

calebzulawski (approved):
Looks good, I just added some suggestions to make the language a little simpler.

crates/core_simd/src/vector.rs
workingjubilee: Hmm. So I think some of these comments are good improvements and I would probably just adopt them directly, but this is revealing a kinda mental-view difference here that I want to dig into a bit:

I don't want to talk about "mask" for individual lanes because I think that's confusing: the way I am looking at it, the overall "data structure" is a `Mask` and it functionally has "lanes" of its own, even if they are a bit opaque or are only 1-bit-wide. So a `Mask` has multiple `true` and `false` values in it, and thus talking about it being either value _per se_ is... odd, to me.

Notionally:
`self` has lanes containing numeric values.
`mask` has lanes containing go/no-go values.
`idxs` has lanes containing indices.
The controlling structure is `{ value, index } & bool`, which one could call "masked lanes" or "masked indices", depending on how you look at it.

calebzulawski: I am now thinking, should we perhaps instead call the mask variable `enabled` or similar?  That encodes the write-on-`true` nature in the name better than simple `mask`.

workingjubilee: `enable`, as Rust is an imperative language. 👑

PR #171, opened 2021-09-29
Add lanes()

### LINKED ISSUE #151: Missing function for number of lanes
One of my projects is currently using `packed_simd`, and uses a compile-time feature flag to choose a vector size. I tried porting this to `stdsimd`, which worked perfectly, apart from a missing function, to get the current number of lanes for a given simd type. `packed_simd` has the const fn [`lanes`](https://rust-lang.github.io/packed_simd/packed_simd_2/struct.Simd.html#method.lanes) which I used in this context. It would be great to have this also exposed for all types in this crate, and it should be rather trivial to implement.

Edit: It seems the trait item T::LANES is available, but this is awkward to find  in the documentation, and requires the `Vector` trait to be in scope. I am not sure whether adding the `lanes()` function is worth it for the ease of documentation.

mulimoen:
Adds the lanes function for easily getting the number of lanes of a SIMD element

Fixes #151

### DIFF
--- a/crates/core_simd/src/vector.rs
+++ b/crates/core_simd/src/vector.rs
@@ -24,6 +24,9 @@ where
     LaneCount<LANES>: SupportedLaneCount,
     T: SimdElement,
 {
+    /// Number of lanes in this vector.
+    pub const LANES: usize = LANES;
+
     /// Get the number of lanes in this vector.
     pub const fn lanes(&self) -> usize {
         LANES
--- a/crates/core_simd/src/vector.rs
+++ b/crates/core_simd/src/vector.rs
@@ -24,6 +24,11 @@ where
     LaneCount<LANES>: SupportedLaneCount,
     T: SimdElement,
 {
+    /// Get the number of lanes in this vector.
+    pub const fn lanes(&self) -> usize {
+        LANES
+    }
+
     /// Construct a SIMD vector by setting all lanes to the given value.
     pub const fn splat(value: T) -> Self {
         Self([value; LANES])


calebzulawski:
I wonder if it makes more sense to do this or have a const associated with the struct.  This makes it easier to pull from an instance of a vector, of course.

Lokathor:
could have both, but associated consts are "weirder" to use, so if there's just one then this should be it.

mulimoen:
I'd say it would be best to include both to enable the most flexibility for users

crates/core_simd/src/vector.rs
```diff
@@ -24,6 +24,13 @@ where
     LaneCount<LANES>: SupportedLaneCount,
     T: SimdElement,
 {
+    const LANES: usize = LANES;
```
calebzulawski, line 27: Missing `pub` and documentation

mulimoen: Added `pub` and a comment

workingjubilee:
So I don't want to bikeshed excessively, but should this be `len`, in sync with e.g. Vec, slice, and other containers?

calebzulawski:
Yeah, it may make more sense calling it `len`.  It's worth noting that arrays don't have a `len` function at all, you need to use `as_slice` or similar.

Lokathor:
well if you call array.len you'll fail the method lookup, automatically deref to a slice, and then the slice will have it. So *effectively* arrays have this.

But i think lanes makes sense because generally one should not be "looking inside" a simd vector horizontally that much

mulimoen:
I don't feel strongly for `len` over `lanes`, but `lanes` might be more intuitive as the documentation is heavy on the use of `LANES`.

workingjubilee:
Fair enough. If we change it we should change it elsewhere, then.

PR #172, opened 2021-09-29
Renovate for Edition 2021

workingjubilee:
In a still-future edition, `unsafe_op_in_unsafe_fn` may error.
Let's get ahead of that.

### DIFF
--- a/crates/core_simd/Cargo.toml
+++ b/crates/core_simd/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "core_simd"
 version = "0.1.0"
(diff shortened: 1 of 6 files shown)


Lokathor:
Well it should never hard error, just deny by default.

But this is seemingly a small enough amount of unsafe I guess it's fine.

workingjubilee:
Yeah, this adds a little bit of noise but it's good because it shows where our unsafe API is pretty directly exposing internal unsafety.

calebzulawski (approved):

PR #175, opened 2021-10-18
Delete travis config, move tests to github actions.

### LINKED ISSUE #92: Failing wasm tests

### LINKED ISSUE #174: Move remaining arch tests off TravisCI
Travis has become totally unsuitable.

calebzulawski:
Fixes #174.  Incidentally fixes #92 too, by reenabling wasm tests.


programmerjake:
you also disabled powerpc64le, which I expect probably still works. powerpc64 was disabled by cross before v0.2.1 was released in this commit: https://github.com/rust-embedded/cross/commit/e05d6a6b9f75265a4e0ad4b33e978b3ce3d26021

calebzulawski:
So a summary, the ppc64 errors look like a bug in QEMU: https://rust-lang.zulipchat.com/#narrow/stream/257879-project-portable-simd

The remaining powerpc test works, but it appears to time out.  It's exceptionally slow compared to the other cross targets.

PR #181, opened 2021-11-06
rotate_{left,right} -> rotate_lanes_{left,right}

### LINKED ISSUE #180: The meaning of rotate_left/right
Switching a project from `packed_simd_2` to `simd_core`, I noticed that [`rotate_left`](https://rust-lang.github.io/portable-simd/core_simd/simd/struct.Simd.html#method.rotate_left) and [`rotate_right`](https://rust-lang.github.io/portable-simd/core_simd/simd/struct.Simd.html#method.rotate_right) have [become](https://docs.rs/packed_simd_2/0.3.6/packed_simd_2/type.u32x4.html#method.rotate_left) horizontal operations, borrowing from [`std::slice`](https://doc.rust-lang.org/std/primitive.slice.html#method.rotate_left).

I believe this to be inconsistent and surprising

- since I consider SIMD types arithmetic types first,
- since `rotate_left` is (conceptually) just `(x << n) | (x >> (BITS - n))` and these constituent operations remain vertical,
- since [`rotate_left`](https://doc.rust-lang.org/std/primitive.u32.html#method.rotate_left)/[`right`](https://doc.rust-lang.org/std/primitive.u32.html#method.rotate_right) already exist in `std` as (trivially) vertical operations.

Also, there is precedent for `horizontal_` operations but not for explicitly vertical ones.

Therefore, whether or not `simd_core` ends up providing vertical rotation, I propose renaming the methods in question to something like `horizontal_rotate_left`.

calebzulawski:
Fixes #180.
cc @nixberg

### DIFF
--- a/crates/core_simd/examples/matrix_inversion.rs
+++ b/crates/core_simd/examples/matrix_inversion.rs
@@ -180,58 +180,58 @@ pub fn simd_inv4x4(m: Matrix4x4) -> Option<Matrix4x4> {
     let row2 = simd_swizzle!(tmp, row3, SHUFFLE02);
     let row3 = simd_swizzle!(row3, tmp, SHUFFLE13);
 
-    let tmp = (row2 * row3).reverse().rotate_right::<2>();
+    let tmp = (row2 * row3).reverse().rotate_lanes_right::<2>();
     let minor0 = row1 * tmp;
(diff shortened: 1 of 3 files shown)


programmerjake (approved):
lgtm after somewhat cursory examination

PR #182, opened 2021-11-07
Document internal unsafety

### LINKED ISSUE #173: Document internal unsafety
Currently, we document public `unsafe fn` and their invariants.

However, as a highly `unsafe` crate, we should also document internally that we are maintaining invariants for `unsafe` blocks internally. That said, we are also so `unsafe` because almost everything is a thin wainscotting over `extern "platform-intrinsic"` calls. Those are **not** the focus of this issue. Rather, what we really want to document is anything that **isn't** just a straight pass-through to an internal `unsafe fn` intrinsic. These comments need not be publically exposed.

A few of these were surfaced in https://github.com/rust-lang/portable-simd/pull/172/files. Mostly, it's things like our internal implementation of existing trait-level `unsafe fn`. Also we should probably document the safety invariants on the `impl` too, not just on the trait.

calebzulawski:
Fixes #173.

I found an ICE in clippy along the way, so we shouldn't actually merge this until https://github.com/rust-lang/rust-clippy/pull/7945 is upstream.

### DIFF
--- a/crates/core_simd/src/comparisons.rs
+++ b/crates/core_simd/src/comparisons.rs
@@ -9,12 +9,16 @@ where
     /// Test if each lane is equal to the corresponding lane in `other`.
     #[inline]
     pub fn lanes_eq(self, other: Self) -> Mask<T::Mask, LANES> {
+        // Safety: `self` is a vector, and the result of the comparison
+        // is always a valid mask.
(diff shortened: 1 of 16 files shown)


programmerjake (approved):
lgtm

calebzulawski:
I think we're going to have to put this off a bit more--clippy doesn't seem to recognize safety comments inside macros any longer (and of course nearly everything is in macros for now)

PR #183, opened 2021-11-10
Configure CI to bite on warnings

workingjubilee:


calebzulawski:
Ah, that's how that error happened in the other PR. Lgtm

workingjubilee:
That's fine, I don't actually want to test the later features at the moment.

programmerjake:
try:
```bash
lscpu | grep -o ' avx512[^ ]*' | sort --unique | sed -z 's/\n* \([^\n]*\)/+\1,/g; s/\(.*\),\n$/\1/'
```

you *do* need to match the whole cpu feature name, otherwise rustc will rightly complain that it isn't recognized

workingjubilee:
I can't read that.

programmerjake:
> I can't read that.

any better?

workingjubilee:
Eh?

I also don't want to test `avx512_fp16`, by design. I am only interested in some AVX512 features at the moment (that is: the regex **deliberately** under-matches), and I am not persuaded any later ones are good ideas. I will rewrite this entire feature detection script in Rust if I have to do anything more complicated than this.

workingjubilee:
After discussion, we realized belatedly this just duplicates running `cargo check` with a `native` target. 😅 
No regrets though, I think we can use a more sophisticated conditional model for our CI matrix with these kinds of tricks.

PR #185, opened 2021-11-11
Restrict Arm types to Arm v7+

workingjubilee:
This mostly mirrors the restrictions in std::arch.
It can be loosened slightly with later refactoring.

### DIFF
--- a/crates/core_simd/src/vendor.rs
+++ b/crates/core_simd/src/vendor.rs
@@ -20,7 +20,10 @@ mod x86;
 #[cfg(any(target_arch = "wasm32"))]
 mod wasm32;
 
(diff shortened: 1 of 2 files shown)


calebzulawski:
I guess this is an unfortunate limitation of rustc, since you should be able to write neon in older arm binaries and do runtime dispatch.  Looks fine to me, though. (Looks like fmt failed)

calebzulawski (approved):

programmerjake (approved):
lgtm, I'm assuming you got the set of required `cfg` flags correct

workingjubilee:
Unfortunately, making all of this valid for Arm architectures prior to Arm v7 seems likely to require edits to `std::arch`'s definitions. We will want to add a check build for cross-compiling to various Arm targets, it seems like.

PR #189, opened 2021-11-13
Use intrinsics for Mask::{to,from}_array

### LINKED ISSUE #184: poor performance for Mask::from_array
The existing code copies/converts one bool at a time, unvectorized by llvm.

Suggested fix:
https://rust-lang.zulipchat.com/#narrow/stream/257879-project-portable-simd/topic/is.20it.20mature.20to.20switch.20to.20portable-simd.3F/near/261100448

https://rust.godbolt.org/z/GdTx63axx

```rust
#![feature(repr_simd)]
#![feature(platform_intrinsics)]

extern "platform-intrinsic" {
    fn simd_ne<T, U>(a: T, b: T) -> U;
    fn simd_cast<T, U>(v: T) -> U;
}

#[repr(simd)]
pub struct Simd<T, const N: usize>(pub [T; N]);

pub fn from_array_new(v: [bool; 16]) -> Simd<i32, 16> {
    unsafe {
        let v: [u8; 16] = std::mem::transmute(v);
        let bools: Simd<i8, 16> = simd_ne(Simd(v), Simd([0u8; 16]));
        simd_cast(bools)
    }
}

pub fn from_array_current(v: [bool; 16]) -> Simd<i32, 16> {
    let mut retval = Simd([0; 16]);
    for i in 0..16 {
        retval.0[i] = if v[i] { -1 } else { 0 };
    }
    retval
}
```

workingjubilee:
This significantly simplifies codegen and should improve mask perf.
Thanks to @programmerjake for the example for `from_array`!
This fixes #184.

### DIFF
--- a/crates/core_simd/src/masks.rs
+++ b/crates/core_simd/src/masks.rs
@@ -12,9 +12,10 @@
 )]
 mod mask_impl;
 
+use crate::simd::intrinsics;
 use crate::simd::{LaneCount, Simd, SimdElement, SupportedLaneCount};
 use core::cmp::Ordering;
-use core::fmt;
+use core::{fmt, mem};
 
 mod sealed {
     use super::*;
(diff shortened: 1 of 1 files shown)


workingjubilee:
This version should work except that I believe it requires `simd_cast` to not vomit on `Simd<isize, _>`.

programmerjake (approved):
lgtm, once `simd_cast` is fixed

crates/core_simd/src/masks.rs
```diff
@@ -105,22 +106,39 @@ where
...
     pub fn from_array(array: [bool; LANES]) -> Self {
-        let mut vector = Self::splat(false);
-        for (i, v) in array.iter().enumerate() {
-            vector.set(i, *v);
+        // SAFETY: Rust's bool has a layout of 1 byte (u8) with a value of
+        //     true:    0b_0000_0001
+        //     false:   0b_0000_0000
+        // Thus, an array of bools is also a valid array of bytes: [u8; N]
+        // This would be hypothetically valid as an "in-place" transmute,
+        // but these are "dependently-sized" types, so copy elision it is!
```
bjorn3, line 116: `transmute_copy` doesn't check if the input and output types have the same size.

programmerjake: yup, but we know they're the same size since the input type is `[bool; LANES]` and the output type is `[u8; LANES]` and Rust guarantees that `bool` is a single byte.

crates/core_simd/src/masks.rs
```diff
@@ -105,22 +106,39 @@ where
...
+        unsafe {
+            let bytes: [u8; LANES] = mem::transmute_copy(&array);
+            let bools: Simd<i8, LANES> =
+                intrinsics::simd_ne(Simd::from_array(bytes), Simd::splat(0u8));
```
bjorn3, line 120: Why is this necessary? Can't this be just `Simd::from_array(bytes)`?

workingjubilee: Exactly. It could also be `simd_neg`, come to think of it.

programmerjake: `simd_ne` is best because llvm knows the result is supposed to be a mask, with neg llvm doesn't know that, so will likely optimize more poorly

workingjubilee:
Now that https://github.com/rust-lang/rust/pull/92425 is merged, that should allow this to land. Just have to wait for nightly to roll over.

PR #190, opened 2021-11-14
Update CONTRIBUTING.md for the fact that Travis is no longer used

alex:


calebzulawski:
Thanks! Good catch

alex:
Thanks for making `std::simd` happen!

PR #195, opened 2021-11-19
Generic `core::ops` for `Simd<T, _>`

workingjubilee:
In order to maintain type soundness, we need to be sure we only implement an operation for `Simd<T, _> where T: SimdElement`... and also valid for that operation in general. While we could do this purely parametrically, it is more sound to implement the operators directly for the base scalar type arguments and then use type parameters to extend the operators to the "higher order" operations.

This implements that strategy and cleans up `simd::ops` into a few submodules:
- assign.rs: `core::ops::*Assign`
- deref.rs:  `core::ops` impls which "deref" borrowed versions of the arguments
- unary.rs: encloses the logic for unary operators on `Simd`, as unary ops are much simpler

This is possible since everything need not be nested in a single maze of macros anymore. The result simplifies the logic and allows reasoning about what operators are valid based on the expressed trait bounds, and also reduces the size of the trait implementation output in rustdoc, for a huge win of 4 MB off the size of `struct.Simd.html`! This addresses a common user complaint, as the original was over 5.5 MB and capable of crashing browsers!

This also carries a fix for a type-inference-related breakage, by removing the autosplatting (vector + scalar binop) impls, as unfortunately the presence of autosplatting was capable of busting type inference. We will likely need to see results from a Crater run before we can understand how to re-land autosplatting.

### DIFF
--- a/crates/core_simd/examples/nbody.rs
+++ b/crates/core_simd/examples/nbody.rs
@@ -97,7 +97,7 @@ mod nbody {
         let sun = &mut sun[0];
         for body in rest {
             let m_ratio = body.mass / SOLAR_MASS;
-            sun.v -= body.v * m_ratio;
+            sun.v -= body.v * Simd::splat(m_ratio);
         }
     }
 
@@ -143,14 +143,14 @@ mod nbody {
         let mut i = 0;
         for j in 0..N_BODIES {
             for k in j + 1..N_BODIES {
-                let f = r[i] * mag[i];
-                bodies[j].v -= f * bodies[k].mass;
-                bodies[k].v += f * bodies[j].mass;
+                let f = r[i] * Simd::splat(mag[i]);
+                bodies[j].v -= f * Simd::splat(bodies[k].mass);
+                bodies[k].v += f * Simd::splat(bodies[j].mass);
                 i += 1
             }
         }
         for body in bodies {
-            body.x += dt * body.v
+            body.x += Simd::splat(dt) * body.v
         }
     }
 
--- a/crates/core_simd/src/ops/assign.rs
+++ b/crates/core_simd/src/ops/assign.rs
@@ -1,0 +1,124 @@
+//! Assignment operators
+use super::*;
+use core::ops::{AddAssign, MulAssign}; // commutative binary op-assignment
+use core::ops::{BitAndAssign, BitOrAssign, BitXorAssign}; // commutative bit binary op-assignment
+use core::ops::{DivAssign, RemAssign, SubAssign}; // non-commutative binary op-assignment
+use core::ops::{ShlAssign, ShrAssign}; // non-commutative bit binary op-assignment
+
+// Arithmetic
+
+macro_rules! assign_ops {
+    ($(impl<T, U, const LANES: usize> $assignTrait:ident<U> for Simd<T, LANES>
+        where
+            Self: $trait:ident,
+        {
+            fn $assign_call:ident(rhs: U) {
+                $call:ident
+            }
+        })*) => {
+        $(impl<T, U, const LANES: usize> $assignTrait<U> for Simd<T, LANES>
+        where
+            Self: $trait<U, Output = Self>,
+            T: SimdElement,
+            LaneCount<LANES>: SupportedLaneCount,
+        {
+            #[inline]
+            fn $assign_call(&mut self, rhs: U) {
+                *self = self.$call(rhs);
+            }
+        })*
+    }
+}
+
+assign_ops! {
+    // Arithmetic
+    impl<T, U, const LANES: usize> AddAssign<U> for Simd<T, LANES>
+    where
+        Self: Add,
+    {
+        fn add_assign(rhs: U) {
+            add
+        }
+    }
+
+    impl<T, U, const LANES: usize> MulAssign<U> for Simd<T, LANES>
+    where
+        Self: Mul,
+    {
+        fn mul_assign(rhs: U) {
+            mul
+        }
+    }
+
+    impl<T, U, const LANES: usize> SubAssign<U> for Simd<T, LANES>
+    where
+        Self: Sub,
+    {
+        fn sub_assign(rhs: U) {
+            sub
+        }
+    }
+
+    impl<T, U, const LANES: usize> DivAssign<U> for Simd<T, LANES>
+    where
+        Self: Div,
+    {
+        fn div_assign(rhs: U) {
+            div
+        }
+    }
+    impl<T, U, const LANES: usize> RemAssign<U> for Simd<T, LANES>
+    where
+        Self: Rem,
+    {
+        fn rem_assign(rhs: U) {
+            rem
+        }
+    }
+
+    // Bitops
+    impl<T, U, const LANES: usize> BitAndAssign<U> for Simd<T, LANES>
+    where
+        Self: BitAnd,
+    {
+        fn bitand_assign(rhs: U) {
+            bitand
+        }
+    }
+
+    impl<T, U, const LANES: usize> BitOrAssign<U> for Simd<T, LANES>
+    where
+        Self: BitOr,
+    {
+        fn bitor_assign(rhs: U) {
+            bitor
+        }
+    }
+
+    impl<T, U, const LANES: usize> BitXorAssign<U> for Simd<T, LANES>
+    where
+        Self: BitXor,
+    {
+        fn bitxor_assign(rhs: U) {
+            bitxor
+        }
+    }
+
+    impl<T, U, const LANES: usize> ShlAssign<U> for Simd<T, LANES>
+    where
+        Self: Shl,
+    {
+        fn shl_assign(rhs: U) {
+            shl
+        }
+    }
+
+    impl<T, U, const LANES: usize> ShrAssign<U> for Simd<T, LANES>
+    where
+        Self: Shr,
+    {
+        fn shr_assign(rhs: U) {
+            shr
+        }
+    }
+}
--- a/crates/core_simd/src/ops/deref.rs
+++ b/crates/core_simd/src/ops/deref.rs
@@ -1,70 +1,124 @@
 //! This module hacks in "implicit deref" for Simd's operators.
 //! Ideally, Rust would take care of this itself,
 //! and method calls usually handle the LHS implicitly.
-//! So, we'll manually deref the RHS.
+//! But this is not the case with arithmetic ops.
 use super::*;
 
-macro_rules! deref_ops {
-    ($(impl<T, const LANES: usize> $trait:ident<&Self> for Simd<T, LANES> {
-            fn $call:ident(rhs: &Self)
-        })*) => {
-        $(impl<T, const LANES: usize> $trait<&Self> for Simd<T, LANES>
+macro_rules! deref_lhs {
+    (impl<T, const LANES: usize> $trait:ident for $simd:ty {
+            fn $call:ident
+        }) => {
+        impl<T, const LANES: usize> $trait<$simd> for &$simd
         where
-            Self: $trait<Self, Output = Self>,
             T: SimdElement,
+            $simd: $trait<$simd, Output = $simd>,
             LaneCount<LANES>: SupportedLaneCount,
         {
-            type Output = Self;
+            type Output = Simd<T, LANES>;
 
             #[inline]
             #[must_use = "operator returns a new vector without mutating the inputs"]
-            fn $call(self, rhs: &Self) -> Self::Output {
+            fn $call(self, rhs: $simd) -> Self::Output {
+                (*self).$call(rhs)
+            }
+        }
+    };
+}
+
+macro_rules! deref_rhs {
+    (impl<T, const LANES: usize> $trait:ident for $simd:ty {
+            fn $call:ident
+        }) => {
+        impl<T, const LANES: usize> $trait<&$simd> for $simd
+        where
+            T: SimdElement,
+            $simd: $trait<$simd, Output = $simd>,
+            LaneCount<LANES>: SupportedLaneCount,
+        {
+            type Output = Simd<T, LANES>;
+
+            #[inline]
+            #[must_use = "operator returns a new vector without mutating the inputs"]
+            fn $call(self, rhs: &$simd) -> Self::Output {
                 self.$call(*rhs)
             }
-        })*
+        }
+    };
+}
+
+macro_rules! deref_ops {
+    ($(impl<T, const LANES: usize> $trait:ident for $simd:ty {
+            fn $call:ident
+        })*) => {
+        $(
+            deref_rhs! {
+                impl<T, const LANES: usize> $trait for $simd {
+                    fn $call
+                }
+            }
+            deref_lhs! {
+                impl<T, const LANES: usize> $trait for $simd {
+                    fn $call
+                }
+            }
+            impl<'lhs, 'rhs, T, const LANES: usize> $trait<&'rhs $simd> for &'lhs $simd
+            where
+                T: SimdElement,
+                $simd: $trait<$simd, Output = $simd>,
+                LaneCount<LANES>: SupportedLaneCount,
+            {
+                type Output = $simd;
+
+                #[inline]
+                #[must_use = "operator returns a new vector without mutating the inputs"]
+                fn $call(self, rhs: &$simd) -> Self::Output {
+                    (*self).$call(*rhs)
+                }
+            }
+        )*
     }
 }
 
 deref_ops! {
     // Arithmetic
-    impl<T, const LANES: usize> Add<&Self> for Simd<T, LANES> {
-        fn add(rhs: &Self)
+    impl<T, const LANES: usize> Add for Simd<T, LANES> {
+        fn add
     }
 
-    impl<T, const LANES: usize> Mul<&Self> for Simd<T, LANES> {
-        fn mul(rhs: &Self)
+    impl<T, const LANES: usize> Mul for Simd<T, LANES> {
+        fn mul
     }
 
-    impl<T, const LANES: usize> Sub<&Self> for Simd<T, LANES> {
-        fn sub(rhs: &Self)
+    impl<T, const LANES: usize> Sub for Simd<T, LANES> {
+        fn sub
     }
 
-    impl<T, const LANES: usize> Div<&Self> for Simd<T, LANES> {
-        fn div(rhs: &Self)
+    impl<T, const LANES: usize> Div for Simd<T, LANES> {
+        fn div
     }
 
-    impl<T, const LANES: usize> Rem<&Self> for Simd<T, LANES> {
-        fn rem(rhs: &Self)
+    impl<T, const LANES: usize> Rem for Simd<T, LANES> {
+        fn rem
     }
 
     // Bitops
-    impl<T, const LANES: usize> BitAnd<&Self> for Simd<T, LANES> {
-        fn bitand(rhs: &Self)
+    impl<T, const LANES: usize> BitAnd for Simd<T, LANES> {
+        fn bitand
     }
 
-    impl<T, const LANES: usize> BitOr<&Self> for Simd<T, LANES> {
-        fn bitor(rhs: &Self)
+    impl<T, const LANES: usize> BitOr for Simd<T, LANES> {
+        fn bitor
     }
 
-    impl<T, const LANES: usize> BitXor<&Self> for Simd<T, LANES> {
-        fn bitxor(rhs: &Self)
+    impl<T, const LANES: usize> BitXor for Simd<T, LANES> {
+        fn bitxor
     }
 
-    impl<T, const LANES: usize> Shl<&Self> for Simd<T, LANES> {
-        fn shl(rhs: &Self)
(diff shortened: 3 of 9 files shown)


workingjubilee:
We discussed this and decided
```rust
impl<T, const LANES: usize> BitAnd for Simd<T, LANES>
where
    T: SimdElement + BitAnd<Output = T>,
{
    type Output = Self;

    #[inline]
    fn bitand(self, rhs: Self) -> Self::Output {
        unsafe { intrinsics::simd_and(self, rhs) }
    }
}
```
is a bridge too far, but
```rust
impl<T, const LANES: usize> BitAnd<T> for Simd<T, LANES>
where
    T: SimdElement + BitAnd<Output = T>,
{
    type Output = Self;

    #[inline]
    fn bitand(self, rhs: T) -> Self::Output {
        unsafe { intrinsics::simd_and(self, Simd::splat(rhs)) }
    }
}
```
is fine, since if a vector can be added or whatever then we also want the implied promotion for binary ops.

calebzulawski:
I don't think the second is fine... I was suggesting something more like this, since it uses the explicitly defined operation.
```rust
impl<T, const LANES: usize> BitAnd<T> for Simd<T, LANES>
where
    T: SimdElement,
    Self: BitAnd,
{
    type Output = Self;

    #[inline]
    fn bitand(self, rhs: T) -> Self::Output {
        self & Simd::splat(rhs)
    }
}
```

workingjubilee:
Ah, I wasn't actually paying that close attention to the trait bounds because I am adjusting the way all of those are written to fit the new revision anyways.

pro465:
@calebzulawski 

> ```rust
> impl<T, const LANES: usize> BitAnd<T> for Simd<T, LANES>
> where
>     T: SimdElement,
>     Self: BitAnd,
>     LaneCount<LANES>: SupportedLaneCount,
> {
>     type Output = Self;
> 
>     #[inline]
>     fn bitand(self, rhs: T) -> Self::Output {
>         self & Simd::splat(rhs)
>     }
> }
> ```

this almost seems like a pythony assumption-filled thing used mostly for convinience. almost like the Rust team suddenly diciding to implement `fn iter<'a>(&'a self) -> GraphemeClusters<'a>` for strings

programmerjake:
@pro465 
> @calebzulawski
> 
> this almost seems like a pythony assumption-filled thing used mostly for convinience. almost like the Rust team suddenly diciding to implement `fn iter<'a>(&'a self) -> GraphemeClusters<'a>` for strings

See also the discussion on zulip about removing splatting from operators for simd: https://rust-lang.zulipchat.com/#narrow/stream/257879-project-portable-simd/topic/remove.20splatting.20on.20operators.3F

workingjubilee:
@pro465 Being able to write in a memory safe language instead of raw assembly is **also** a convenience. "It is a convenience" is neither useful nor actionable feedback.

Lokathor:
(other than the name of the function being too plain, the standard library *should* offer a way to iterate grapheme clusters of a &str, actually.)

pro465:
@workingjubilee 
> @pro465 Being able to write in a memory safe language instead of raw assembly is **also** a convenience. "It is a convenience" is neither useful nor actionable feedback.

sorry, i did not mean to come off as rude, but at least we can agree that your point could also be used to defend the addition of iterator over grapheme clusters of string, returned by a plain iter method?

pro465:
@Lokathor 
> (other than the name of the function being too plain)

that is kinda my point, actually, but for addition instead of string iteration!

workingjubilee:
> sorry, i did not mean to come off as rude, but at least we can agree that your point could also be used to defend the addition of iterator over grapheme clusters of string, returned by a plain iter method, right?

No.

The reason grapheme clusters are not in `std` is not that it would be too convenient.
The reason is that the number of visible graphemes for a given sequence of USVs can and does change on a year to year basis,
subject to human politics and other details that are irrelevant to the machine. `std` cannot be in such a position where such a detail that is subject to such flux is in the codebase unless it absolutely must handle such.
But `std` can and does incorporate large swathes of "convenience functions",
and again, being able to write code in Rust at all is a convenience over assembly.

crates/core_simd/src/ops/assign.rs
```diff
@@ -0,0 +1,131 @@
+use super::*;
+use core::ops::{AddAssign, MulAssign}; // commutative binary op-assignment
+use core::ops::{BitAndAssign, BitOrAssign, BitXorAssign}; // commutative bit binary op-assignment
+use core::ops::{DivAssign, RemAssign, SubAssign}; // non-commutative binary op-assignment
+use core::ops::{ShlAssign, ShrAssign}; // non-commutative bit binary op-assignment
+
+// Arithmetic
+
+impl<T, U, const LANES: usize> AddAssign<U> for Simd<T, LANES>
```
calebzulawski, line 9: So this is fine, but since all of these traits are basically identical in form it might be worth using a macro?

crates/core_simd/src/ops/unop.rs
```diff
@@ -0,0 +1,103 @@
+use crate::simd::intrinsics;
```
calebzulawski, line 1: Could we call this file unary.rs instead?

workingjubilee: Sure.

crates/core_simd/src/ops/splat.rs
```diff
@@ -0,0 +1,641 @@
+use super::*;
+use crate::simd::Mask;
+
+// LHS splats
+
+macro_rules! simple_lhs_splat {
+    ( impl<const LANES: usize$(, $tvar:ident)?> $op:ident<Simd<$scalar:ty, LANES>> for $lhs:ty {
+        fn $call:ident
+    }) => {
+        impl<$($tvar,)* const LANES: usize> $op<Simd<$scalar, LANES>> for <Simd<$scalar, LANES> as Vector>::Scalar
```
calebzulawski, line 10: Looking at this again, I'm confused why it's not just implemented for the scalar directly.  Inference issues on existing code?

pro465:
> > sorry, i did not mean to come off as rude, but at least we can agree that your point could also be used to defend the addition of iterator over grapheme clusters of string, returned by a plain iter method, right?
> 
> No.
> 
> The reason grapheme clusters are not in `std` is not that it would be too convenient. The reason is that the number of visible graphemes for a given sequence of USVs can and does change on a year to year basis, subject to human politics and other details that are irrelevant to the machine. `std` cannot be in such a position where such a detail that is subject to such flux is in the codebase unless it absolutely must handle such. But `std` can and does incorporate large swathes of "convenience functions", and again, being able to write code in Rust at all is a convenience over assembly.

how about `fn iter<'a>(&'a self) -> Bytes<'a>`?

crates/core_simd/src/ops/splat.rs
workingjubilee: For most ops it's just fully parametric. I was considering whether it's worth specializing the div/rem cases for some scalars.

workingjubilee:
@pro465 So like https://doc.rust-lang.org/std/primitive.str.html#method.bytes, which is available on everything that derefs to str, including String?

pro465:
@workingjubilee yes, but with just a plain `iter` in the name instead of the clearer `bytes`

workingjubilee:
You would get that if you invoked `iter()` on `[u8]`, however.

pro465:
@workingjubilee well, that's because it isn't too convinience

workingjubilee:
Your subjective definition of "(not) too convenient" is not everyone else's, however.

pro465:
@workingjubilee ok, i guess i will leave before this gets turned into an argument...

crates/core_simd/src/ops/splat.rs
workingjubilee: Unless you mean why not
```rust
impl<T, const LANES: usize> for Op<<T as Scalar>::Vector> for T
```
and that just kinda falls apart due to failing to actually constrain all the parameters in the right places.

calebzulawski: I think I'll check out the branch and build the docs to make sure I understand what's going on.

crates/core_simd/src/ops/assign.rs
workingjubilee: Sure. I just tried to uh, aggressively minimize macro usage (or at least, use more specialized macros).

crates/core_simd/src/ops/splat.rs
workingjubilee: To elaborate on "why sometimes specializing splatting":
For most of the RHS Div/Rem impls, specializing on the "div by zero?" check could potentially bring a performance benefit.
For LHS, it only matters for signed ints (which risk overflow... I think we could make our div "wrap" correctly and so drop the `i{N}::MIN / -1` check, but I have to triple-verify we'd be emitting the right LLVMIR for that, as the alternative is UB).

workingjubilee: I think that entire question complicates this PR so I am going to simply defer it for later.

workingjubilee:
I pulled back some of the changes I had been experimenting with for later review and consideration. This version still has significant wins in terms of reducing the weight of the docs and in general making our implementation a bit cleaner.

Right now, at the tip of this branch, the LHS is not parametrically autosplatted, but the RHS is. I don't actually anticipate any functional differences as a result of this, though.

thomcc:
Hmm, I'm not in favor of the splatting this way — I think having `2.0 * vec` not work, and `vec * 2.0` work is going to be very surprising, and feels pretty janky. We do it for String, but there's a real performance cost there, and it's still a constant source of confusion for beginners.

workingjubilee:
They both work, @thomcc. I simply didn't parameterize the LHS splat impls,
so there's no case where the example you cited actually happens, even if you change up the types involved.

That is, the ops are symmetric in actuality.
I intend to revisit how all of the impls work and do additional cleanup if this is merged, but there were questions about the way the LHS was parameterized that I would rather deal with separately, later, rather than having everything bikeshedded over.

thomcc:
> They both work, @thomcc. I simply didn't parameterize the LHS splat impls,

Ah, sorry! My bad, I misread the patch then.

workingjubilee:
Mmm, that's understandable. I'll adjust the wording on that commit to reflect that.

programmerjake (approved):
lgtm

crates/core_simd/src/ops/splat.rs
```diff
@@ -0,0 +1,74 @@
...
+use super::*;
+
+macro_rules! autosplat_rhs {
+    ($(impl<T, const LANES: usize> $op:ident<$operand:ty> for Simd<$scalar:ty, LANES> {
+        fn $call:ident
+    })*) => {
+        $(impl<T, const LANES: usize> $op<$operand> for Simd<$scalar, LANES>
+        where
+            Self: $op<Self, Output=Self>,
+            $operand: SimdElement + $op<T, Output=T>,
```
calebzulawski, line 18: The op bound on the scalar is unnecessary

crates/core_simd/src/ops/deref.rs
```diff
@@ -0,0 +1,70 @@
...
+//! Ideally, Rust would take care of this itself,
+//! and method calls usually handle the LHS implicitly.
+//! So, we'll manually deref the RHS.
+use super::*;
+
+macro_rules! deref_ops {
+    ($(impl<T, const LANES: usize> $trait:ident<&Self> for Simd<T, LANES> {
+            fn $call:ident(rhs: &Self)
+        })*) => {
+        $(impl<T, const LANES: usize> $trait<&Self> for Simd<T, LANES>
```
calebzulawski, line 11: I think we should also have `impl $trait<Simd<T, LANES>> for &Simd<T, LANES>` and `impl $trait for &Simd<T, LANES>`

crates/core_simd/src/ops/unary.rs
```diff
@@ -0,0 +1,77 @@
+use crate::simd::intrinsics;
+use crate::simd::{LaneCount, Simd, SimdElement, SupportedLaneCount};
+use core::ops::{Neg, Not}; // unary ops
+
+macro_rules! neg {
+    ($(impl<const LANES: usize> Neg for Simd<$scalar:ty, LANES>)*) => {
+        $(impl<const LANES: usize> Neg for Simd<$scalar, LANES>
+        where
+            $scalar: SimdElement + Neg,
```
calebzulawski, line 9: No bounds are necessary on `$scalar` at all

crates/core_simd/src/ops/unary.rs
```diff
@@ -0,0 +1,77 @@
...
+    impl<const LANES: usize> Neg for Simd<i64, LANES>
+
+    impl<const LANES: usize> Neg for Simd<isize, LANES>
+}
+
+macro_rules! not {
+    ($(impl<const LANES: usize> Not for Simd<$scalar:ty, LANES>)*) => {
+        $(impl<const LANES: usize> Not for Simd<$scalar, LANES>
+        where
+            $scalar: SimdElement + Not,
```
calebzulawski, line 43: Ditto re bounds

crates/core_simd/src/ops/splat.rs
workingjubilee: > //! The second bound is not strictly necessary, but enforces symmetry
//! between a vector operation and repeating the scalar equivalent.

I would like to retain this one at the moment to keep us from making the kinds of mistakes that I fixed in 1ce1c645cf27c4acdefe6ec8a11d1f0491954a99, 949f71c0dc8716e285c164eae225db2d18333f0d, and c2f59483f96cf1ab1e92cf10e0f9094432a8374c... essentially, to prevent anything from sneaking past us. I don't expect it to remain in the final implementation and it's completely forward-compatible to **relax** bounds.

calebzulawski: There is no intrinsic used here, only the trait required by the bound on Self is used.

workingjubilee: The previous version did use the intrinsic directly, however, and in the future it might do so again.

workingjubilee: It's not, however, about the intrinsic. It's about whether the operator works normally on the scalar at all. For some of the scalar types you've suggested we might want to include, **it does not**, even though they are also numerical types with an in-range repr and thus **would** be valid to monomorphize. But some do implement **some** operators.

If we are including them in our discussions, then I insist on overconstraining our implementation to discourage accidentally implementing operators for them where those have not been implemented, until we have actually implemented them.

calebzulawski: Well, I still disagree that Add on the scalar means the add intrinsic works on the vector, then.  All it means is that you can call `Add::add` on it, and nothing more.

But in this case, seeing as it's the splatting case, by definition it's bound by whether you can add two vectors of the same type.

workingjubilee: Precisely, the bound is separately restated for each case.

workingjubilee: This is all especially desirable so if you want the macro implementation on the base types to not have bounds, where they would admittedly be even more superfluous. But I want to rely a **lot** less on human memory and human eyes in code review. The alternative is writing a machine to verify we don't implement an operator for a vector if its scalar doesn't, and we have a compiler right in front of us to do that.

workingjubilee: Since relaxing bounds is perfectly forward compatible, we could stabilize this version tomorrow and it would **still** be possible to remove bounds... but not add them.

crates/core_simd/src/ops/deref.rs
programmerjake: I would have that be `impl<'a, 'b, T, const LANES: usize> Add<&'a Simd<T, LANES>> for &'b Simd<T, LANES>`, cuz that's more flexible.

workingjubilee: Does that work when `'b: 'a`?

programmerjake: > 

yes, cuz it works for any `'a` and `'b` no matter their relation.

crates/core_simd/src/ops/splat.rs
workingjubilee: Updating this to fix rust-lang/rust#90904, which eliminates this discussion.

workingjubilee:
All review remarks have been addressed. Specifically, the conversation about splats has been resolved by removing splats entirely, although this is primarily due to autosplatting creating unacceptable levels of type inference related breakage at the moment. Thus, this now carries a fix for a regression currently on beta.

crates/core_simd/examples/nbody.rs
```diff
@@ -143,14 +143,14 @@ mod nbody {
...
-                bodies[k].v += f * bodies[j].mass;
+                let f = r[i] * Simd::splat(mag[i]);
+                bodies[j].v -= f * Simd::splat(bodies[k].mass);
+                bodies[k].v += f * Simd::splat(bodies[j].mass);
                 i += 1
             }
         }
         for body in bodies {
-            body.x += dt * body.v
+            body.x += Simd::splat(dt) * body.v
```
workingjubilee, line 153: Look how they massacred my boy.

Lokathor: yeah it's not great, but needs must

crates/core_simd/src/ops/assign.rs
```diff
@@ -0,0 +1,124 @@
...
+        {
+            #[inline]
+            fn $assign_call(&mut self, rhs: U) {
+                *self = self.$call(rhs);
+            }
+        })*
+    }
+}
+
+assign_ops! {
```
calebzulawski, line 33: I understand why you did it this way, but it would save something like 100 lines implementing it along the lines of:

```rust
assign_ops! {
   AddAssign::add_assign -> Add:add,
   MulAssign::mul_assign -> Mul::mul,
   ...
}
```

Using the rust-like syntax usually helps IMO but here the file would be only like 40 lines so it might actually be easier to read.

calebzulawski: (This probably applies to the other macros too, as long as they all stay in their own files to limit the scope of what you're looking at)

calebzulawski (approved):
lgtm, just the remaining stylistic comment which could go either way

calebzulawski:
also, worth noting that we probably shouldn't update nightly with these changes right away, since there are some other changes on master that are somewhat half-complete

crates/core_simd/src/ops/assign.rs
workingjubilee: I am not quiiite inclined to minimize it that far. I found the style of making the relationships in the macro more implicit slowed down finding what parts I needed to refactor out.
I _am_ inclined to minimize it in a way that still exposes all the relationships.
This needs a followup PR with more... housekeeping-y cleanup that isn't strictly related to this change anyways, so I will try to address this concern then.

crates/core_simd/examples/nbody.rs
pdx-trader: SIMD is all about performance, I know we can all get protective of our code, but you realize you're representing a larger community in both how you represent yourself and how performant your code is.  I don't think you've demonstrated either with this.  Hope Rust core fixes this.

crates/core_simd/src/ops/assign.rs
pdx-trader: Rust will never compete with Julia or C/C++ in AI/ML if we don't push and demand the fastest iterations in the world.  We need everyone giving feedback and constant pursuit of excellence here.

crates/core_simd/examples/nbody.rs
Lokathor: This is not a performance issue at all, so I'm not sure what you mean.

workingjubilee: @pdx-trader I am not sure where this statement of yours is actually coming from. There should be no actual change in Rust executable performance from this branch's changes, except for a significant positive change in compiler performance that was garnered when this branch was synced into std in https://github.com/rust-lang/rust/pull/91484

I made the original remark (a quote from The Godfather, issued by Vito Corleone played by Marlon Brando, after glancing at a badly mangled corpse) purely because with these changes, the library has become more unwieldy. There is obviously no question as to whether more syntactic noise has been introduced here. But while ease of use of the library might not be directly related to performance, it eventually is linked: if people can't use it proficiently and with relatively low friction then they won't write explicitly vectorized code as often. That is why I chose to express regret, in an... idiosyncratic manner. This syntax is quite natural for lots of programmers who write a lot of numerical code.

And it is not my nor Caleb's assessment that our library's performance is significantly less than that of other high-level SIMD APIs that run on the CPU (it may be less than custom-rolled assembly for a specific program, but it should be faster than e.g. kernels written as library routines because our library, unlike most C++ code, will be inlined and monomorphized for a specific program), but that is precisely why we have examples like this one to try to test that theory, and also give us some code samples to reflect on how we could improve.

I am happy to entertain further questions if you are confused on this subject, or anything I say. I know I am prone to bantersome commentary, and also that I view things in a way others do not, so I am used to having to explain things multiple times. However, my occasional burst of levity or off-beat-sounding remark does not mean I lack competence, and explaining things often requires participation from the inquirer as well. Namely, I have difficulty answering questions that aren't posed as questions, but rather as seemingly aimless critique.

Issue #187, opened 2021-11-13
Consider migrating vendor transmute code to stdarch


workingjubilee:
The reason why is PRs #185 and #186.

PR #197, opened 2021-11-21
Attempt to support to 64 lanes

workingjubilee:
We can try to expand our support, and test suite, to allow `Simd<T, 64>`.

### DIFF
--- a/crates/core_simd/src/lane_count.rs
+++ b/crates/core_simd/src/lane_count.rs
@@ -37,3 +37,6 @@ impl SupportedLaneCount for LaneCount<16> {
 impl SupportedLaneCount for LaneCount<32> {
     type BitMask = [u8; 4];
 }
+impl SupportedLaneCount for LaneCount<64> {
+    type BitMask = [u8; 8];
+}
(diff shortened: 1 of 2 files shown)


workingjubilee:
While highly unscientific, this branch's test suite passes compilation on @thomcc's Apple Silicon laptop also, so I think it would be fair to go for it, here.

calebzulawski:
I would add aliases for the new sizes, then lgtm.

workingjubilee:
We... never actually removed the 64-lane aliases, and we've generally been following the "what aliases would make sense if the lanes were `{B}` bits wide" where B == LEN * T::BITS, so we already have those, actually. So! If there are any new alias conventions we can bikeshed them later. 

Merging.

pthariensflame:
I think the vendor conversions may be missing.  In particular AVX512 byte vectors lack conversions.

workingjubilee:
Hm. True but also https://github.com/rust-lang/portable-simd/issues/187

jorgecarleitao:
fwiw we already accepted `usizex8` and `isizex8`, which in x86 is 512 bits wide.

PR #198, opened 2021-11-22
Generically implement `fn horizontal_{min,max}`

workingjubilee:
This collapses all instances of these two functions by merely depending on `T`  being `SimdElement + PartialOrd`.

### DIFF
--- a/crates/core_simd/src/reduction.rs
+++ b/crates/core_simd/src/reduction.rs
@@ -22,18 +22,6 @@ macro_rules! impl_integer_reductions {
             pub fn horizontal_product(self) -> $scalar {
                 unsafe { simd_reduce_mul_ordered(self, 1) }
             }
-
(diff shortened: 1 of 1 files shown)


workingjubilee:
This only saves 18kb, but it makes the docs much easier to navigate, so.

programmerjake (approved):
lgtm

calebzulawski:
I mentioned this in zulip, I don't think `PartialOrd` is the correct bound, we probably want our own trait that specifically means that these functions monomorphize.

calebzulawski:
A trivial example--pointers implement `PartialOrd` but do we want them to implement these functions?  We might, might not, and I have no idea if these functions would monomorphize.

workingjubilee:
That is not an actual concern in any of our exposed interfaces and was not a blocking concern for introducing the prior versions.

workingjubilee:
Consensus coalesced around a different direction than here.

PR #199, opened 2021-11-22
Generically implement `fn horizontal_{and,or,xor}`

workingjubilee:
20kb off, and it makes the docs significantly less crowded.
Implementation here depends on both `Self` and `T` implementing the `Bit{Op}` trait in question (so that it's coherent to reason about it as equivalent to a manual implementation where it does a series of vector `Bit{Op}`s and then a final fold of `Bit{Op}`s on essentially `Iterator<Item=T>`).


workingjubilee:
...well, that didn't happen locally.

workingjubilee:
I do not think that segfault is the fault of this PR's code, because we've been having random segfaults on rust-lang/rust also (see https://github.com/rust-lang/rust/issues/90812).

programmerjake (approved):
also lgtm

PR #201, opened 2021-11-25
Uncomment AVX512 byte vector conversions

pthariensflame:
Resolves my comment in #197, at least for now; #187 is pending but since these are already here, just commented, it seemed to make sense to me to re-enable them anyway.


workingjubilee:
I honestly hope to remove those before our next sync, but assuming I don't, this is fair yeah.

PR #203, opened 2021-11-28
Add spectral_norm example from packed_simd

deantvv:
Ported packed_simd spectral_norm [example](https://github.com/rust-lang/packed_simd/blob/master/examples/spectral_norm/src/simd.rs)
Related to #110


programmerjake (approved):
didn't look too closely, but lgtm

workingjubilee:
Ah! Sorry for not getting around to this earlier. Thank you!

PR #206, opened 2021-12-04
Use specialized traits to generalize lane comparisons

calebzulawski:
This fixes the issue where lane comparisons are determined by `PartialEq` and `PartialOrd`, which isn't correct.  It also allows `min`, `max`, `clamp`, and `horizontal_{min,max}` to be generic over element type.  This means that this PR also adds `min`, `max`, and `clamp` to integers.

The new traits `SimdEq` and `SimdOrd` probably should be part of the public interface, to allow calling the functions generically, but the functions in those traits should never be stabilized (and should probably be hidden, once that bug is worked out).  Both of these traits are sealed by `SimdElement`.

We can use the results of this PR to inform the conversion of other functions to generic over element type.

### DIFF
--- a/crates/core_simd/src/comparisons.rs
+++ b/crates/core_simd/src/comparisons.rs
@@ -1,106 +1,125 @@
 use crate::simd::intrinsics;
 use crate::simd::{LaneCount, Mask, Simd, SimdElement, SupportedLaneCount};
 
-/// SIMD vector element types that can be compared for equality.
-pub trait SimdEq: SimdElement {
-    /// Implementation detail of [`Simd::lanes_eq`].
-    fn lanes_eq<const LANES: usize>(
-        lhs: Simd<Self, LANES>,
-        rhs: Simd<Self, LANES>,
-    ) -> Mask<Self::Mask, LANES>
-    where
-        LaneCount<LANES>: SupportedLaneCount;
+mod eq {
+    use super::*;
 
-    /// Implementation detail of [`Simd::lanes_ne`].
-    fn lanes_ne<const LANES: usize>(
-        lhs: Simd<Self, LANES>,
-        rhs: Simd<Self, LANES>,
-    ) -> Mask<Self::Mask, LANES>
-    where
-        LaneCount<LANES>: SupportedLaneCount;
-}
+    pub trait Sealed: SimdElement {
+        /// Implementation detail of [`Simd::lanes_eq`].
+        fn lanes_eq<const LANES: usize>(
+            lhs: Simd<Self, LANES>,
+            rhs: Simd<Self, LANES>,
+        ) -> Mask<Self::Mask, LANES>
+        where
+            LaneCount<LANES>: SupportedLaneCount;
 
-/// SIMD vector element types that can be compared for order.
-///
-/// Note that this trait has less strict requirements than [`Ord`].  The type does not need to form a
-/// total order, but it does need to have a defined behavior for computing the minimum and maximum.
-pub trait SimdOrd: SimdElement {
-    /// Implementation detail of [`Simd::lanes_lt`].
-    fn lanes_lt<const LANES: usize>(
-        lhs: Simd<Self, LANES>,
-        rhs: Simd<Self, LANES>,
-    ) -> Mask<Self::Mask, LANES>
-    where
-        LaneCount<LANES>: SupportedLaneCount;
-
-    /// Implementation detail of [`Simd::lanes_gt`].
-    fn lanes_gt<const LANES: usize>(
-        lhs: Simd<Self, LANES>,
-        rhs: Simd<Self, LANES>,
-    ) -> Mask<Self::Mask, LANES>
-    where
-        LaneCount<LANES>: SupportedLaneCount;
-
-    /// Implementation detail of [`Simd::lanes_le`].
-    fn lanes_le<const LANES: usize>(
-        lhs: Simd<Self, LANES>,
-        rhs: Simd<Self, LANES>,
-    ) -> Mask<Self::Mask, LANES>
-    where
-        LaneCount<LANES>: SupportedLaneCount;
-
-    /// Implementation detail of [`Simd::lanes_ge`].
-    fn lanes_ge<const LANES: usize>(
-        lhs: Simd<Self, LANES>,
-        rhs: Simd<Self, LANES>,
-    ) -> Mask<Self::Mask, LANES>
-    where
-        LaneCount<LANES>: SupportedLaneCount;
-
-    /// Implementation detail of [`Simd::min`].
-    fn min<const LANES: usize>(lhs: Simd<Self, LANES>, rhs: Simd<Self, LANES>) -> Simd<Self, LANES>
-    where
-        LaneCount<LANES>: SupportedLaneCount;
-
-    /// Implementation detail of [`Simd::max`].
-    fn max<const LANES: usize>(lhs: Simd<Self, LANES>, rhs: Simd<Self, LANES>) -> Simd<Self, LANES>
-    where
-        LaneCount<LANES>: SupportedLaneCount;
-
-    /// Implementation detail of [`Simd::horizontal_min`].
-    fn horizontal_min<const LANES: usize>(x: Simd<Self, LANES>) -> Self
-    where
-        LaneCount<LANES>: SupportedLaneCount;
-
-    /// Implementation detail of [`Simd::horizontal_max`].
-    fn horizontal_max<const LANES: usize>(x: Simd<Self, LANES>) -> Self
-    where
-        LaneCount<LANES>: SupportedLaneCount;
-
-    /// Implementation detail of [`Simd::clamp`].
-    #[inline]
-    fn clamp<const LANES: usize>(
-        mut x: Simd<Self, LANES>,
-        min: Simd<Self, LANES>,
-        max: Simd<Self, LANES>,
-    ) -> Simd<Self, LANES>
-    where
-        LaneCount<LANES>: SupportedLaneCount,
-    {
-        assert!(
-            min.lanes_le(max).all(),
-            "each lane in `min` must be less than or equal to the corresponding lane in `max`",
-        );
-        x = x.lanes_lt(min).select(min, x);
-        x = x.lanes_gt(max).select(max, x);
-        x
+        /// Implementation detail of [`Simd::lanes_ne`].
+        fn lanes_ne<const LANES: usize>(
+            lhs: Simd<Self, LANES>,
+            rhs: Simd<Self, LANES>,
+        ) -> Mask<Self::Mask, LANES>
+        where
+            LaneCount<LANES>: SupportedLaneCount;
     }
 }
 
+mod ord {
+    use super::*;
+
+    pub trait Sealed: SimdElement {
+        /// Implementation detail of [`Simd::lanes_lt`].
+        fn lanes_lt<const LANES: usize>(
+            lhs: Simd<Self, LANES>,
+            rhs: Simd<Self, LANES>,
+        ) -> Mask<Self::Mask, LANES>
+        where
+            LaneCount<LANES>: SupportedLaneCount;
+
+        /// Implementation detail of [`Simd::lanes_gt`].
+        fn lanes_gt<const LANES: usize>(
+            lhs: Simd<Self, LANES>,
+            rhs: Simd<Self, LANES>,
+        ) -> Mask<Self::Mask, LANES>
+        where
+            LaneCount<LANES>: SupportedLaneCount;
+
+        /// Implementation detail of [`Simd::lanes_le`].
+        fn lanes_le<const LANES: usize>(
+            lhs: Simd<Self, LANES>,
+            rhs: Simd<Self, LANES>,
+        ) -> Mask<Self::Mask, LANES>
+        where
+            LaneCount<LANES>: SupportedLaneCount;
+
+        /// Implementation detail of [`Simd::lanes_ge`].
+        fn lanes_ge<const LANES: usize>(
+            lhs: Simd<Self, LANES>,
+            rhs: Simd<Self, LANES>,
+        ) -> Mask<Self::Mask, LANES>
+        where
+            LaneCount<LANES>: SupportedLaneCount;
+        /// Implementation detail of [`Simd::min`].
+        fn min<const LANES: usize>(
+            lhs: Simd<Self, LANES>,
+            rhs: Simd<Self, LANES>,
+        ) -> Simd<Self, LANES>
+        where
+            LaneCount<LANES>: SupportedLaneCount;
+
+        /// Implementation detail of [`Simd::max`].
+        fn max<const LANES: usize>(
+            lhs: Simd<Self, LANES>,
+            rhs: Simd<Self, LANES>,
+        ) -> Simd<Self, LANES>
+        where
+            LaneCount<LANES>: SupportedLaneCount;
+
+        /// Implementation detail of [`Simd::horizontal_min`].
+        fn horizontal_min<const LANES: usize>(x: Simd<Self, LANES>) -> Self
+        where
+            LaneCount<LANES>: SupportedLaneCount;
+
+        /// Implementation detail of [`Simd::horizontal_max`].
+        fn horizontal_max<const LANES: usize>(x: Simd<Self, LANES>) -> Self
+        where
+            LaneCount<LANES>: SupportedLaneCount;
+
+        /// Implementation detail of [`Simd::clamp`].
+        #[inline]
+        fn clamp<const LANES: usize>(
+            mut x: Simd<Self, LANES>,
+            min: Simd<Self, LANES>,
+            max: Simd<Self, LANES>,
+        ) -> Simd<Self, LANES>
+        where
+            LaneCount<LANES>: SupportedLaneCount,
+        {
+            assert!(
+                Self::lanes_le(min, max).all(),
+                "each lane in `min` must be less than or equal to the corresponding lane in `max`",
+            );
+            x = Self::lanes_lt(x, min).select(min, x);
+            x = Self::lanes_gt(x, max).select(max, x);
+            x
+        }
+    }
+}
+
+/// SIMD vector element types that implement [`PartialEq`].
+pub trait SimdPartialEq: SimdElement + PartialEq + eq::Sealed {}
+
+/// SIMD vector element types that implement [`PartialOrd`] and can always be compared.
+///
+/// Note that this trait is has one additional requirement beyond [`PartialOrd`]: all values can be
+/// compared with all other values.
+/// This is similar to [`Ord`], but without the requirement that comparisons are symmetric
+/// (e.g. `a < b` and `a > b` can both be true for some values).
+pub trait SimdPartialOrd: SimdElement + PartialOrd + ord::Sealed {}
+
 macro_rules! impl_integer {
-    { $($type:ty),* } => {
+    { unsafe { $($type:ty),* } } => {
         $(
-        impl SimdEq for $type {
+        impl eq::Sealed for $type {
             #[inline]
             fn lanes_eq<const LANES: usize>(
                 lhs: Simd<Self, LANES>,
@@ -124,7 +143,9 @@ macro_rules! impl_integer {
             }
         }
 
-        impl SimdOrd for $type {
+        impl SimdPartialEq for $type {}
+
+        impl ord::Sealed for $type {
             #[inline]
             fn lanes_lt<const LANES: usize>(
                 lhs: Simd<Self, LANES>,
@@ -208,15 +229,18 @@ macro_rules! impl_integer {
             {
                 unsafe { intrinsics::simd_reduce_max(x) }
             }
+
         }
+
+        impl SimdPartialOrd for $type {}
         )*
     }
 }
 
 macro_rules! impl_float {
-    { $($type:ty),* } => {
+    { unsafe { $($type:ty),* } } => {
         $(
-        impl SimdEq for $type {
+        impl eq::Sealed for $type {
             #[inline]
             fn lanes_eq<const LANES: usize>(
                 lhs: Simd<Self, LANES>,
@@ -240,7 +264,9 @@ macro_rules! impl_float {
             }
         }
 
-        impl SimdOrd for $type {
+        impl SimdPartialEq for $type {}
+
+        impl ord::Sealed for $type {
             #[inline]
             fn lanes_lt<const LANES: usize>(
                 lhs: Simd<Self, LANES>,
@@ -327,16 +353,18 @@ macro_rules! impl_float {
                 unsafe { intrinsics::simd_reduce_max(x) }
             }
         }
+
+        impl SimdPartialOrd for $type {}
         )*
     }
 }
 
-impl_integer! { u8, u16, u32, u64, usize, i8, i16, i32, i64, isize }
-impl_float! { f32, f64 }
+impl_integer! { unsafe { u8, u16, u32, u64, usize, i8, i16, i32, i64, isize } }
+impl_float! { unsafe { f32, f64 } }
 
 impl<T, const LANES: usize> Simd<T, LANES>
 where
-    T: SimdEq,
+    T: SimdPartialEq,
     LaneCount<LANES>: SupportedLaneCount,
 {
     /// Test if each lane is equal to the corresponding lane in `other`.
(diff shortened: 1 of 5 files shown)


Lokathor:
If they should be public so people can use them generically then why should the methods never be stabilized?

calebzulawski:
The traits are used in the bounds of the member fns of `Simd`, so you may want to check `T: SimdOrd` in order to be able to call `Simd::<T, _>::min`, but you shouldn't call `<T as SimdOrd>::min` directly.

Lokathor:
what is the harm if someone did?

calebzulawski:
There's no harm, it just complicates the API.  There's no good reason to expose it because then we have the liberty to change it in the future.  I suggested hiding them to simplify the docs as well.

The trait is effectively a marker to map type to intrinsic.  A similar trait is `std::convert::FloatToInt`, which has a hidden intrinsic method as well.

calebzulawski:
It's worth noting that `Simd::min` and `SimdOrd::min` are completely identical, so I don't see any good reason to provide the latter, in any situation where you have `SimdOrd` you can just call `Simd::min`.

Lokathor:
I just (1) am fundamentally against perma-unstable anything (2) the kind of person that frequently implements traits for newtypes via macro, and i would usually like the possibility to implement all of a base type's traits on the newtype as well, should it be appropriate.

calebzulawski:
In this case, these traits are literally just mapping of intrinsics to types.  They're sealed so users can't implement them on newtypes.

programmerjake (approved):
lgtm, though the `Simd` methods could just call the intrinsics directly if we wanted, rather than calling the methods on `SimdOrd` which then calls the intrinsics.

programmerjake:
so, kinda like:
```rust
pub trait SimdOrd: ... {}

impl SimdOrd for f32 {}

impl<T: SimdOrd, const LANES: usize> Simd<T, LANES> {
    pub fn max(self, rhs: Self) -> Self {
        unsafe { max_intrinsic(self, rhs) }
    }
}
```

calebzulawski:
That isn't possible in some cases, since the implementations for different types aren't the same (for example, `min` and `max` are different for integers and floats).  I expect some other functions will work this way, too.  This also makes it much easier in the future to add new element types that may have very different implementations (like the `NonZero` integers, which would probably convert to the underlying type and defer to their implementation).

programmerjake:
I expect we will definitely want fp min/max intrinsics (llvm may have issues recognizing the patterns)...so, while we're at it, just have int min/max too.

for this pr, yeah, we can just leave it with the calling `SimdOrd` methods, for future extension and for `min`/`max` right now.

workingjubilee:
NonZero types that interact with their "base" type using ops that return a type other than bool have ops that return the NonZero type, not the underlying type.

calebzulawski:
All I'm saying is that `Simd::<NonZeroU8, _>::horizontal_max` can't call the intrinsic directly, it'll call something like `NonZeroU8::new_unchecked(self.get().horizontal_max())`, and therefore the trait fn is necessary.

workingjubilee:
> There's no harm, it just complicates the API. There's no good reason to expose it because then we have the liberty to change it in the future. I suggested hiding them to simplify the docs as well.

If there's no harm in having them exposed, then we should simply have them exposed, unless we actually intend to remove them later. And if this proposal is serious, then we do not. I disagree that it simplifies the docs: `#[doc(hidden)]` is not actually a promise that no one can see or use the function, and hiding the methods needlessly obfuscates the code to Rust programmers. I am aware that people on crates.io use that as a convention, but it isn't how Rust's stability promises work regarding the traits `std` implements.

This is true even if the trait is also sealed, and there is no reason to leave these traits as permanently unstable. People should be able to depend on the trait after stabilization for the types for which it is implemented for.

crates/core_simd/src/comparisons.rs
```diff
@@ -1,56 +1,447 @@
 use crate::simd::intrinsics;
 use crate::simd::{LaneCount, Mask, Simd, SimdElement, SupportedLaneCount};
 
+/// SIMD vector element types that can be compared for equality.
+pub trait SimdEq: SimdElement {
```
workingjubilee, line 5: This **must** have a PartialEq bound, otherwise it is incorrect.

workingjubilee (changes requested):

crates/core_simd/src/comparisons.rs
calebzulawski: I disagree.  `PartialEq` only means that you can call `PartialEq::eq` and friends.  While it is logically true, it has no bearing on whether `SimdEq` can be implemented, and `SimdEq` never uses `PartialEq`.

calebzulawski:
I understand your sentiment but I disagree with what these traits mean.  `SimdEq` exists purely for the purpose of gating `Simd::lanes_eq` to types that actually implement equality, and for allowing us to define that implementation per type.  The interface that users call should always be `Simd::lanes_eq`.  `SimdEq::lanes_eq` exists as an implementation detail.  We should be free to remove it, change it, etc, without worrying if any users are using it.  Identically, something that's already in `std`: when you convert an `f32` to an int you use `f32::to_int_unchecked`, not `<f32 as FloatToInt<{integer}>>::to_int_unchecked`, because the latter is an implementation detail, a trait wrapper around an intrinsic.  Making that part of the public rust API by stabilizing it (and removing the `#[doc(hidden)]`) is unnecessary and bloats the API with no benefit to the user (and adds some confusion because now there are two duplicate functions!).

crates/core_simd/src/comparisons.rs
calebzulawski: I should clarify, I don't disagree that it would be bizarre for us to implement `SimdEq` on something that doesn't implement `PartialEq`, but it's not unsound, and I'd like to be as conservative as possible with the implications of these traits.

workingjubilee: For all vector intrinsics, they should be lowerable to a simple scalar loop, there is no reason Rust cannot be the target language for that lowering, and we have extensive tests precisely to guarantee that equivalency.

So no, I fully believe it is unsound.

workingjubilee:
There is a tracking issue for that trait's stabilization.
https://github.com/rust-lang/rust/issues/67057

So no, I do not believe that reasoning applies.

crates/core_simd/src/comparisons.rs
Lokathor: Perhaps unsound isn't quite the correct term, because the lowering could just fail and crash the build if it's actually attempted, but either way it's going to allow footgun situations we shouldn't allow. PartialEq should be part of the bound.

calebzulawski: In that case, then I think you're arguing that the intrinsic should be `simd_eq<T: PartialEq, U>`, which sounds good to me.  `SimdEq` can be implemented perfectly safely without requiring `trait SimdEq: PartialEq`.

Lokathor:
I would argue the other way: `Simd::lanes_eq` exists because we don't want to fill the prelude with more stuff and we don't yet have intrinsic trait impls.

crates/core_simd/src/comparisons.rs
Lokathor: If the intrinsic has that bound, and the trait is for calling that intrinsic, i don't understand how the trait doesn't also carry the bound.

calebzulawski:
I guess whether or not it's part of the API is somewhat immaterial to this PR--I wish it weren't but it doesn't make a huge difference either way.  Like `FloatToInt`, I'd prefer if we make the functions hidden purely from the documentation perspective (isn't documentation the whole reason we went down this road of traits in the first place?) but if we must stabilize them I won't argue against it and I still think this PR is the right way to go.

crates/core_simd/src/comparisons.rs
calebzulawski: Maybe it's not obvious, but there are no default implementations of `lanes_eq` etc in the traits.  The intrinsics are only used in the individual implementations over the types, and the safety must be confirmed in those individual `unsafe`s around the intrinsics.

workingjubilee: While I mentioned the intrinsic, we already decided, a long time ago, that we would not pass the buck to the intrinsics, but would instead assure type-safety at the level of the functions that call them, and prevent monomorphization errors. That means that if we know there is a relevant bound then it must inhere at a higher level than the intrinsic function call per se. This is doubly true since we can adjust the intrinsics any-which-way if we want.

The internal safety constraints of `simd_eq`, `simd_ne`, and such, should in fact include "there is a valid scalar implementation". This was not worth noting before, but now that we are discussing adding arbitrary extensions to our implementation surface, it is.

calebzulawski: Please look back at the trait and its implementations.  The trait has no default implementations, so the uses of the intrinsics are only in the `impl` blocks for floats and integers, for which of course the intrinsics monomorphize.  There is no way to produce either unsoundness or monomorphization errors.

workingjubilee: I am aware that it is currently fine, in the currently implemented cases. But so is the existing implementation.

My understanding is that this PR is intended to allow us to open up the design of this library to all types that could _possibly_ be given as a type argument of `Simd<T, _>`, present or future. This includes, potentially, "long after both Caleb Zulawski and Jubilee Young are gone". If so, then we should brace for explaining ourselves to Posterity, and we must take seriously the idea that someone may PR an implementation to rust-lang/rust that we would see as incorrect, and it will be reviewed by someone else while we cannot be reached for comment (on account of it being the year 2230, perhaps?). Merely relying on a bound of SimdElement will not be enough.

I would have the PartialEq and PartialOrd bounds so that we are able to offer the most succinct explanation of the intended behavior of SimdEq and SimdOrd's functions, by explaining the lemma that Simd{Eq,Ord} must uphold in terms of Partial{Eq,Ord}. That is, that
```rust
forall<w0,w1> {
  SimdOrd::lane_lt(w0, w1) == w0.to_array()
                                .into_iter()
                                .zip(w1.to_array())
                                .map(|(u0, u1)| PartialOrd::lt(u0, &u1))
                                .enumerate()
                                .fold(Mask::<_, _>::splat(false), |acc, (i, bit)| {
                                    let mut m = acc; // ...This is extremely grody?!
                                    m.set(i, bit);
                                    m
                                }))
}
```

Mind, this is about as much logical and reasoning-based bounds as actual programmatic in-language bounds. I could possibly believe that merely tediously documenting everything we expect to be upheld is fine.

I could also believe that the intention is to explicitly **not** uphold the lemma posed, but I don't see why, in that case. The more conservative we are in our implications, the less the programmer using this library can rely on our implementation being sound and simple to reason about.

programmerjake: I'd be fine having the bound `SimdEq: PartialEq` or not .. either way's fine with me. I don't think any of std should usually depend on that bound though (e.g. code shouldn't have `T: SimdEq` => call `T::eq` everywhere)

workingjubilee: Speaking of `std`: assuming the lemma holds, the trait should also be `unsafe`, to declare that it is a trusted implementation and allow impls in `std` to specialize based on that knowledge. Otherwise any such specialization would _definitely_ be unsound.

calebzulawski: Okay, we're definitely not remotely on the same page here.

The trait is by no means `unsafe`.  It has no `unsafe` implications.  `PartialEq` isn't unsafe, and this trait is no different in that regard.  An incorrect implementation is just incorrect--it doesn't cause any UB.

Regarding the trait, `PartialEq` and `SimdEq` might mean the same thing, but `PartialOrd` and `SimdOrd` definitely do not, and I documented that.  `SimdOrd` is stricter than `PartialOrd`, but less strict than `Ord`.  While it wouldn't be invalid to have `SimdOrd: PartialOrd`, the implication that they're the same thing is wrong.

This gets more inconsistent when we start implementing more traits in the future for things that don't have equivalent traits in `std`.  There's no trait for `abs`, so why is it unsound for `SimdOrd` to not require `PartialOrd`, but it's fine to have `SimdAbs` with no bounds?

workingjubilee: A trait can have no unsafe code in it whatsoever yet be marked `unsafe`. 
Doing so allows **other** unsafe code to rely on it as a "trusted implementation".
`std` has implemented many `unsafe` traits specifically for the purpose of extracting optimizations in specialized implementations based on this premise. In general, `std` cannot rely on e.g. the correctness of an implementation of Ord or Eq because those implementations are safe, thus it is acceptable for the implementations to simply be incorrect. It will sometimes overlook this requirement, but only for cases where it is absolutely dead certain this is acceptable (i.e. something has gone seriously awry if a `u8` is not a `u8`).

I only mentioned it because there is a fair amount of interest in using `Simd` types to attempt optimizations in `std` or the compiler. It's not required. I just believe we should reject any such optimizations that don't hinge on well-defined properties, and making it an `unsafe trait` would lower the bar somewhat, by demanding that the safety bounds were upheld by later impls (and offering a valid reason to roll back an implementation that broke them).

I don't know what "more strict" or "less strict" mean. They don't have any natural meaning to me. It says "a defined behavior for computing the minimum and maximum", but that means `SimdOrd::lane_lt` could return `Mask::splat(false)`, as that is a "defined behavior".

calebzulawski: Well the documentation makes it clear:
```
/// SIMD vector element types that can be compared for order.
///
/// Note that this trait has less strict requirements than [`Ord`].  The type does not need to form a
/// total order, but it does need to have a defined behavior for computing the minimum and maximum.
```

workingjubilee:
Regarding SimdAbs:

I suggested traits because I felt we should have {Simd?}Float and {Simd?}{S,U,}Int, not SimdAbs, and that the traits should define that the vectors they are implemented for implement a natural subset of the behaviors of floats or integers, with very well-defined variations. That would keep the number of traits down to a sane and simple scope. I did offer a SimdAbs impl variation, briefly, but I no longer am inclined to such. Even so, I am not perturbed by the possible asymmetry.

crates/core_simd/src/comparisons.rs
calebzulawski: Look, I think you missed the point.  We can improve the documentation, my point was that `SimdOrd` is NOT `PartialOrd`, nor is it `Ord`.

calebzulawski: I'm thinking this can go on and on, it seems like you think the trait should be `unsafe` and imply `PartialOrd` because you want to say "no really, this is really important to implement correctly", but that's not what either of those mean, that's really a job for documentation.  I guess I'm going to leave it with an objective measure: can you provide a code example that doesn't use `unsafe` that results in UB because the trait isn't unsafe, or because it isn't bound by `PartialEq` etc?

workingjubilee: > Well the documentation makes it clear:

That does not make it clear to me, and not in a way that suggests implementing `SimdOrd::lane_lt` with an unconditional `Mask::splat(false)` is incorrect.

> I'm thinking this can go on and on, it seems like you think the trait should be unsafe and imply PartialOrd because you want to say "no really, this is really important to implement correctly", but that's not what either of those mean, that's really a job for documentation.

I insist on adding well-specified relationships so as to encapsulate reasoning or bound correctness. If there is no encapsulation of reasoning going on then there is no reason to merge an additional 400 lines that complicate our interface considerably when there are also no examples of incorrectness in the current interface. Your hypothesis that the current bounds are incorrect also lacks practical examples.

calebzulawski: I have many, many times indicated that NonZero integers will not work with the current bounds.

workingjubilee: The way in which they "will not work" is that it "doesn't work" if we implement `T: SimdElement` for `NonZero*` unless we add rules that allow that to monomorphize to the intrinsic in question. But that suggests the current interface, which was designed around the presumption of using entirely raw primitive types, is non-extensible, not incorrect. It means our work tightly scoped for types which reflect actual values in `Simd` lanes and that also incidentally do not violate "safe transmute" rules, as `NonZero*` does.

Incorrect would be to allow `impl Add for Simd<NonZero*, _>`, when `NonZero*` does not implement `core::ops::Add` precisely because the obvious way to do so could break its invariants. Even if our implementation was sound, to preempt the scalar type's implementation would allow unacceptable degrees of meaning drift. And I see no reason to assume all the code we will write will be sound, nor that we will be the only contributors and reviewers.

For that reason, none of these hypothetical extension traits should preempt a scalar implementation, and they should all be bounded on the presence of a scalar trait where applicable and have a well-defined relationship between the scalar and vector implementation, ideally something we can test, where it applies across all types.

This is not the first time I have attempted to explain this, either.

thomcc:
This is not a good direction to go in, IMO.

As far as I can tell, the only reason to have `SimdEq` and `SimdOrd` is to support user types in Simd, which we should just not support. If you recall, we were careful to require that vectorized comparison (and essentially all other operations) was semantically the same as the scalar comparison in a loop — which isn't something we get for free (portable NaN handling has a cost on e.g. x86).

Keep in mind: even having `SimdElement` *at all* is already a step beyond most similar features in the stdlib (compare with `core::sync::atomic`, scalar integers/floats, ...), so we're *already* much more flexible and generic than we need to be.

People who want `Simd<UserType, N>` can write a transparent wrapper around `Simd<Scalar, N>` (where `UserType` is a transparent wrapper around `Scalar`), and convert/transmute on use. Again, precedent here is `core::sync::atomic` which we're *already way more flexible and generic than*.

Similarly, `Simd<NonZeroInt, _>` sounds like a headache for everybody. People who want to have niches in their vectors are more than capable of using `[NonZeroInt; N]` and convert to a vector on use. People who want to semantically forbid zeros can just wrap `Simd<Int, _>`.

TBH, given the bikeshedding and discussion about hypothetical future use cases we won't be able to add compatibly... I can see why the stdlib usually opts not to make traits for these things (or opts not to make them public). Designing a trait like this without excluding some hypothetical future use cases is both impossible, and seen as the Worst Possible Thing.

crates/core_simd/src/comparisons.rs
calebzulawski: Okay, I didn't understand that you were suggesting that we should expand the intrinsics to cover non-primitive types.  I agree that if were were to do that, it's not necessarily incorrect (though I'm not convinced that it is absolutely correct, either).  I'm writing a more comprehensive comment below and I'll include more detail there.

workingjubilee: I am not.
I am stating that it is an alternative path.

workingjubilee: > I'm thinking this can go on and on, it seems like you think the trait should be unsafe and imply PartialOrd because you want to say "no really, this is really important to implement correctly", 

That is not the reason I gave.

calebzulawski:
Okay, I'm going to roll the discussions from the comment thread and the main thread into this comment, to go through all the critiques in one place.

**Are these traits required at all?**
`SimdEq` is roughly equivalent to `PartialEq`, but there is no way to provide `min`, `max`, and `clamp` for floats, because `PartialOrd` does not provide those functions!  They are part of `Ord`, but we can't use that either, since it is not implemented for `f32` or `f64`.  Even if we _did_ use `PartialOrd` for the comparisons, we would _still_ need a separate trait for these functions.  Not to mention all of the other functions (saturating math, abs, sqrt, etc) that don't correspond to an existing trait in `std` will need their own `std::simd` trait, anyway.  In the future I imagine having something like `SimdSaturating` (saturating arithmetic), `SimdSigned` (for `abs`, checking signs, etc), and then perhaps `SimdInt`, `SimdFloat`, for more specific operations (conversions, `sqrt`, trig, etc).

**User types**
This PR has nothing to do with user types.  These traits are sealed by `SimdElement` and I (personally) have no desire to ever support user types.

**Library newtypes types**
`NonZero` integers is just one good example, I think another good example is `std::num::Saturating`.  Considering how common saturating arithmetic is in SIMD, I could see that being very commonly used.  `NonNull` is also a good candidate, considering it has some useful semantic meaning beyond providing niches.  It's not a type yet, but I've seen some people advocate for `NonNan` float types, which would also help `std::simd` significantly.  I'm not suggesting we rush to support every single primitive newtype that exists, but I don't think a design that expressly prohibits them is a good idea, either.

**Implementing newtypes by extending intrinsics**
In the thread above, @workingjubilee suggested that instead of individually implementing each function as done in this PR, you could instead extend the intrinsics to support newtypes.  While this is theoretically possible, I don't think this is the way to go.  Consider the following example:

Imagine you want to add support for an arbitrary newtype (let's use all of the `NonZero` integers, and assume `#[repr(simd)]` has been appropriately updated to support newtypes as well).  Now you must modify all of the following:
* the LLVM backend intrinsics
* the cranelift backend intrinsics
* the hypothetical scalar polyfill library

My concerns with this approach:
* newtypes are promoted to something "special" in the compiler, somewhere near the level of primitives
* newtype logic that can be easily represented in plain rust in `std::simd` (like `simd_abs(x.get())`) must now instead be duplicated in each backend
* implementing a new backend would require an inordinate amount of knowledge about each individual newtype

While this approach could theoretically work, I think it makes more sense for the intrinsics to only use primitive types, and handle logic around newtypes within `std::simd`.  Using traits like in this PR allows you to customize how the various functions work for each newtype and maintain a clear distinction between newtypes at the library level and primitives at the compiler level.

crates/core_simd/src/comparisons.rs
calebzulawski: > I am not.
I am stating that it is an alternative path.

Okay, well I wrote a bit below about why I believe this is the better path at least, so we can at least try to rule that particular path out.  I'm definitely not a fan of it.

> That is not the reason I gave.

In that case I think I don't quite understand your reason.  I still definitely don't think this trait is `unsafe`, I could be willing to compromise on requiring `PartialEq`/`PartialOrd` but I want to at least come to an agreement on whether or not that's required for soundness, or whether we are just using it as a semantic hint.

workingjubilee:
> In the thread above, @workingjubilee suggested that instead of individually implementing each function as done in this PR, you could instead extend the intrinsics to support newtypes. 

I do not desire this, I stated that it is what limits the current interface, and that we originally designed SimdElement to mean "A machine scalar type", exactly matching the bounds in typeck as we found them. I do not desire this proposed extension, and I do not desire this infinite increase to the scope of our work. I do not want you to say I "suggested" something when I enumerate alternate paths, as it would have to mean I desire it, or even think it is a good idea, for it to actually be a suggestion. I have stated this boundary multiple times. Please do not transgress it again.

I agree with Thom. We should not go down this path.
NonZeroInt is just another user type: it merely happens that the user is `std`.
And there is no clear reason to support `Simd<Saturating<i32>, N>` instead of `Saturating<Simd<i32, N>>`.

crates/core_simd/src/comparisons.rs
workingjubilee: Not understanding is fine. I will try again a bit later to explain.

calebzulawski:
Sorry, when I said suggest I didn't mean "advocate", I meant "propose to be considered".

You're right that newtypes could potentially be handled by wrapping `Simd`, but that complicates passing through the more complicated APIs like scatter/gather.  The nice thing about `Simd` is that it's a container.  Again, what I am getting at is that I'm not ready to make the decision today to prohibit using newtypes within a vector.

Maybe it was my mistake picking the comparison functions first, but I wanted to target the most complicated ones first.  If we don't go down this path, can I ask how we should implement `min`, `max`, and `clamp`?

workingjubilee:
> Sorry, when I said suggest I didn't mean "advocate", I meant "propose to be considered".

Thank you.

I prefer "enumerate", as it is appropriately tediously dry and void of implication beyond the monotonic iteration of a thinking machine over multiple variations.

calebzulawski:
After a discussion with @workingjubilee, I've made some changes that I believe clarify the purpose and reasoning behind these changes.

There are two traits:
* `SimdPartialEq`, which indicates that a type can be used to compute vector `PartialEq` operations
* `SimdPartialOrd`, which indicates that a type can be used to compute vector `PartialOrd` operations

These traits indicate vector semantics, marking which vector operations can be done.  The implementations of these functions are handled by sealed traits.

The reasoning behind these changes is simply that different types need different implementations, and relying on `PartialEq` and `PartialOrd` does not convey enough information to implement everything.  Type wrappers such as the `NonZero` integers are one example, but even primitive integers and floats are significantly different and require different implementations.  `clamp`, for example, stands out as it's not possible to write a generic implementation for both integers and floats.

programmerjake (approved):
generally lgtm, though we should consider renaming `lanes_eq` to `eq` (and other similar functions) or `max` to `lanes_max` (and other similar functions) so the lane-wise operations are named consistently.

PR #207, opened 2021-12-04
Remove Select trait

calebzulawski:
I realized that our `select` implementation predated `Simd` being generic over element type, and we don't really need the `Select` trait at all.  The function signature is much simpler now (generic over element type, rather than over the entire vector).  This did require changing mask select to be a different function, but I think that's fine considering they're not necessarily vectors.


workingjubilee:
Yeah, Mask is something we should always be considering as its own type. This looks good!

Issue #208, opened 2021-12-05
[QUESTION] what do the simd cmp intrinsics lower to?


pro465:
https://github.com/rust-lang/portable-simd/blob/master/crates/core_simd/src/intrinsics.rs#L49-L54

they didn't have any comments explaining what specific LLVM instruction they get lowered to, and i also couldn't find any instruction on the LLVM reference that closely resembled them

bjorn3:
They get lowered to the `icmp` or `fcmp` llvm instruction depending on the lane type: https://github.com/rust-lang/rust/blob/1c0ed0d4c4bd7b39299966e57806d29ce38c95af/compiler/rustc_codegen_ssa/src/base.rs#L98-L124

calebzulawski:
It's worth noting that there are no guarantees as to what anything lowers to--the only guarantee is that a function will perform the operation defined in its documentation, and do it relatively efficiently.

Lokathor:
However please note that the entire portable-simd project is "best effort" in terms of giving a specific instruction.

We guarantee that the comparison will happen (and produce correct outputs) at a logical level, but the *exact* hardware path that happens is not intended to be guaranteed.

pro465:
@Lokathor @bjorn3 @calebzulawski 

thank you all very much for your helpful analswers. i however, didnt understand anything about the code that @bjorn3 pointed to. does it just sign-extend `<N x i1>` to whatever the return type is? if so, calling `intinsics::simd_eq::<Simd<_, _>, bool` should not work, right?

bjorn3:
In LLVM `i1` is the type for booleans.

pro465:
@bjorn3 oh so, is calling `instrinsics::<Simd<N, _>, [bool; N]>` fine?

pro465:
ok lemme try

pro465:
@bjorn3 didn't work :(

https://github.com/pro465/portable-simd/runs/4429119312?check_suite_focus=true

bjorn3:
[...]

PR #209, opened 2021-12-06
use instrinsics for simd equality

pro465:
this does exactly what the `//TODO` comments planned to do.

special thanks to @bjorn3 for helping me understand the simd instrinsics and the compiler errors, and proposing solutions (#208).


calebzulawski:
I think it's worth waiting for #206, since that will definitely affect this implementation.

workingjubilee:
Closed in #274.

PR #210, opened 2021-12-09
Refactor ops.rs with wrapping shifts 

workingjubilee:
This approaches reducing macro nesting in a slightly different way. Instead of just flattening details, make one macro apply another. This allows specifying all details up-front in the first macro invocation, making it easier to audit and refactor in the future.

This refactor also has some functional changes. Only one is a true behavior change, however:
- The visible one is that SIMD shifts are now wrapping, not panicking on overflow
- `core::simd` now has a lot more instances of `#[must_use]`, which merely lints
- div/rem now perform a SIMD check but remain as before, which should improve performance but be invisible

### DIFF
--- a/crates/core_simd/src/ops.rs
+++ b/crates/core_simd/src/ops.rs
@@ -220,6 +220,70 @@ bitops! {
     }
 }
 
+macro_rules! float_arith {
+    ($(impl<const LANES: usize> FloatArith for Simd<$float:ty, LANES> {
+        fn add(self, rhs: Self) -> Self::Output;
+        fn mul(self, rhs: Self) -> Self::Output;
+        fn sub(self, rhs: Self) -> Self::Output;
+        fn div(self, rhs: Self) -> Self::Output;
+        fn rem(self, rhs: Self) -> Self::Output;
+     })*) => {
+        $(
+            unsafe_base_op!{
+                impl<const LANES: usize> Add for Simd<$float, LANES> {
+                    fn add(self, rhs: Self) -> Self::Output {
+                        unsafe { simd_add }
+                    }
+                }
+
+                impl<const LANES: usize> Mul for Simd<$float, LANES> {
+                    fn mul(self, rhs: Self) -> Self::Output {
+                        unsafe { simd_mul }
+                    }
+                }
+
+                impl<const LANES: usize> Sub for Simd<$float, LANES> {
+                    fn sub(self, rhs: Self) -> Self::Output {
+                        unsafe { simd_sub }
+                    }
+                }
+
+                impl<const LANES: usize> Div for Simd<$float, LANES> {
+                    fn div(self, rhs: Self) -> Self::Output {
+                        unsafe { simd_div }
+                    }
+                }
+
+                impl<const LANES: usize> Rem for Simd<$float, LANES> {
+                    fn rem(self, rhs: Self) -> Self::Output {
+                        unsafe { simd_rem }
+                    }
+                }
+            }
+        )*
+    };
+}
+
+// We don't need any special precautions here:
+// Floats always accept arithmetic ops, but may become NaN.
+float_arith! {
+    impl<const LANES: usize> FloatArith for Simd<f32, LANES> {
+        fn add(self, rhs: Self) -> Self::Output;
+        fn mul(self, rhs: Self) -> Self::Output;
+        fn sub(self, rhs: Self) -> Self::Output;
+        fn div(self, rhs: Self) -> Self::Output;
+        fn rem(self, rhs: Self) -> Self::Output;
+    }
+
+    impl<const LANES: usize> FloatArith for Simd<f64, LANES> {
+        fn add(self, rhs: Self) -> Self::Output;
+        fn mul(self, rhs: Self) -> Self::Output;
+        fn sub(self, rhs: Self) -> Self::Output;
+        fn div(self, rhs: Self) -> Self::Output;
+        fn rem(self, rhs: Self) -> Self::Output;
+    }
+}
+
 /// Automatically implements operators over references in addition to the provided operator.
 macro_rules! impl_ref_ops {
     // binary op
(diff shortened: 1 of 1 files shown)


crates/core_simd/src/ops.rs
```diff
@@ -32,232 +31,497 @@ where
...
-            fn $fn:ident($self_tok:ident, $rhs_arg:ident: $rhs_arg_ty:ty) -> Self::Output $body:tt
+/// SAFETY: This macro should not be used for anything except Shl or Shr, and passed the appropriate shift intrinsic.
+/// It handles performing a bitand in addition to calling the shift operator, so that the result
+/// is well-defined: LLVM can return a poison value if you shl, lshr, or ashr if rhs >= <Int>::BITS
+/// At worst, this will maybe add another instruction and cycle,
+/// at best, it may open up more optimization opportunities,
+/// or simply be elided entirely, especially for SIMD ISAs which default to this.
+///
+// FIXME: Consider implementing this in cg_llvm instead?
+// cg_clif defaults to this, and scalar MIR shifts also default to wrapping
```
calebzulawski, line 65: I don't know if I would go into this much detail about backends... realistically it's pretty transparent to the optimizer and it's reasonable to decide that the intrinsic does *not* allow overflow.  It makes more sense to put the logic here so the logic isn't duplicated in backends, and so it's still possible to make unchecked variants.

crates/core_simd/src/ops.rs
```diff
@@ -32,232 +31,497 @@ where
...
-    { impl BitXor for $scalar:ty } => {
-        impl_op! { @binary $scalar, BitXor::bitxor, simd_xor }
-    };
+macro_rules! bitops {
+    ($(impl<const LANES: usize> BitOps for Simd<$int:ty, LANES> {
+        fn bitand(self, rhs: Self) -> Self::Output;
+        fn bitor(self, rhs: Self) -> Self::Output;
+        fn bitxor(self, rhs: Self) -> Self::Output;
+        fn shl(self, rhs: Self) -> Self::Output;
+        fn shr(self, rhs: Self) -> Self::Output;
```
calebzulawski, line 96: Personally I'm not a huge fan of so much of the pseudo-rust arguments, I actually find it harder to parse because it's not clear what's rust and what's not.  IMO now you have a whole line that's just a verbose way of saying "shr" or whatever.

calebzulawski: I should say that I would be fine with it if it weren't dramatically more verbose, but this doubles the length of the file with no significant functionality changes.

crates/core_simd/src/ops.rs
workingjubilee: The reason to move the logic to cg_llvm is that cg_clif already emits this in terms of instructions, if necessary, so the extra AND may simply be more noise for less burly optimizers like cg_clif. It's not necessarily clear that this is the answer, however, which is why I left a note.

calebzulawski: Sure. I'm actually thinking it would make sense to go the other way if possible, use an unchecked variant in cg_clif.

crates/core_simd/src/ops.rs
workingjubilee: A **few** extra lines actually helps me scan and do various text manipulation tricks, but I agree that I was aiming for... less than literally twice as many lines, yes. I will take another pass through it for savings.

calebzulawski: One more thing I found slightly misleading is the `impl BitOps` etc, because it makes me think there's an actual `BitOps` trait and not that it's shorthand for many

calebzulawski: Looking over this again, I think it's unnecessary to specify `BitOps` anyway, since it's also the name of the macro.  What about a single line like `impl BitAnd, BitOr, BitXor, Shl, Shr for i32` or whatever?

workingjubilee: Hm, the other way around, it should be like

`impl BitAnd for {i8, i16, i32, i64, isize}`
because the target types will remain consistent and the implementation for a _set_ of types (ints, uints, floats) will be consistent across that set, but a given implementation may vary subtly, and what we want to make easiest to edit is the implementation.

workingjubilee:
Well now it's a net reduction in lines but I can't tell whether this is clearer or not to people who aren't me.

programmerjake:
seems simpler, idk about clearer

crates/core_simd/src/ops.rs
```diff
@@ -32,232 +31,207 @@ where
...
+        {   const PANIC_ZERO: &'static str = $zero:literal;
+            const PANIC_OVERFLOW: &'static str = $overflow:literal;
+            $simd_call:ident
+        },
+        $int:ident ) => {
+        if $rhs.lanes_eq(Simd::splat(0)).any() {
+            panic!($zero);
+        } else if <$int>::MIN != 0
+            && $lhs.lanes_eq(Simd::splat(<$int>::MIN)) & $rhs.lanes_eq(Simd::splat(-1 as _))
+                != Mask::splat(false)
```
calebzulawski, line 74: This should use `any()`

crates/core_simd/src/ops.rs
```diff
@@ -32,232 +31,207 @@ where
...
-            impl<const LANES: usize> core::ops::$trait<Self> for Simd<$scalar, LANES>
-            where
-                LaneCount<LANES>: SupportedLaneCount,
-            {
-                type Output = Self;
+macro_rules! for_base_types {
+    (   T = ($($scalar:ident),*);
+        type Lhs = Simd<T, N>;
+        type Rhs = Simd<T, N>;
+        type Output = $out:ty;
```
calebzulawski, line 87: Something slightly bothers me about these types, but it doesn't really matter I suppose

calebzulawski (changes requested):

crates/core_simd/src/ops.rs
workingjubilee: oh duh.

calebzulawski:
Looks good to me, but do you think this is actually clearer than the old macros?

workingjubilee:
I realize this may sound insane but:

Yep.
While the TT muncher certainly... does a "trick"... when I tried to navigate the macros before, I found it difficult to find the exact point an impl (as in the fn expression body)  was placed, and where the impl (as in the actual line of code that declared it would be implemented) was located. This is why my initial impulse was to just carve up, flatten, and simplify as much as possible, even at the cost of extra lines.

Here, the `for_base_ops!` macro invocation puts the identifiers "of note" in the same block, and while tweaking the details "behind" things may require backing up a bit, changing an identifier in that one block... what I usually want to do when refactoring... instantly alters what is implemented for what.

calebzulawski (approved):

Issue #205, opened 2021-12-02
Scatter into `&mut [MaybeUninit<scalar>]`s instead of `&mut [scalar]`


fee1-dead:
Title. The current API only allows `scatter`ing to initialized buffers, but it shouldn't matter whether the buffer is actually initialized.

workingjubilee:
This is a reasonable request but I can already see "scatter should allow `MyPersonalUnionType<T: SimdElement>`", and I am not sure how to provide that at the moment. It could be done by exposing the Vector of Raw Pointers API, but that doesn't merely require a bikeshed to be painted. Rather, we are apparently commissioning an elaborate mural.

Hrm... Is this a safe function? It seems it has to be, as it fulfills the relevant validity invariant for an ambiguously-typed memory location, so we aren't accessing an invalid value, we're describing a valid value, and that is safe unless done through `*mut T`, which is only `unsafe` because it is dereferencing a raw pointer which doesn't know if it's obtained validly. Yes? So the gather equivalent may be `unsafe`, but the scatter would not be, although the current implementation wraps some very `unsafe` spooky actions.

Open to ideas. Maybe a `BorrowMut` constraint?

thomcc:
We discussed this elsewhere, but to update this issue: `MaybeUninit` is a `#[repr(transparent)]` union (unlike anything from users), and is special in certain ways.

IMO we should support this for `MaybeUninit<T>`, and not for arbitrary user types or at least, the request to support it for user types would be a separate request (one which personally I'd probably be against).

workingjubilee:
To elaborate a little, my current understanding:
Scatter functionally is a vectorization of
```rust
unsafe { u.as_mut_ptr().write(v); }
```
which requires the union to be `#[repr(transparent)]` or possibly `#[repr(C)]`, and preferably one with exactly equal size to the `T` in question. Anything validly writable would have to just be soundly transmutable to `[MaybeUninit<T>]`, in essence. So that settles "why this and not something else": any type that could be valid to write to must establish it can be reinterpreted as one of these two types anyways.

Allowing it with anything that doesn't fulfill that requires us to either make the function `unsafe` or somehow vectorize
```rust
*u = Union::new(val);
```
[...]

Issue #200, opened 2021-11-24
vectors of pointers


programmerjake:
I didn't see an issue tracking vectors of pointers/references, so I'm opening one. Feel free to edit as necessary.

programmerjake:
One thing that just occurred to me in relation to #195, #198, and #199:
We can't depend on user-implementable traits such as `Add` or `PartialOrd` when we eventually get `Simd<&T, N>` cuz references are `#[fundamental]` so users can implement `&T: Add` which would allow unsafely calling `simd_reduce_add` with argument type `&T`.

workingjubilee:
...**if** we get `Simd<&T, N>`. That is the best argument I have heard for not allowing so.

calebzulawski:
Or, we should _not_ use the ops traits on the scalar type and instead use our own sealed traits that explicitly indicate which functions can monomorphize (as I have said repeatedly), and then we can allow vectors of references just fine.  The ops traits do not actually mean anything whatsoever regarding intrinsics and this is an example of why using them for this is a hack that only works with number types.  I think vectors of references would be useful for gather and maybe a few other things and this isn't a good reason to rule them out.

workingjubilee:
We can only use thin pointers anyway, though. If it's a fat borrow, and any instance of `&T` can be, it suddenly won't fit inside the same vector lanes as a thin `&T` and breaks all the invariants we have about scalars, including the internal logic `gather_select_unchecked` relies on.

programmerjake:
vectors of fat pointers should totally be doable by just allocating `i128` lanes or twice as many lanes or similar (maybe both at different times?), though, yeah, we'll probably at first want a `T: Thin + ?Sized` bound for `Simd<&'a T, N>`. Vectors of fat pointers will also need to deal with the possible future extensions of fat pointers to more arbitrary metadata types for `&dyn Debug + Display` or custom metadata.

Lokathor:
what would a vector of fat pointers even do. Like are we trying to imagine that at some point you could vectorize all of rust, or what.

programmerjake:
One example where vectors of references would be nice:
```rust
pub struct TreeNode<'a> {
    data: Cell<isize>,
    left: &'a TreeNode<'a>,
    right: &'a TreeNode<'a>,
}

[...]

PR #213, opened 2021-12-12
Enable scatters with `&mut [MaybeUninit<T>]`

workingjubilee:
This PR makes `unsafe fn scatter_select_unchecked` more unsafe, by allowing it to accept a raw `*mut T`, and thus exposing all the invariants underneath the previous function but also allowing more uses. It then adds `scatter_uninit` as a variant on `scatter_select`, for the safe scatter to `&mut [MaybeUninit<T>]`.

This closes rust-lang/portable-simd#205.

### DIFF
--- a/crates/core_simd/src/vector.rs
+++ b/crates/core_simd/src/vector.rs
@@ -238,19 +238,42 @@ where
         idxs: Simd<usize, LANES>,
     ) {
         let enable: Mask<isize, LANES> = enable & idxs.lanes_lt(Simd::splat(slice.len()));
-        // SAFETY: We have masked-off out-of-bounds lanes.
-        unsafe { self.scatter_select_unchecked(slice, enable, idxs) }
+        // SAFETY: Out-of-bounds lanes have been masked, and the ptr is immediately materialized.
+        unsafe { self.scatter_select_unchecked(slice.as_mut_ptr(), enable, idxs) }
     }
 
-    /// Writes the values in a SIMD vector to multiple potentially discontiguous indices in `slice`.
+    /// Writes the values in a SIMD vector to multiple potentially discontiguous locations,
+    /// offset from the `ptr` provided by indices, assuming it is a base ptr to `[T]`.
     /// The mask `enable`s all `true` lanes and disables all `false` lanes.
     /// If two enabled lanes in the scattered vector would write to the same index,
     /// only the last lane is guaranteed to actually be written.
     ///
     /// # Safety
     ///
-    /// Calling this function with an enabled out-of-bounds index is *[undefined behavior]*,
-    /// and may lead to memory corruption.
+    /// - Calling this function with any `*mut T` that is not a base pointer to `[T]` or a type
+    ///   with equivalent layout (such as arrays or SIMD types) results in *[undefined behavior]*.
+    /// - Calling this function with an `enable`d out-of-bounds index results in *[undefined behavior]*.
+    /// - If using `&mut [T]` or similar, such as `Vec<T>` or `&mut [T; N]`, it is safest to
+    /// construct the bounds mask first if applicable and then use `fn as_mut_ptr(&mut self)`
+    /// to obtain *mut T` immediately as part of this method call, with nothing else between,
+    /// as even deriving `&[T]` from by calling `fn len` creates an `&[T]` which,
+    /// as an immutable borrow derived from a mutable (and therefore unique) borrow,
+    /// interacts in a way that invalidates then-live raw `*mut T`s according to the
+    /// rules of "Stacked Borrows", which can lead to *[undefined behavior]*.
+    ///
+    /// The last safety clause does not apply directly to raw pointers like `*mut T`, but
+    /// involves the semantics of borrows like `&mut [T]` and `&[T]`, so it does not apply
+    /// to instances of `*mut T` that are never converted to or from `&mut [T]`.
+    /// This includes `*mut T` only used in FFI with languages which lack borrow semantics.
+    ///
+    /// Undefined behavior allows memory corruption and invalidates the safety assumptions
+    /// safe Rust code and optimizing compilers may rely on.
+    ///
+    /// The guidelines for using this unstable, unsafe function may change in the future,
+    /// as the precise semantics of unsafe code in Rust are underspecified.
+    /// The caveats here are expressed restrictively to allow calling this function
+    /// with minimal likelihood that callers have to be rewritten in the future,
+    /// **IF** they are correct according to these guidelines.
     ///
     /// # Examples
     /// ```
@@ -265,7 +288,7 @@ where
     /// let enable = enable & idxs.lanes_lt(Simd::splat(vec.len()));
     ///
     /// // We have masked the OOB lane, so it's safe to scatter now.
-    /// unsafe { vals.scatter_select_unchecked(&mut vec, enable, idxs); }
+    /// unsafe { vals.scatter_select_unchecked(vec.as_mut_ptr(), enable, idxs); }
     /// // index 0's second write is masked, thus was omitted.
     /// assert_eq!(vec, vec![-41, 11, 12, 82, 14, 15, 16, 17, 18]);
     /// ```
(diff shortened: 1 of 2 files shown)


crates/core_simd/src/vector.rs
```diff
@@ -238,19 +240,42 @@ where
...
+    /// - Calling this function with any `*mut T` that is not a base pointer to `[T]` or a type
+    ///   with equivalent layout (such as arrays or SIMD types) results in *[undefined behavior]*.
+    /// - Calling this function with an `enable`d out-of-bounds index results in *[undefined behavior]*.
+    /// - If using `&mut [T]` or similar, such as `Vec<T>` or `&mut [T; N]`, it is safest to
+    /// construct the bounds mask first if applicable and then use `fn as_mut_ptr(&mut self)`
+    /// to obtain *mut T` immediately as part of this method call, with nothing else between,
+    /// as even deriving `&[T]` from by calling `fn len` creates an `&[T]` which,
+    /// as an immutable borrow derived from a mutable (and therefore unique) borrow,
+    /// interacts in a way that invalidates then-live raw `*mut T`s according to the
+    /// rules of "Stacked Borrows", which can lead to *[undefined behavior]*.
```
calebzulawski, line 264: I'm still not really convinced this is currently the case, rather than a proposed rule...

thomcc: Right, stacked borrows is a proposed aliasing model, rather than an accepted one. There are things in the language that it does not properly model, and so it probably cannot be the final one. I think we can probably be a lot more vague about what is UB here, though.

Also, I'd expect it to have rules derived from a combination of the rules of https://doc.rust-lang.org/std/ptr/fn.write.html (which heavily leans on https://doc.rust-lang.org/std/ptr/index.html) and the rules of https://doc.rust-lang.org/std/primitive.pointer.html#method.offset

crates/core_simd/src/vector.rs
```diff
@@ -265,38 +290,76 @@ where
...
+        // 3. a *mut T which is our base ptr.
+        // The user is warned about this in the documentation for this function, and hopefully
+        // is either playing entirely with raw ptrs or calls `as_mut_ptr` immediately before.
         unsafe {
             // Now Entering ☢️ *mut T Zone
-            let base_ptr = crate::simd::ptr::SimdMutPtr::splat(slice.as_mut_ptr());
+            let base_ptr = crate::simd::ptr::SimdMutPtr::splat(ptr);
             // Ferris forgive me, I have done pointer arithmetic here.
             let ptrs = base_ptr.wrapping_add(idxs);
             // The ptrs have been bounds-masked to prevent memory-unsafe writes insha'allah
```
programmerjake, line 320: this comment line no longer matches the code

crates/core_simd/src/vector.rs
workingjubilee: As there is no rule that is known to be true,
there is also no rule that is certain to be false.
therefore, in the interests of users not having to rewrite their code
we should advise them to duck the maybe-brick that may or may not collide with their skull.

It is true the user's cat may still be alive if we open the box,
or perhaps the user will be a butterfly.

workingjubilee: Ah yes, I need to specify the aligned/non-null clause.

calebzulawski: I think it is even more confusing to bring up rules that don't currently exist.  This is certainly more strict than LLVM's rules so I don't think it's right to portray it as the current state of rust.  In my opinion scatter is just one of many, many APIs that use pointers and isn't special enough to warrant any explanation at all.  I think the simplest and most correct way to document this is to specify that it follows all the same requirements as `ptr::write` (whatever those may be at any particular point in time).

calebzulawski:
Maybe unchecked gather should use a pointer, as well?

programmerjake:
I quite like the new scatter API, cuz it's general enough to let a user construct their own poor-man's vector of pointers:
```rust
pub struct SimdPtr<T, const N: usize>(Simd<usize, N>, PhantomData<*mut T>);

impl SimdPtr<T, const N: usize> where ... {
    pub unsafe fn scatter_store(self, values: Simd<T, N>, mask: Mask<isize, N>) {
        values.scatter_select_unchecked(null_mut(), mask, self.0);
    }
}
```

crates/core_simd/src/vector.rs
workingjubilee: >  This is certainly more strict than LLVM's rules 
Correct.

But one should not backpropagate from LLVM's rules, which are themselves ambiguous, poorly specified, and likely to become more strict in the future, to the Rust programming language, for the same reason one should not backpropagate rules from x86 to the C programming language. C is certainly more restrictive in many ways than x86 is, often.

I also think that it's likely that Rust will have a more lax model in the future.

But as far as I have observed, the users of unsafe code complain most about the prospect of having to rewrite that code based on future model changes. Specifying this in the way least likely to be invalidated seems soundest, even if it is arguably "most incorrect", as it takes the upper bound on constraints, rather than an average.

thomcc: It's problematic for vector scatters/gathers to have stricter rules than then underlying operations. It makes it harder to verify if vectorized code is correct, and it would mean that things that analyze UB (which includes, but is not limited to miri, CC @RalfJung) have to special-case this, or have undetectable UB (which is undesirable).

That is, assuming these are intended to be language-level UB rather than library-level UB (which seems to be the case)

workingjubilee: It's not a stricter rule than the underlying operations, it's that the most obvious ways to set up the environment **before** you call the function have a gotcha if you do them out of order. This documentation actually has nothing to do with the function itself, but has everything to do with me trying to address that gotcha in the documentation. I am not interested, however, in entertaining an argument here over this. I am withdrawing this PR.

crates/core_simd/src/vector.rs
```diff
@@ -265,38 +290,76 @@ where
...
         enable: Mask<isize, LANES>,
         idxs: Simd<usize, LANES>,
     ) {
-        // SAFETY: This block works with *mut T derived from &mut 'a [T],
+        // SAFETY: This block works with a *mut T that may be from &mut 'a [T],
         // which means it is delicate in Rust's borrowing model, circa 2021:
-        // &mut 'a [T] asserts uniqueness, so deriving &'a [T] invalidates live *mut Ts!
-        // Even though this block is largely safe methods, it must be exactly this way
-        // to prevent invalidating the raw ptrs while they're live.
+        // &mut 'a [T] asserts uniqueness, so deriving &'a [T] invalidates live *mut T!
```
RalfJung, line 307: I don't quite understand this comment.

`deriving &'a [T]` *from the original `&mut` that sources this `*mut` would invalidate the `*mut`, but we cannot even do that here since we do not have access to that original `&mut`. As long as we work with raw pointers only, no aliasing restrictions arise.

I am probably suffering from a lack of imagination here -- what is the alternative version of this code that would be wrong?

crates/core_simd/src/vector.rs
RalfJung: FWIW I agree that ideally the safety contract here would be stated in the same terms as the `core::ptr` functions. If that seems insufficiently clear, we should probably clarify those rules.

> It's not a stricter rule than the underlying operations, it's that the most obvious ways to set up the environment before you call the function have a gotcha if you do them out of order. 

Looks like it's probably not worth further discussing this PR, but for the future -- I think what I am missing here is what that most obvious way looks like. It would help to have a concrete example. If there is something about SIMD code that makes this more likely to be a problem than scalar code, I would understand better why you are suggesting SIMD should document this more carefully than the scalar functions we have.

crates/core_simd/src/vector.rs
workingjubilee: This is a leftover comment from the code on main in fact accepting `&mut [T]`.

crates/core_simd/src/vector.rs
workingjubilee: You yourself admonished me:
image

RalfJung: So in the usual way of calling this raw-ptr-taking function, there is a call to `len` involved in a way that is problematic?

Basically I don't quite understand what *concrete* problem these comments are referring to. The block of code "must be exactly this way" to make sure things work out, but what would the other way(s) even be? It is not possible to reorder these lines since each uses a result just computed in the previous line.

Lokathor: I think that Jubilee is referring to the general fact that Stacked Borrows makes *all* interaction with a mutable slice (including this example here) into a general footgun if the user makes the mutable pointer first and then gets the length of the slice after that. Since people generally consider slices to be "a pointer and a length", they are often inclined to write the code similarly:
```rust
let p = my_slice.as_mut_ptr();
let len = my_slice.len();
```
and now they've blown off their foot because `p` is invalid.

It's nothing particular to this function, and nothing to do with simd. Stacked Borrows just has a footgun for all mutable slices.

workingjubilee: Yes, if a caller starts from a Vec (which is to be expected) then they may call `scatter_select_unchecked` in a way that would mimic the following pattern:
```rust
    let mut v = vec![4, 5, 6, 7];
    let mut x = v.as_mut_ptr();
    unsafe {
        x = x.offset(1);
        if x < v.as_mut_ptr().add(v.len()) {
            x.write(0);
        }
    }
```
Because why wouldn't they?

workingjubilee: Lokathor is correct, yes. I just think it is important to reiterate the warning in appropriately flamingly large text, but if I say that, then everyone complains at me that Miri doesn't issue a warning on that code.

The oddity of the (old) internal comments being distracting in the presence of the (new) external documentation is merely a relic of me focusing on trying to be agonizingly clear in the external documentation.

RalfJung: Okay, I see -- thanks.

I think people were mostly complaining that *because* this is such a general issue, it seems odd to have it explained in user-visible documentation in this particular function. If we want to warn people in the docs about these patterns which Stacked Borrows disallows (but which other future aliasing models might allow), we should probably do that more centrally and more consistently.

workingjubilee: Sure. But my primary concern isn't really if all of `core::ptr` is properly documented, it's that one method, `core::simd::Simd::scatter_select_unchecked`, is properly documented, one likely to be used by people who **don't** want to read a ton of general documentation.

If that is truly the concern, then we could have figured out where to move the docs from this method (and replace it with links to that location) after this PR.

RalfJung: > If that is truly the concern, then we could have figured out where to move the docs from this method (and replace it with links to that location) after this PR.

That seems to be assuming the consensus that we want to have Stacked Borrows specific docs in the official doc comments at all, which AFAIK is an open question.

workingjubilee: And so we come full circle to "But as far as I have observed, the users of unsafe code complain most about the prospect of having to rewrite that code based on future model changes."
 ¯\_(ツ)_/¯

fee1-dead:
> ```rust
> scatter_select_unchecked(null_mut()
> ```

This is unsound because you are producing a null `&mut [T]`. This isn't really a poorman's scatter to pointers, but a undefined behaviorman's scatter to pointers...

Please at least enable us to use raw pointers for this purpose..

calebzulawski:
That never creates any slice, let alone a null slice, so it's not UB.  Using a null base pointer is however incompatible with strict provenance.  We do intend on allowing vectors of raw pointers (see #200).

fee1-dead:
`scatter_select_unchecked` currently only accepts slices. so you can't do such a thing without causing UB either way.

https://github.com/rust-lang/portable-simd/blob/fcc5ca0f93e751f2e748c7654c78a541c0392575/crates/core_simd/src/vector.rs#L409-L411

PR #217, opened 2021-12-19
rebase "Add some safety comments"

workingjubilee:
This is a partial rebase of https://github.com/rust-lang/portable-simd/pull/182, I just wanted to see if this passes CI now.

### DIFF
--- a/crates/core_simd/src/comparisons.rs
+++ b/crates/core_simd/src/comparisons.rs
@@ -10,6 +10,8 @@ where
     #[inline]
     #[must_use = "method returns a new mask and does not mutate the original value"]
     pub fn lanes_eq(self, other: Self) -> Mask<T::Mask, LANES> {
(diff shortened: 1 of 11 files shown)


workingjubilee:
Caleb is going to follow up on this so closing.

workingjubilee:
This wound up falling behind again so reopening and rebasing, dropping dated remarks. This keeps not making it in due to changes, but I would rather have half of it than none of it, so I am going to merge this in unless @calebzulawski *strenuously* objects.

calebzulawski:
Yep, might as well merge it.

calebzulawski:
If you want, later today I can rebase the entire original branch, though, since I think this is only a small portion of the original comments.

workingjubilee:
A lot of the ones I dropped need to just be rewritten, so you're welcome to take a swing at that but I don't think it's actually worth thrashing git for it.

PR #219, opened 2021-12-22
impl std::simd::StdFloat

workingjubilee:
This introduces an extension trait to allow use of floating point methods that need runtime support. It is *excessively* documented because its mere existence is quite vexing, as the entire thing constitutes a leakage of implementation details into user observable space. Eventually the entire thing will ideally be folded into core and restructured to match the rest of the library, whatever that structure might look like at the time. This is preferred in lieu of the "lang item" path because any energy the lang items require (and it will be not-insignificant, by Simulacrum's estimation) is better spent on implementing our libmvec.

I have already tested a sync of this branch into std and it works fine.

For further (and purely optional) reading on "why this", see:
- https://github.com/rust-lang/portable-simd/issues/76
- https://github.com/rust-lang/portable-simd/issues/109
- https://github.com/rust-lang/rust/issues/50145
- https://github.com/rust-lang/rust/pull/63455

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,5 +2,6 @@
 
 members = [
     "crates/core_simd",
+    "crates/std_float",
     "crates/test_helpers",
 ]
--- a/crates/core_simd/Cargo.toml
+++ b/crates/core_simd/Cargo.toml
@@ -26,3 +26,6 @@ features = ["alloc"]
 
 [dev-dependencies.test_helpers]
 path = "../test_helpers"
+
+[dev-dependencies]
+std_float = { path = "../std_float/", features = ["as_crate"] }
--- a/crates/core_simd/examples/nbody.rs
+++ b/crates/core_simd/examples/nbody.rs
@@ -1,11 +1,13 @@
-#![cfg_attr(feature = "std", feature(portable_simd))]
+#![feature(portable_simd)]
+extern crate std_float;
 
 /// Benchmarks game nbody code
 /// Taken from the `packed_simd` crate
 /// Run this benchmark with `cargo test --example nbody`
-#[cfg(feature = "std")]
 mod nbody {
-    use core_simd::*;
+    use core_simd::simd::*;
+    #[allow(unused)] // False positive?
+    use std_float::StdFloat;
 
     use std::f64::consts::PI;
     const SOLAR_MASS: f64 = 4.0 * PI * PI;
@@ -167,7 +169,6 @@ mod nbody {
     }
 }
 
-#[cfg(feature = "std")]
 #[cfg(test)]
 mod tests {
     // Good enough for demonstration purposes, not going for strictness here.
@@ -184,7 +185,6 @@ mod tests {
 }
 
 fn main() {
-    #[cfg(feature = "std")]
     {
         let (energy_before, energy_after) = nbody::run(1000);
         println!("Energy before: {}", energy_before);
--- a/crates/std_float/src/lib.rs
+++ b/crates/std_float/src/lib.rs
@@ -1,0 +1,165 @@
+#![cfg_attr(feature = "as_crate", no_std)] // We are std!
+#![cfg_attr(
+    feature = "as_crate",
+    feature(platform_intrinsics),
+    feature(portable_simd)
+)]
+#[cfg(not(feature = "as_crate"))]
+use core::simd;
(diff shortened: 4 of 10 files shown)


Cargo.toml
```diff
@@ -2,5 +2,6 @@
 
 members = [
     "crates/core_simd",
+    "crates/std_float",
```
calebzulawski, line 5: Should this crate maybe be called `std_simd`?  It only has the extra float stuff for now, but it could end up with something unexpected.

crates/core_simd/Cargo.toml
```diff
@@ -26,3 +26,6 @@ features = ["alloc"]
 
 [dev-dependencies.test_helpers]
 path = "../test_helpers"
+
+[dev-dependencies]
+std_float = { path = "../std_float/" }
```
calebzulawski, line 31: I think we want to reverse this dependency, since we might want to put more complex functions in std that use `core::simd`.  Those tests would need to be moved, of course.

workingjubilee: It already has a dependency flowing the other way, actually, which is why this is a dev dependency: Cargo would not accept it if it was a true circular dependency. So this is only linked for running tests andexamples currently. In `./std_float/Cargo.toml`:
```toml
[dependencies]
core_simd = { path = "../core_simd" }
```

Cargo.toml
workingjubilee: It's not necessarily obvious we would want to keep anything else we include in `std` in the same crate and we'd likely have to adjust the paths for inclusion anyways (to use the more complicated scheme). It's fine to rename it if something does come up.

crates/core_simd/Cargo.toml
calebzulawski: Got it--makes sense

crates/core_simd/examples/nbody.rs
```diff
@@ -1,11 +1,17 @@
...
 /// Benchmarks game nbody code
 /// Taken from the `packed_simd` crate
 /// Run this benchmark with `cargo test --example nbody`
-#[cfg(feature = "std")]
 mod nbody {
-    use core_simd::*;
+    #[cfg(not(feature = "std"))]
+    use core::simd::*;
+    #[cfg(feature = "std")]
+    use core_simd::simd::*;
```
calebzulawski, line 12: what's going on here? shouldn't it always use `core_simd`--that's what we're testing

crates/std_float/src/lib.rs
```diff
@@ -0,0 +1,165 @@
+#![cfg_attr(feature = "as_crate", no_std)] // We are std!
```
calebzulawski, line 1: Why is this different than the `lib.rs`/`mod.rs` shenanigans in `core_simd`?

crates/core_simd/examples/nbody.rs
workingjubilee: Oh yeah. 🙃 Just was doing the same impl flagging dance everywhere. Thanks! (it... shouldn't really matter in practice.)

crates/std_float/src/lib.rs
workingjubilee: Since it's just a single file we can just name it easily, and it doesn't cause any "dancing".

But when it's not "actually a crate", it's not allowed to set `#![no_std]`. I could probably adjust things more to be less complicated in this case (truly build as a crate and just reimport from `std`?) but I followed the same model as with `core_simd` so as to minimize any possible dependency issues in the Rust build system, as I would rather explore that as cleanup afterwards.

calebzulawski (approved):

calebzulawski:
I think we'll probably revisit this once we figure out what we're doing with traits at a larger scale (this may end up getting renamed to `FloatExt` or something, if get a `core::simd::Float`?)

workingjubilee:
Hm. `StdFloat`, I think, would be most "appropriate", in that event.

workingjubilee:
Unresolved questions:
- [ ] Should `./crates/std_float` depend on details internal to `std`, or should it be a self-contained reexport?
- [ ] Will `StdFloat` coexist with some other trait?
- [ ] Will we want to add more to `std` specifically and use of this as a `std_simd` crate?
- [ ] How fast can we make this crate obsolete, rendering these questions irrelevant? :smiling_imp:

PR #232, opened 2022-01-27
Add Simd::cast

workingjubilee:
We've gotten a few requests for some kind of cast, and I needed a reason to force myself to work through learning about 2FA hardware keys. I know there's been some discussion about what exactly to do, but I believe we were drifting in the direction of providing this function?

Also, thanks @calebzulawski for implementing the saturating intrinsic!

### DIFF
--- a/crates/core_simd/src/intrinsics.rs
+++ b/crates/core_simd/src/intrinsics.rs
@@ -39,6 +39,9 @@ extern "platform-intrinsic" {
 
     /// fptoui/fptosi/uitofp/sitofp
     pub(crate) fn simd_cast<T, U>(x: T) -> U;
+    /// follows Rust's `T as U` semantics, including saturating float casts
+    /// which amounts to the same as `simd_cast` for many cases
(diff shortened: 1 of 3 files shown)


calebzulawski:
Could you add a test for this?  I know it uses the same conversion code as `as`, but I'd like to guarantee it.

programmerjake (approved):

calebzulawski (approved):

programmerjake:
you could remove `round_from_int` afaict because it's functionality is completely supplanted by `cast`

programmerjake:
also `to_int_unchecked` should be generic over return type

workingjubilee:
Yes, I was thinking "hm, do I want to touch `fn round_from_int`?" but then I realized I didn't have a good vibe anymore for how strong the consensus was around `fn round_from_int` and `fn to_int_unchecked`, so I was going to bring that up at our next sync.

programmerjake:
ok, let's just merge this PR as is then, and handle `to_int_unchecked` and `round_from_int` in separate PR(s).

PR #237, opened 2022-02-05
Removed need for generic_const_expr in bitmask

### LINKED ISSUE #223: Make `from/to_bitmask` not depend on `generic_const_exprs`
I.e. [these](https://github.com/rust-lang/portable-simd/blob/65cb2c90a0688c110d983a2dbb9932900cd6b5d9/crates/core_simd/src/masks/full_masks.rs#L115) are under `generic_const_exprs` which will take some time to stabilize.

Goal: find a way of re-writing those two functions without using `generic_const_exprs`

jorgecarleitao:
Closes #223

### DIFF
--- a/crates/core_simd/src/masks.rs
+++ b/crates/core_simd/src/masks.rs
@@ -217,18 +217,16 @@ where
     }
 
     /// Convert this mask to a bitmask, with one bit set per lane.
-    #[cfg(feature = "generic_const_exprs")]
     #[inline]
     #[must_use = "method returns a new array and does not mutate the original value"]
-    pub fn to_bitmask(self) -> [u8; LaneCount::<LANES>::BITMASK_LEN] {
+    pub fn to_bitmask(self) -> <LaneCount<LANES> as SupportedLaneCount>::BitMask {
         self.0.to_bitmask()
     }
 
     /// Convert a bitmask to a mask.
-    #[cfg(feature = "generic_const_exprs")]
     #[inline]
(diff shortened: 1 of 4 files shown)


jorgecarleitao:
I am not sure if this is sufficient - I can't find the tests where we cover compiling under certain feature sets.

workingjubilee:
[Most of it is a simple feature-test on GitHub Actions.](https://github.com/rust-lang/portable-simd/blob/4910274686bcd144228a04d8d4d5dece4c7f5e3d/.github/workflows/ci.yml#L242-L243)

You would need to remove the cfg from [inside the tests.](https://github.com/rust-lang/portable-simd/blob/081240a6637c1dec77e5451e36dac4fe67773464/crates/core_simd/tests/masks.rs#L71-L82)

You should also remove the feature entirely from the crate, as it is used nowhere else.

jorgecarleitao:
I think that the feature is still in use in another part of the crate (module `to_bytes`).

workingjubilee:
Ah, you are right actually. I forgot about that.
Also for some reason altering the ci.yaml stopped the tests from running entirely...

jorgecarleitao:
Sorry, that was a bad commit. Fixed it.

workingjubilee:
???
...I have no idea what's wrong here.

jorgecarleitao:
Closing in favor of  #238, since there is something going on with the CI in this one

PR #238, opened 2022-02-05
Removed need for generic_const_expr in bitmask

### LINKED ISSUE #223: Make `from/to_bitmask` not depend on `generic_const_exprs`

jorgecarleitao:
Closes #223

### DIFF
--- a/crates/core_simd/src/masks.rs
+++ b/crates/core_simd/src/masks.rs
@@ -217,18 +217,16 @@ where
     }
 
     /// Convert this mask to a bitmask, with one bit set per lane.
-    #[cfg(feature = "generic_const_exprs")]
     #[inline]
     #[must_use = "method returns a new array and does not mutate the original value"]
-    pub fn to_bitmask(self) -> [u8; LaneCount::<LANES>::BITMASK_LEN] {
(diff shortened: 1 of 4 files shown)


workingjubilee:
Problem is: GitHub itself. Cool.
https://www.githubstatus.com/

calebzulawski:
I should mention I took a stab at a more complete solution for this in the `feature/better-bitmask` branch, but was blocked by an unusual const generics issue at the time (no clue if that's still the case)

calebzulawski:
I just posted #239 with the changes I was talking about, we can compare these two PRs.

jorgecarleitao:
fwiw, I am fine with any. This is hopefully the last blocking issue forbiding arrow2 from adopting portable-simd. So either PR is fine - I would just like to give this a spin without the major performance regression vs packed_simd.

workingjubilee:
Resolved in #239.

PR #239, opened 2022-02-06
Bitmask conversion trait

calebzulawski:
Another approach that fixes rust-lang/portable-simd#223, as an alternative to rust-lang/portable-simd#238.

This adds the `ToBitMask` trait, which is implemented on a vector for each bitmask type it supports.  This includes all unsigned integers with enough bits to contain it.  The byte array variant has been separated out for now into rust-lang/portable-simd#246 and still requires `generic_const_exprs`, but the integer variants no longer require it and can make it to nightly.

### DIFF
--- a/crates/core_simd/src/masks/bitmask.rs
+++ b/crates/core_simd/src/masks/bitmask.rs
@@ -1,7 +1,7 @@
 #![allow(unused_imports)]
 use super::MaskElement;
 use crate::simd::intrinsics;
-use crate::simd::{LaneCount, Simd, SupportedLaneCount};
+use crate::simd::{LaneCount, Simd, SupportedLaneCount, ToBitMask};
 use core::marker::PhantomData;
 
 /// A mask where each lane is represented by a single bit.
@@ -116,13 +116,20 @@ where
     }
 
     #[inline]
-    pub unsafe fn to_bitmask_integer<U>(self) -> U {
+    pub fn to_bitmask_integer<U>(self) -> U
+    where
+        super::Mask<T, LANES>: ToBitMask<BitMask = U>,
+    {
+        // Safety: these are the same types
         unsafe { core::mem::transmute_copy(&self.0) }
     }
 
-    // Safety: U must be the integer with the exact number of bits required to hold the bitmask for
     #[inline]
-    pub unsafe fn from_bitmask_integer<U>(bitmask: U) -> Self {
+    pub fn from_bitmask_integer<U>(bitmask: U) -> Self
+    where
+        super::Mask<T, LANES>: ToBitMask<BitMask = U>,
+    {
+        // Safety: these are the same types
         unsafe { Self(core::mem::transmute_copy(&bitmask), PhantomData) }
     }
 
--- a/crates/core_simd/src/masks/bitmask.rs
+++ b/crates/core_simd/src/masks/bitmask.rs
@@ -115,20 +115,14 @@ where
         unsafe { Self(intrinsics::simd_bitmask(value), PhantomData) }
     }
 
-    #[cfg(feature = "generic_const_exprs")]
     #[inline]
-    #[must_use = "method returns a new array and does not mutate the original value"]
-    pub fn to_bitmask(self) -> [u8; LaneCount::<LANES>::BITMASK_LEN] {
-        // Safety: these are the same type and we are laundering the generic
+    pub unsafe fn to_bitmask_intrinsic<U>(self) -> U {
         unsafe { core::mem::transmute_copy(&self.0) }
     }
 
-    #[cfg(feature = "generic_const_exprs")]
     #[inline]
-    #[must_use = "method returns a new mask and does not mutate the original value"]
-    pub fn from_bitmask(bitmask: [u8; LaneCount::<LANES>::BITMASK_LEN]) -> Self {
-        // Safety: these are the same type and we are laundering the generic
-        Self(unsafe { core::mem::transmute_copy(&bitmask) }, PhantomData)
+    pub unsafe fn from_bitmask_intrinsic<U>(bitmask: U) -> Self {
+        unsafe { Self(core::mem::transmute_copy(&bitmask), PhantomData) }
     }
 
     #[inline]
--- a/crates/core_simd/src/masks/bitmask.rs
+++ b/crates/core_simd/src/masks/bitmask.rs
@@ -116,12 +116,13 @@ where
     }
 
     #[inline]
-    pub unsafe fn to_bitmask_intrinsic<U>(self) -> U {
+    pub unsafe fn to_bitmask_integer<U>(self) -> U {
         unsafe { core::mem::transmute_copy(&self.0) }
     }
 
+    // Safety: U must be the integer with the exact number of bits required to hold the bitmask for
     #[inline]
-    pub unsafe fn from_bitmask_intrinsic<U>(bitmask: U) -> Self {
+    pub unsafe fn from_bitmask_integer<U>(bitmask: U) -> Self {
         unsafe { Self(core::mem::transmute_copy(&bitmask), PhantomData) }
     }
 
--- a/crates/core_simd/tests/masks.rs
+++ b/crates/core_simd/tests/masks.rs
@@ -68,16 +68,16 @@ macro_rules! test_mask_api {
                 assert_eq!(core_simd::Mask::<$type, 8>::from_int(int), mask);
             }
 
-            #[cfg(feature = "generic_const_exprs")]
             #[test]
             fn roundtrip_bitmask_conversion() {
+                use core_simd::ToBitMask;
                 let values = [
                     true, false, false, true, false, false, true, false,
                     true, true, false, false, false, false, false, true,
                 ];
                 let mask = core_simd::Mask::<$type, 16>::from_array(values);
-                let bitmask = mask.to_bitmask();
-                assert_eq!(bitmask, [0b01001001, 0b10000011]);
+                let bitmask: u16 = mask.to_bitmask();
+                assert_eq!(bitmask, 0b1000001101001001);
                 assert_eq!(core_simd::Mask::<$type, 16>::from_bitmask(bitmask), mask);
             }
         }
--- a/crates/core_simd/tests/masks.rs
+++ b/crates/core_simd/tests/masks.rs
@@ -76,7 +76,7 @@ macro_rules! test_mask_api {
                     true, true, false, false, false, false, false, true,
                 ];
                 let mask = core_simd::Mask::<$type, 16>::from_array(values);
-                let bitmask: u16 = mask.to_bitmask();
+                let bitmask = mask.to_bitmask();
                 assert_eq!(bitmask, 0b1000001101001001);
                 assert_eq!(core_simd::Mask::<$type, 16>::from_bitmask(bitmask), mask);
             }
--- a/crates/core_simd/src/masks.rs
+++ b/crates/core_simd/src/masks.rs
@@ -12,8 +12,10 @@
 )]
 mod mask_impl;
 
-use crate::simd::intrinsics;
-use crate::simd::{LaneCount, Simd, SimdElement, SupportedLaneCount};
+mod to_bitmask;
+pub use to_bitmask::ToBitMask;
+
+use crate::simd::{intrinsics, LaneCount, Simd, SimdElement, SupportedLaneCount};
 use core::cmp::Ordering;
 use core::{fmt, mem};
 
@@ -214,22 +216,6 @@ where
         unsafe {
             self.set_unchecked(lane, value);
         }
-    }
-
-    /// Convert this mask to a bitmask, with one bit set per lane.
-    #[cfg(feature = "generic_const_exprs")]
-    #[inline]
-    #[must_use = "method returns a new array and does not mutate the original value"]
-    pub fn to_bitmask(self) -> [u8; LaneCount::<LANES>::BITMASK_LEN] {
-        self.0.to_bitmask()
-    }
-
-    /// Convert a bitmask to a mask.
-    #[cfg(feature = "generic_const_exprs")]
-    #[inline]
-    #[must_use = "method returns a new mask and does not mutate the original value"]
-    pub fn from_bitmask(bitmask: [u8; LaneCount::<LANES>::BITMASK_LEN]) -> Self {
-        Self(mask_impl::Mask::from_bitmask(bitmask))
     }
 
     /// Returns true if any lane is set, or false otherwise.
--- a/crates/core_simd/src/masks/full_masks.rs
+++ b/crates/core_simd/src/masks/full_masks.rs
@@ -110,13 +110,15 @@ where
     }
 
     #[inline]
-    pub unsafe fn to_bitmask_intrinsic<U>(self) -> U {
+    pub unsafe fn to_bitmask_integer<U>(self) -> U {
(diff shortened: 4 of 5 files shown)


calebzulawski:
The array implementation produces this warning, which I think is a false alarm (`const_evaluatable_unchecked` is gone, and we have `generic_const_exprs` enabled):
```
warning: cannot use constants which depend on generic parameters in types
   --> crates/core_simd/src/masks/to_bitmask.rs:93:24
    |
93  |                   self.0.to_bitmask()
    |                          ^^^^^^^^^^
...
106 | / impl_array_bitmask! {
107 | |     impl ToBitMask<[u8; _]> for Mask<_, 1>
108 | |     impl ToBitMask<[u8; _]> for Mask<_, 2>
109 | |     impl ToBitMask<[u8; _]> for Mask<_, 4>
...   |
113 | |     impl ToBitMask<[u8; _]> for Mask<_, 64>
114 | | }
    | |_- in this macro invocation
    |
    = note: `#[warn(const_evaluatable_unchecked)]` on by default
    = warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!
    = note: for more information, see issue #76200 <https://github.com/rust-lang/rust/issues/76200>
    = note: this warning originates in the macro `impl_array_bitmask` (in Nightly builds, run with -Z macro-backtrace for more info)

warning: cannot use constants which depend on generic parameters in types
   --> crates/core_simd/src/masks/to_bitmask.rs:97:22
    |
97  |                   Mask(mask_impl::Mask::from_bitmask(bitmask))
    |                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
...
106 | / impl_array_bitmask! {
107 | |     impl ToBitMask<[u8; _]> for Mask<_, 1>
108 | |     impl ToBitMask<[u8; _]> for Mask<_, 2>
109 | |     impl ToBitMask<[u8; _]> for Mask<_, 4>
...   |
113 | |     impl ToBitMask<[u8; _]> for Mask<_, 64>
114 | | }
    | |_- in this macro invocation
    |
    = warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!
    = note: for more information, see issue #76200 <https://github.com/rust-lang/rust/issues/76200>
    = note: this warning originates in the macro `impl_array_bitmask` (in Nightly builds, run with -Z macro-backtrace for more info)

warning: cannot use constants which depend on generic parameters in types
  |
  = warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!
  = note: for more information, see issue #76200 <https://github.com/rust-lang/rust/issues/76200>

```

calebzulawski:
Just realized CI fails since we use `-Dwarnings`.  Note that this warning only appears with the `generic_const_exprs` feature and shouldn't appear on nightly.

thomcc:
I don't really see why this needs to be a generic trait. It feels like a degree of freedom that gives very little and will make understanding the API more complex than it needs to be.

It seems like it mostly is this way to save a `as` or `.into()` call on the returned integer, which doesn't really feel like it's worth the indirection and API surface.

Concretely: is there a reason Bitmask can't just be an associated type on ToBitmask, which just is implemented once per vector with bitmask being the unsigned integer that fits all the bits? (at least for the current vector sizes)

thomcc:
I also kind of still oppose using `[u8; N]` for bitmasks for various reasons: byte-order, it seems to have confused most people who came across the existing API, and it won't easily support a lot of natural operations you want to perform on the resulting bitmask (like counting leading/trailing zeros/one and such)...

Is there a reason we want that? I also don't think it would be the end of the world if Simd<T, 256> had to use `[u8; N]` if we can't come up with a better approach by then (although there are for sure other options there too)

calebzulawski:
Well, we've had `[u8; N]` bitmasks for quite a while, this PR doesn't add them.  Yeah, the reason is so that you can use them for vectors wider than 64 lanes (There is u128, but we've had repeated issues with it so I'm not counting on it.  What about 512+ lanes?)

Regarding complexity, I'm not sure `ToBitMask<T>` is really any more complicated than `ToBitMask<BitMask=T>`?  The benefit of always offering arrays is being able to use bitmasks in function generic over lane count.

calebzulawski:
I did just consider, however, that we could offer two separate traits, one for conversion to integer bitmasks, and one to arrays.  Then it would look something like `ToBitMask<BitMask=u64>` and `ToBitMaskArray<const BYTES = 8>`.

thomcc:
I actually think that has a good chance of being less confusing, honestly.

(I do still think that `[u8; N]` would be very inconvenient to use in very many cases though...)

thomcc:
At the risk of regretting saying it, I suspect that `BitMask<const N: usize>` would be a lot better for code generic over lane count.

Like, I pretty much always want to figure out stuff like the leading/trailing zeros/ones, in order to find the index of highest/lowest lane (in whatever produced the mask) for which the predicate that generated thet BitMask returned true/false... (And a dozen other things that just suck to do with [u8; N]).

That said we already have many masks... (Actually, I wonder if those operations are just better on Mask<T, N>, and if they need to convert to bitmask internally to answer that question, they can, but aren't forced to)

I guess much of this came up in the The Great Simd Mask Bikeshed Of 2021...

calebzulawski:
If that's the specific operation you want, then you can cast the array to [u64; N] etc and run those operations.  Another common thing would be to match on specific values, which you could use the byte array directly for.  I think the point of providing it directly as a byte array is it's not restrictive about what you can do with it, for the same reason that we return integer bitmasks as integers at all and not just an opaque type.

workingjubilee:
I would like to see `core_simd` outmode the use of feature flags rather than require allowing warnings on some feature, since they are of no additional benefit for actual core inclusion.

calebzulawski:
In retrospect, it was pretty easy removing `generic_const_exprs` (from the bitmask stuff, `to_bytes` still needs it).  Sometimes takes a few tries...

calebzulawski:
Of course mips failed again, not just endianness but completely reversed bit order...

crates/core_simd/src/masks/bitmask.rs
```diff
@@ -115,22 +115,34 @@ where
         unsafe { Self(intrinsics::simd_bitmask(value), PhantomData) }
     }
 
-    #[cfg(feature = "generic_const_exprs")]
+    // Safety: N must be the exact number of bytes required to hold the bitmask for this mask
     #[inline]
     #[must_use = "method returns a new array and does not mutate the original value"]
-    pub fn to_bitmask(self) -> [u8; LaneCount::<LANES>::BITMASK_LEN] {
+    pub unsafe fn to_bitmask_array<const N: usize>(self) -> [u8; N] {
```
programmerjake, line 121: if you assert, you get a safe function afaict.
```suggestion
    pub fn to_bitmask_array<const N: usize>(self) -> [u8; N] {
        assert_eq!(N, LaneCount::<LANES>::BITMASK_LEN, "invalid bitmask array size");
```

crates/core_simd/src/masks/bitmask.rs
```diff
@@ -115,22 +115,34 @@ where
...
         // Safety: these are the same type and we are laundering the generic
         unsafe { core::mem::transmute_copy(&self.0) }
     }
 
-    #[cfg(feature = "generic_const_exprs")]
+    // Safety: N must be the exact number of bytes required to hold the bitmask for this mask
     #[inline]
     #[must_use = "method returns a new mask and does not mutate the original value"]
-    pub fn from_bitmask(bitmask: [u8; LaneCount::<LANES>::BITMASK_LEN]) -> Self {
+    pub unsafe fn from_bitmask_array<const N: usize>(bitmask: [u8; N]) -> Self {
```
programmerjake, line 129: likewise, assert can be added. this can't necessarily be safe, though, cuz we have to worry about what happens if the caller feeds garbage into the padding bits.

crates/core_simd/src/masks/bitmask.rs
```diff
@@ -115,22 +115,34 @@ where
...
         // Safety: these are the same type and we are laundering the generic
         Self(unsafe { core::mem::transmute_copy(&bitmask) }, PhantomData)
     }
 
+    // Safety: U must be the integer with the exact number of bits required to hold the bitmask for
+    #[inline]
+    pub unsafe fn to_bitmask_integer<U>(self) -> U {
```
programmerjake, line 136: we'd want a trait to bound this so it can be safe.

programmerjake:
> Of course mips failed again, not just endianness but completely reversed bit order...

icr where I posted the link to the relevant code in llvm, but llvm arbitrarily decided that bit order always follows byte endianness. it might be nice to mention that in the code comments somewhere

crates/core_simd/src/masks/bitmask.rs
programmerjake: having transmute_copy of a wrong-sized object in a code-path that's never executed is at least just as safe as calling `unreachable_unchecked` -- perfectly safe imho.

```rust
pub fn this_is_safe() {
    assert!(1 == 2);
    unsafe { unreachable_unchecked() }
}
```

workingjubilee: `to_bitmask_array` feels like a more confusing name than what we had before (i.e. giving a bitmask, but as bytes) since it may imply, since many masks are 1 byte per lane, somehow lane masks, but as a set of bytes...? I think the "bitmask" terminology doesn't adequately distinguish it from the "width = lane" mask, since those are also used in the same ways that "bitmasks" are used elsewhere in programming.

If we are going to have such a verbose method, maybe `to_bitmask_as_bytes`.

crates/core_simd/tests/masks.rs
```diff
@@ -68,18 +68,31 @@ macro_rules! test_mask_api {
...
             fn roundtrip_bitmask_conversion() {
+                use core_simd::ToBitMask;
                 let values = [
                     true, false, false, true, false, false, true, false,
                     true, true, false, false, false, false, false, true,
                 ];
                 let mask = core_simd::Mask::<$type, 16>::from_array(values);
                 let bitmask = mask.to_bitmask();
-                assert_eq!(bitmask, [0b01001001, 0b10000011]);
+                assert_eq!(bitmask, 0b1000001101001001);
```
workingjubilee, line 80: this is a nice win in ergonomics.

crates/core_simd/src/masks/bitmask.rs
```diff
@@ -115,22 +115,34 @@ where
...
+        unsafe { core::mem::transmute_copy(&self.0) }
+    }
+
+    // Safety: U must be the integer with the exact number of bits required to hold the bitmask for
+    #[inline]
+    pub unsafe fn from_bitmask_integer<U>(bitmask: U) -> Self {
+        unsafe { Self(core::mem::transmute_copy(&bitmask), PhantomData) }
+    }
```
workingjubilee, line 144: Publicly extending the API with a bunch of unsafe methods to implement what seemingly ought to be a simple operation is... less than ideal. I am also not certain that this is actually an inherently sound transmutation... it feels like it would break if `Simd<T, 17>` was legal (3 bytes), and I would prefer to not have to audit our internals later for such code if we relax the lane count restrictions, and while it is unsafe, an unsafe method that some masks could not use strikes me as an unnecessary footgun.

crates/core_simd/src/masks/bitmask.rs
workingjubilee, line 136:
```suggestion
    pub unsafe fn to_bitmask_int<U>(self) -> U {
```

crates/core_simd/src/masks/bitmask.rs
```diff
@@ -115,22 +115,34 @@ where
...
 
+    // Safety: U must be the integer with the exact number of bits required to hold the bitmask for
+    #[inline]
+    pub unsafe fn to_bitmask_integer<U>(self) -> U {
+        unsafe { core::mem::transmute_copy(&self.0) }
+    }
+
+    // Safety: U must be the integer with the exact number of bits required to hold the bitmask for
+    #[inline]
+    pub unsafe fn from_bitmask_integer<U>(bitmask: U) -> Self {
```
workingjubilee, line 142:
```suggestion
    pub unsafe fn from_bitmask_int<U>(bitmask: U) -> Self {
```

workingjubilee (commented):
I kind of want this, but I also kind of am starting to get the vibe of "did we lose the plot somewhere?" Insofar as whether this is an alternative or exception to the other PR, as I see it, there is no reason this PR cannot be rebased atop the other one, so I am inclined to say we should wave Jorge's through first, collect some user feedback, and work on this some more.

crates/core_simd/src/masks/bitmask.rs
calebzulawski: This is not public.

crates/core_simd/src/masks/bitmask.rs
calebzulawski: I'm okay with that name too, or any other suggestions.

workingjubilee:
My apologies, I would like to downscale all previous remarks in import and instead observe that this code induces an ICE when I issue the command `cargo doc`.

calebzulawski:
Actually, I realized I jumped the gun on that because of the limitations of our bitmask-layout masks, but let me explain what I mean:

`<LaneCount<LANES> as SupportedLaneCount>::BitMask` is _not_ user-friendly, it only exists there so that we can implement bitmask-layout masks with the limitations of today's const generics.  With that in mind, I don't think `to_bitmask()` should return that type.  The `ToBitMask::BitMask` type is what the user can directly reference as the return type, and even when const generics are improved and we can remove anything related to bitmasks from `LaneCount`, that type will remain.

The difference between this PR and #238 is this one doesn't expose that ugly implementation detail that we'd really like to remove ASAP.

crates/core_simd/src/masks/bitmask.rs
calebzulawski: I was able to use the `ToBitMask` trait directly to make this safe

crates/core_simd/src/masks/bitmask.rs
calebzulawski: As far as I can tell, it really doesn't matter what those padding bits are because they're entirely unused.

crates/core_simd/src/masks/bitmask.rs
calebzulawski: I'm not going to resolve this whole thread because of the naming issue, but the original concern about the function being safe has been addressed.

calebzulawski:
The commits got a little messed up since I removed a broken commit, but I was able to make all of the bitmask functions safe by making `ToBitMask` an unsafe trait that specifies the required bitmask types, and using that trait to determine correct bitmask types.

programmerjake (approved):
other than maybe renaming `ToBitmask` to `ToIntBitmask` since it only works for a few specific lane counts due to us not having generic integers yet so people will wonder why it doesn't work on `f32x4` or similar, this looks good to me!

workingjubilee:
Mmm, I tried building this on a subtree pull into std and I can confirm it breaks, so we can't land this as-is. It is almost entirely lamenting about the way `ToBitMaskArray` works. Possibly because it has to handle that during bootstrap...

If you split out the `ToBitMaskArray` impl and rebase this accordingly, we can probably push it in, otherwise not so much.
:thought_balloon: ( `ToBitArray::to_bitarray`? )

calebzulawski:
@workingjubilee I removed bitmask arrays from this PR, to be handled in #246

Issue #56, opened 2021-01-27
Should SIMD add/sub/etc. trigger panics on overflow in debug mode?


workingjubilee:
Per topic, this affects our vector implementations of Add, Sub, and Neg.

This came up during discussion for #54 and #55.

Lokathor:
I don't think so.

First of all a technical question: can rust code even cfg on overflow-checks being true or false?

Second: even if we could, we should not. it would make the debug code unacceptably slow. rust has a lot of things that make for slow debug code and long compiles in release, we should not compound the problem if we don't have to.

calebzulawski:
I do tend to agree that maybe SIMD should default to wrapping implementations (but provide checked member fns as well).  I would have left `Neg` alone in #54 if I hadn't misread it...

workingjubilee:
> First of all a technical question: can rust code even cfg on overflow-checks being true or false?

I do not believe so.

> Second: even if we could, we should not. it would make the debug code unacceptably slow. rust has a lot of things that make for slow debug code and long compiles in release, we should not compound the problem if we don't have to.

In the spirit of sounding out the alternatives (or at least warming up our counterarguments for when someone inevitably asks), I must note that scalar math triggers panics on overflows in debug mode. Why should this not be consistent? Yes, it is slower, but what special need is there to speed this up in debug mode versus any special need in scalar math?

Lokathor:
"because you only pick simd in the first place if you want to go faster"

but I'm quite biased. I've never ever wanted checked in debug wrapping in release anyway. it was a bad decision then, we shouldn't double down on it now.

calebzulawski:
I think I agree, the _only_ reason to use SIMD is for performance, so the default should probably be the most performant.  I tend to agree that wrapping would probably have been the better default.  When safety is involved I use `add_checked` or whatever anyway.  I'm probably biased in the same way.

thomcc:
> but I'm quite biased. I've never ever wanted checked in debug wrapping in release anyway. it was a bad decision then, we shouldn't double down on it now.

I don't agree with this, but I agree with the broad sentiment that it doesn't make sense for SIMD.

calebzulawski:
I do think it's worth pointing out that there are generic bit tricks you can do to produce relatively performant "any of" mask checks (and special instructions like movemask on some architectures).  The branch will really be what hurts, though, and that's unavoidable.

thomcc:
[...]

PR #243, opened 2022-02-08
Remove overflow panic from divrem and add basic docs to Simd<T, N>

workingjubilee:
This finishes normalizing `Simd<T, N>` to being approximately equivalent to `Simd<Wrapping<T>, N>` for all implemented operations I can think of. It also documents this fact, allowing this to close rust-lang/portable-simd#56.

### DIFF
--- a/crates/core_simd/src/vector.rs
+++ b/crates/core_simd/src/vector.rs
@@ -12,7 +12,39 @@ pub(crate) mod ptr;
 use crate::simd::intrinsics;
 use crate::simd::{LaneCount, Mask, MaskElement, SupportedLaneCount};
 
-/// A SIMD vector of `LANES` elements of type `T`.
+/// A SIMD vector of `LANES` elements of type `T`. `Simd<T, N>` has the same shape as [`[T; N]`](array), but operates like `T`.
+/// This type is commonly known by names like `f32x4` or `Vec4` in many programming languages.
+///
+/// Two vectors of the same type and length will, by convention, support the binary operations (+, *, etc.) that `T` does.
+/// These take the lanes at each index on the left-hand side and right-hand side, perform the binary operation,
+/// and return the result in the same lane in a vector of equal size. For a given operator, this is equivalent to zipping
+/// the two arrays together and mapping the operator over each lane.
+///
+/// ```rust
+/// # #![feature(array_zip, portable_simd)]
+/// # use core::simd::{Simd};
+/// let a0: [i32; 4] = [-2, 0, 2, 4];
+/// let a1 = [10, 9, 8, 7];
+/// let zm_add = a0.zip(a1).map(|(lhs, rhs)| lhs + rhs);
+/// let zm_mul = a0.zip(a1).map(|(lhs, rhs)| lhs * rhs);
+///
+/// // `Simd<T, N>` implements `From<[T; N]>
+/// let [v0, v1] = [a0, a1].map(|a| Simd::from(a));
+/// // Which means arrays implement `Into<Simd<T, N>>`.
+/// assert_eq!(v0 + v1, zm_add.into());
+/// assert_eq!(v0 * v1, zm_mul.into());
+/// ```
+///
+/// `Simd` with integers has the quirk that these operations are also inherently wrapping, as if `T` was [`Wrapping<T>`].
+/// Thus, `Simd` does not implement `wrapping_add`, because that is the behavior of the normal operation.
+/// This means there is no warning on overflows, even in "debug" builds.
+/// For most applications where `Simd` is appropriate, it is "not a bug" to wrap,
+/// and even "debug builds" are unlikely to tolerate the loss of performance.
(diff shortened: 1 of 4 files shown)


crates/core_simd/src/vector.rs
```diff
@@ -12,7 +12,39 @@ pub(crate) mod ptr;
 use crate::simd::intrinsics;
 use crate::simd::{LaneCount, Mask, MaskElement, SupportedLaneCount};
 
-/// A SIMD vector of `LANES` elements of type `T`.
+/// A SIMD vector of `LANES` elements of type `T`. `Simd<T, N>` has the same shape as [`[T; N]`](array), but operates like `T`.
+/// This type is commonly known by names like `f32x4` or `Vec4` in many programming languages.
```
calebzulawski, line 16: Not sure I would bother including this line, there are so many different languages with so many different names

crates/core_simd/src/vector.rs
```diff
@@ -12,7 +12,39 @@ pub(crate) mod ptr;
...
+///
+/// Two vectors of the same type and length will, by convention, support the binary operations (+, *, etc.) that `T` does.
```
calebzulawski, line 18: I know what you mean here, but binary might be confusing wrt bitwise operations, too.  I would maybe just say arithmetic operations.

crates/core_simd/src/vector.rs
```diff
@@ -12,7 +12,39 @@ pub(crate) mod ptr;
...
+/// ```rust
+/// # #![feature(array_zip, portable_simd)]
+/// # use core::simd::{Simd};
+/// let a0: [i32; 4] = [-2, 0, 2, 4];
+/// let a1 = [10, 9, 8, 7];
+/// let zm_add = a0.zip(a1).map(|(lhs, rhs)| lhs + rhs);
+/// let zm_mul = a0.zip(a1).map(|(lhs, rhs)| lhs * rhs);
+///
+/// // `Simd<T, N>` implements `From<[T; N]>
+/// let [v0, v1] = [a0, a1].map(|a| Simd::from(a));
```
calebzulawski, line 32:
```suggestion
/// let (v0, v1) = (Simd::from(a0), Simd::from(a1));
```

crates/core_simd/src/vector.rs
```diff
@@ -12,7 +12,39 @@ pub(crate) mod ptr;
...
+/// // Which means arrays implement `Into<Simd<T, N>>`.
+/// assert_eq!(v0 + v1, zm_add.into());
+/// assert_eq!(v0 * v1, zm_mul.into());
+/// ```
+///
+/// `Simd` with integers has the quirk that these operations are also inherently wrapping, as if `T` was [`Wrapping<T>`].
+/// Thus, `Simd` does not implement `wrapping_add`, because that is the behavior of the normal operation.
```
calebzulawski, line 39:
```suggestion
/// Thus, `Simd` does not implement `wrapping_add`, because that is the default behavior.
```

crates/core_simd/tests/ops_macros.rs
```diff
@@ -210,15 +210,21 @@ macro_rules! impl_signed_tests {
                     )
                 }
 
-            }
+                fn div_min_may_overflow<const LANES: usize>() {
+                    let a = Vector::<LANES>::splat(Scalar::MIN);
+                    let b = Vector::<LANES>::splat(-1);
+                    assert_eq!(a / b, a / (b * b));
```
calebzulawski, line 216:
```suggestion
                    assert_eq!(a / b, a);
```

crates/core_simd/tests/ops_macros.rs
```diff
@@ -210,15 +210,21 @@ macro_rules! impl_signed_tests {
...
+                }
 
-            test_helpers::test_lanes_panic! {
-                fn div_min_overflow_panics<const LANES: usize>() {
+                fn rem_min_may_overflow<const LANES: usize>() {
                     let a = Vector::<LANES>::splat(Scalar::MIN);
                     let b = Vector::<LANES>::splat(-1);
-                    let _ = a / b;
+                    assert_eq!(a % b, a % (b * b));
```
calebzulawski, line 222:
```suggestion
                    assert_eq!(a % b, Vector::<LANES>::splat(0));
```

calebzulawski (approved):
Looks good to me, just added some suggestions where I thought it could be a little clearer.

crates/core_simd/src/vector.rs
workingjubilee: :upside_down_face: Mmm. I am going to just say "operators", thinking about it, because some of the operators are **not** arithmetical...

crates/core_simd/src/vector.rs
workingjubilee: Hm. I was originally thinking "well, we use the names for our aliases..." but even that's not set in stone, and would be better pointed out in those terms if we did decide so, so I think you're right to be uncertain.

programmerjake (approved):
lgtm

PR #245, opened 2022-02-10
Explain unsafe contracts of core::simd

### LINKED ISSUE #244: docs for to_int_unchecked should mention the safe version
The docs for to_int_unchecked should mention that round is the safe version of the function.

workingjubilee:
This permeates the module with remarks on safety for pub methods,
layout of the Simd type, correct use of intrinsics, et cetera.
This is mostly to help others curious about how core::simd works,
including other Rust contributors, `unsafe` library authors,
and eventually ourselves.

This closes #244.

### DIFF
--- a/crates/core_simd/src/intrinsics.rs
+++ b/crates/core_simd/src/intrinsics.rs
@@ -10,7 +10,9 @@
 //! - undef: "a value that is every value". functionally like poison, insofar as Rust is concerned. poison may become this. note:
 //!   this means that division by poison or undef is like division by zero, which means it inflicts...
 //! - "UB": poison and undef cover most of what people call "UB". "UB" means this operation immediately invalidates the program:
-//!   in particular, LLVM is allowed to lower it to `ud2` or other opcodes that may cause an illegal instruction exception.
+//!   LLVM is allowed to lower it to `ud2` or other opcodes that may cause an illegal instruction exception, and this is the "good end".
+//!   The "bad end" is that LLVM may reverse time to the moment control flow diverged on a path towards undefined behavior,
+//!   and destroy the other branch, potentially deleting safe code and violating Rust's `unsafe` contract.
 //!
 //! Note that according to LLVM, vectors are not arrays, but they are equivalent when stored to and loaded from memory.
 //!
@@ -100,7 +102,7 @@ extern "platform-intrinsic" {
     /// ptr: vector of pointers to read from
     /// mask: a "wide" mask of integers, selects as if simd_select(mask, read(ptr), val)
     /// note, the LLVM intrinsic accepts a mask vector of <N x i1>
-    /// FIXME: consider reconciling this somehow if/when we fix up our mask story in general?
+    /// FIXME: review this if/when we fix up our mask story in general?
     pub(crate) fn simd_gather<T, U, V>(val: T, ptr: U, mask: V) -> T;
     /// llvm.masked.scatter
     /// like gather, but more spicy, as it writes instead of reads
--- a/crates/core_simd/src/intrinsics.rs
+++ b/crates/core_simd/src/intrinsics.rs
@@ -2,16 +2,29 @@
 //! crate.
 //!
 //! The LLVM assembly language is documented here: <https://llvm.org/docs/LangRef.html>
+//!
+//! A quick glossary of jargon that may appear in this module, mostly paraphrasing LLVM's LangRef:
+//! - poison: "undefined behavior as a value". specifically, it is like uninit memory (such as padding bytes). it is "safe" to create poison, BUT
+//!   poison MUST NOT be observed from safe code, as operations on poison return poison, like NaN. unlike NaN, which has defined comparisons,
+//!   poison is neither true nor false, and LLVM may also convert it to undef (at which point it is both). so, it can't be conditioned on, either.
+//! - undef: "a value that is every value". functionally like poison, insofar as Rust is concerned. poison may become this. note:
+//!   this means that division by poison or undef is like division by zero, which means it inflicts...
+//! - "UB": poison and undef cover most of what people call "UB". "UB" means this operation immediately invalidates the program:
+//!   in particular, LLVM is allowed to lower it to `ud2` or other opcodes that may cause an illegal instruction exception.
+//!
+//! Note that according to LLVM, vectors are not arrays, but they are equivalent when stored to and loaded from memory.
+//!
+//! Unless stated otherwise, all intrinsics for binary operations require SIMD vectors of equal types and lengths.
 
 /// These intrinsics aren't linked directly from LLVM and are mostly undocumented, however they are
-/// simply lowered to the matching LLVM instructions by the compiler.  The associated instruction
-/// is documented alongside each intrinsic.
+/// mostly lowered to the matching LLVM instructions by the compiler in a fairly straightforward manner.
+/// The associated LLVM instruction or intrinsic is documented alongside each Rust intrinsic function.
 extern "platform-intrinsic" {
     /// add/fadd
     pub(crate) fn simd_add<T>(x: T, y: T) -> T;
 
     /// sub/fsub
-    pub(crate) fn simd_sub<T>(x: T, y: T) -> T;
+    pub(crate) fn simd_sub<T>(lhs: T, rhs: T) -> T;
 
     /// mul/fmul
     pub(crate) fn simd_mul<T>(x: T, y: T) -> T;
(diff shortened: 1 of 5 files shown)


crates/core_simd/src/intrinsics.rs
```diff
@@ -71,19 +91,31 @@ extern "platform-intrinsic" {
...
+    // idx: LLVM calls it a "shuffle mask vector constant", a vector of i32s
     pub(crate) fn simd_shuffle<T, U, V>(x: T, y: T, idx: U) -> V;
 
+    /// llvm.masked.gather
+    /// like a loop of pointer reads
+    /// val: vector of values to select if a lane is masked
+    /// ptr: vector of pointers to read from
+    /// mask: a "wide" mask of integers, selects as if simd_select(mask, read(ptr), val)
+    /// note, the LLVM intrinsic accepts a mask vector of <N x i1>
+    /// FIXME: consider reconciling this somehow if/when we fix up our mask story in general?
```
calebzulawski, line 103: reconcile with?

crates/core_simd/src/round.rs
```diff
@@ -19,6 +19,11 @@ macro_rules! implement {
             /// * Not be NaN
             /// * Not be infinite
             /// * Be representable in the return type, after truncating off its fractional part
+            ///
+            /// If these requirements are infeasible or costly, consider using the safe function [cast],
```
calebzulawski, line 23: I think I would word it the other way really, prefer `cast` unless for some reason you absolutely know these are all true.

crates/core_simd/src/vector.rs
```diff
@@ -44,6 +44,39 @@ use crate::simd::{LaneCount, Mask, MaskElement, SupportedLaneCount};
 ///
 /// [`Wrapping<T>`]: core::num::Wrapping
 ///
+/// # Layout
+/// `Simd<T, N>` has a layout similar to `[T; N]` (identical "shapes"), but with a higher alignment.
```
calebzulawski, line 48:
```suggestion
/// `Simd<T, N>` has a layout similar to `[T; N]` (identical "shapes"), but with a greater alignment.
```

crates/core_simd/src/vector.rs
```diff
@@ -44,6 +44,39 @@ use crate::simd::{LaneCount, Mask, MaskElement, SupportedLaneCount};
...
+/// `[T; N]` is aligned to `T`, but `Simd<T, N>` will have an alignment based on both `T` and `N`.
+/// It is thus sound to transmute from `Simd<T, N>` into `[T; N]` but not vice versa.
+///
+/// # ABI Bugs
+/// Due to an unfortunate bug in the Rust ABI, `Simd<T, N>` does not participate in argument-passing in SIMD registers,
```
calebzulawski, line 53: I wouldn't consider this a bug so much as the current definition of the ABI... It's certainly intentional behavior, even if unfortunate.

crates/core_simd/src/vector.rs
```diff
@@ -44,6 +44,39 @@ use crate::simd::{LaneCount, Mask, MaskElement, SupportedLaneCount};
...
+/// except as an optimization. `#[inline]` hints are recommended on functions that accept `Simd<T, N>` or return it.
+/// This may be corrected in the future.
+///
+/// # Safe SIMD with Unsafe Rust
+///
+/// `Simd<T, N>` has a higher alignment requirement than an array. It is aligned based on the size of the entire vector,
+/// not individual elements. This means it may be sound to [`transmute`] `Simd<T, N>` to `[T; N]` but **not** vice versa.
```
calebzulawski, line 60: ditto from previous comments

crates/core_simd/src/vector.rs
```diff
@@ -44,6 +44,39 @@ use crate::simd::{LaneCount, Mask, MaskElement, SupportedLaneCount};
...
+/// When using `unsafe` Rust to read and write `Simd<T, N>` through [raw pointers], it is usually best to use
```
calebzulawski, line 61: I wouldn't say it is best, so much as it's something to be aware of.  I've definitely seem performance differences in some circumstances with aligned loads and stores on modern processors, not necessarily significant but there are times to use each IMO.

calebzulawski: If you're using slices as mentioned here, `align_to` may be both the safest and most performant option, for example.

crates/core_simd/src/vector.rs
workingjubilee: A feature is a bug we don't intend to correct,
a bug is a feature we intend to remove. :^)

crates/core_simd/src/vector.rs
workingjubilee: Mmm, I want to convey something like "try this first, rather than assuming it is automatically slower".

crates/core_simd/src/intrinsics.rs
```diff
@@ -2,16 +2,29 @@
...
 //!
 //! The LLVM assembly language is documented here: <https://llvm.org/docs/LangRef.html>
+//!
+//! A quick glossary of jargon that may appear in this module, mostly paraphrasing LLVM's LangRef:
+//! - poison: "undefined behavior as a value". specifically, it is like uninit memory (such as padding bytes). it is "safe" to create poison, BUT
+//!   poison MUST NOT be observed from safe code, as operations on poison return poison, like NaN. unlike NaN, which has defined comparisons,
+//!   poison is neither true nor false, and LLVM may also convert it to undef (at which point it is both). so, it can't be conditioned on, either.
+//! - undef: "a value that is every value". functionally like poison, insofar as Rust is concerned. poison may become this. note:
+//!   this means that division by poison or undef is like division by zero, which means it inflicts...
+//! - "UB": poison and undef cover most of what people call "UB". "UB" means this operation immediately invalidates the program:
```
programmerjake, line 12: UB can do more than act like a `ud2` trap...it can also time-travel to delete/modify code before the UB...basically LLVM will optimize assuming anything that causes UB can't ever occur, allowing LLVM to modify/delete whole sections of code.

crates/core_simd/src/vector.rs
```diff
@@ -44,6 +44,39 @@ use crate::simd::{LaneCount, Mask, MaskElement, SupportedLaneCount};
 ///
 /// [`Wrapping<T>`]: core::num::Wrapping
 ///
+/// # Layout
+/// `Simd<T, N>` has a layout similar to `[T; N]` (identical "shapes"), but with a higher alignment.
+/// `[T; N]` is aligned to `T`, but `Simd<T, N>` will have an alignment based on both `T` and `N`.
+/// It is thus sound to transmute from `Simd<T, N>` into `[T; N]` but not vice versa.
```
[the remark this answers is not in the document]
programmerjake: i think you're referring to casting/transmuting references/pointers to arrays/simd values, which is where the alignment of simd types comes in. the actual arrays/simd values themselves are values -- they aren't necessarily in memory so the concept of alignment doesn't necessarily apply to them. transmuting an array value from/to a simd value should be safe, cuz they are values, llvm/rustc conceptually moves them into memory with the correct alignment before anything can access them by reference.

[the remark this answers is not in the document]
workingjubilee: You're right, I forget it's equivalent to `memcpy`, just usually in-place.

crates/core_simd/src/vector.rs
calebzulawski: I still think it's misleading.  It's definitely not a mistake or unintended behavior, it's just a limitation.

crates/core_simd/src/vector.rs
calebzulawski: I think it would be fine to say exactly that :)

crates/core_simd/src/vector.rs
Lokathor: It's "unintended" in the sense that it's a non-zero cost fix for a compiler smartness limitation.

calebzulawski: Bug carries the connotation that your programs will not work, which is the opposite of what this does, it's specifically a solution for avoiding mismatched calling conventions, which _is_ a bug and will cause your programs to not work.

Lokathor: perhaps "problem with the Rust ABI"?

calebzulawski: I would just note that it's not optimal, but it's avoiding a catastrophic bug.

Lokathor: (i mean sorta, but doesn't it happen even when the calling conventions are known to match?)

calebzulawski: On architectures where they're known to match (like wasm) I don't believe they're passed by reference.  But on an architecture with multiple possible conventions, the callee can't possibly know all of it's caller's conventions.

workingjubilee:
```suggestion
/// # ABI "Features"
/// Due to Rust's safety guarantees, `Simd<T, N>` is passed to and from functions via memory, not SIMD registers, 
```

programmerjake: maybe phrase it "Due to Rust's safety guarantees, `Simd<T, N>` is *currently* passed to and from functions via memory, not SIMD registers,"

workingjubilee: And that's incorrect, Caleb: it happens on wasm too.

workingjubilee:
> On architectures where they're known to match (like wasm) I don't believe they're passed by reference. But on an architecture with multiple possible conventions, the callee can't possibly know all of it's caller's conventions.

My memory says this happens for wasm too. We did attempt to undo it for wasm but then had to revert that at some point to the memory-passing behavior due to it uncovering another bug.

crates/core_simd/src/vector.rs
```diff
@@ -44,6 +44,47 @@ use crate::simd::{LaneCount, Mask, MaskElement, SupportedLaneCount};
 ///
 /// [`Wrapping<T>`]: core::num::Wrapping
 ///
+/// # Layout
+/// `Simd<T, N>` has a layout similar to `[T; N]` (identical "shapes"), but with a greater alignment.
+/// `[T; N]` is aligned to `T`, but `Simd<T, N>` will have an alignment based on both `T` and `N`.
+/// It is thus sound to [`transmute`] `Simd<T, N>` to `[T; N]`, and will typically optimize to zero cost,
+/// but the reverse transmutation is more likely to require a copy the compiler cannot simply elide.
+///
+/// # ABI "Features"
```
calebzulawski, line 53: I might change this to "ABI Limitations" and "pass by reference" instead of "pass by memory", though both of these are correct too

calebzulawski:
There was a bug passing u128 by value in wasm, too, so maybe it's not the best example...

PR #246, opened 2022-02-12
Add bitmask array

calebzulawski:
A follow up on the somewhat problematic feature from #239 (based on that branch).

### DIFF
--- a/crates/core_simd/src/masks/full_masks.rs
+++ b/crates/core_simd/src/masks/full_masks.rs
@@ -2,7 +2,7 @@
 
 use super::MaskElement;
 use crate::simd::intrinsics;
-use crate::simd::{LaneCount, Simd, SupportedLaneCount, ToBitMask};
+use crate::simd::{LaneCount, Simd, SupportedLaneCount, ToBitMask, ToBitMaskArray};
 
 #[repr(transparent)]
 pub struct Mask<T, const LANES: usize>(Simd<T, LANES>)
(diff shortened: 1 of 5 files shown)


workingjubilee:
Aaaand my condolences.

calebzulawski:
Ok, the problem was just not gating enough on the `generic_const_exprs` feature, should be fine now I think.

crates/core_simd/src/masks/full_masks.rs
```diff
@@ -139,6 +142,74 @@ where
...
+    pub fn to_bitmask_array<const N: usize>(self) -> [u8; N]
+    where
+        super::Mask<T, LANES>: ToBitMaskArray,
+        [(); <super::Mask<T, LANES> as ToBitMaskArray>::BYTES]: Sized,
+    {
+        assert_eq!(<super::Mask<T, LANES> as ToBitMaskArray>::BYTES, N);
+
+        // Safety: N is the correct bitmask size
+        //
+        // The transmute below allows this function to be marked safe, since it will prevent
```
programmerjake, line 157: `transmute_copy` doesn't check size iirc, so the comment should be updated to not state that the transmutation is what's checking size.

programmerjake (changes requested):

crates/core_simd/src/masks/full_masks.rs
calebzulawski: Yep.  It's asserted above, so I just removed the explanation entirely.

programmerjake (approved):
lgtm

PR #250, opened 2022-03-01
Add bitmask i{N <8} -> u8 impls

workingjubilee:
...and copy the notes for why they're legal.
I think this should be fine to Just Do but I want to make sure that if this is wrong or problematic that someone pokes me about it soooooooooooooooo I guess I am poking everyone today:

cc @calebzulawski @programmerjake @RalfJung

### DIFF
--- a/crates/core_simd/src/intrinsics.rs
+++ b/crates/core_simd/src/intrinsics.rs
@@ -130,6 +130,14 @@ extern "platform-intrinsic" {
     pub(crate) fn simd_reduce_xor<T, U>(x: T) -> U;
 
     // truncate integer vector to bitmask
+    // `fn simd_bitmask(vector) -> unsigned integer` takes a vector of integers and
+    // returns either an unsigned integer or array of `u8`.
+    // Every element in the vector becomes a single bit in the returned bitmask.
+    // If the vector has less than 8 lanes, a u8 is returned with zeroed trailing bits.
+    // The bit order of the result depends on the byte endianness. LSB-first for little
+    // endian and MSB-first for big endian.
+    //
+    // UB if called on a vector with values other than 0 and -1.
     #[allow(unused)]
     pub(crate) fn simd_bitmask<T, U>(x: T) -> U;
 
--- a/crates/core_simd/src/masks/to_bitmask.rs
+++ b/crates/core_simd/src/masks/to_bitmask.rs
@@ -50,6 +50,9 @@ macro_rules! impl_integer_intrinsic {
 }
 
 impl_integer_intrinsic! {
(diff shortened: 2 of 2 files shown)


crates/core_simd/src/intrinsics.rs
```diff
@@ -130,6 +130,16 @@ extern "platform-intrinsic" {
...
+    // `fn simd_bitmask(vector) -> unsigned integer` takes a vector mask and
+    // returns the most significant bit (MSB) of each lane in the form
+    // of either:
+    // * an unsigned integer
+    // * an array of `u8`
+    // If the vector has less than 8 lanes, a u8 is returned with zeroed trailing bits.
+    //
+    // The bit order of the result depends on the byte endianness, LSB-first for little
+    // endian and MSB-first for big endian.
+    // Not intended to be called on a vector with values other than 0 and -1.
```
RalfJung, line 142: Is "Not intended to be called" any different from "it is UB to be called that way"?

If 0 and -1 are the only possible values, the "returns the most significant bit (MSB)" part seems unnecessarily specific -- basically this returns whether each lane is logically `true` or `false`, right?

workingjubilee: The commentary before that is a barely-trimmed copy of the comments from the implementation.
So at the current moment, it's implemented in such a way that it is **not** UB to do that.
"Not intended to be called" is me adding an ominous suggestion that I am reconsidering whether it should be well-defined or not.

calebzulawski: I think there's an unwritten contract that the mask arguments must always be valid masks.  We should probably just say "don't do it".  It could certainly end up being UB in another future backend.  All of the intrinsics are poorly documented but I don't think we want to give the impression you can pass weird inputs to them without it being UB.

calebzulawski (approved):

crates/core_simd/src/intrinsics.rs
RalfJung: Well, specifically when implementing them in Miri one needs to know what is and is not UB. :)

workingjubilee: I think so, yeah.

As far as I know Rust is currently against partially init data, so a u8 is either u8 or is MaybeUninit<u8>. I could see us getting uint<4> and then not worrying about the next 4 bits and making `simd_select_bitmask` legal on that, but as far as I understand the current Rust model for uninit data, the entire vector risks becoming uninit if we ever passed a single actually uninit bit into the mask for `simd_select_bitmask`.

workingjubilee: I checked and we're currently the only actual users of this intrinsic so yeah, we can just define it as UB. Bam.

PR #251, opened 2022-03-03
Add Mask::cast

workingjubilee:
Requested. Users are taking surprisingly generic tacks here so this seems warranted, even if it may or may not optimize well. Unfortunately, we can't simply just implement `From` in a way that yields this naturally because of the following conflict:

<pre><b>error[E0119]: conflicting implementations
of trait `core::convert::From&lt;core_simd::masks::Mask&lt;_, {_: usize}&gt;&gt;`
for type `core_simd::masks::Mask&lt;_, {_: usize}&gt;`</b>
   <b>--&gt; </b>crates/core_simd/src/masks.rs:579:1
    <b>|</b>
<b>579| /</b> impl&lt;T, U, const N: usize&gt; From&lt;Mask&lt;T, N&gt;&gt; for Mask&lt;U, N&gt;
<b>580| |</b> where
<b>581| |</b>     LaneCount&lt;N&gt;: SupportedLaneCount,
<b>582| |</b> {
<b>...|</b>
<b>585| |</b>     }
<b>586| |</b> }
    <b>| |_^</b>
    <b>|</b>
    <b>= note</b>: conflicting implementation in crate `core`:
            - impl&lt;T&gt; From&lt;T&gt; for T;

<b>For more information about this error, try `rustc --explain E0119`.</b>
<b>error:</b> could not compile `core_simd` due to previous error
</pre>

### DIFF
--- a/crates/core_simd/src/masks.rs
+++ b/crates/core_simd/src/masks.rs
@@ -181,6 +181,13 @@ where
         self.0.to_int()
     }
 
+    /// Converts the mask to a mask of any other lane size.
+    #[inline]
+    #[must_use = "method returns a new mask and does not mutate the original value"]
+    pub fn cast<U: MaskElement>(self) -> Mask<U, LANES> {
+        Mask(self.0.convert())
+    }
+
     /// Tests the value of the specified lane.
     ///
     /// # Safety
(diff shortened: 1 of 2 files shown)


programmerjake:
a simple test would probably be useful

PR #255, opened 2022-03-06
run Miri on CI (allowed to fail)

RalfJung:
With this (and an upcoming Miri PR), a number of tests are actually passing under Miri :D
```
cargo miri test --test i32_ops -- --skip cast --skip max --skip min --skip is_
cargo miri test --test f32_ops -- --skip cast --skip max --skip min --skip is_
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,6 +58,23 @@ jobs:
       - name: Run Clippy
         run: cargo clippy --all-targets --target ${{ matrix.target }}
 
+  miri:
+    name: "miri"
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install Miri
+        run: |
+          rustup toolchain install nightly --component miri
+          rustup override set nightly
+          cargo miri setup
+      - name: Test with Miri (failures allowed)
+        continue-on-error: true
+        run: |
+          cargo miri test --test i32_ops
+          cargo miri test --test f32_ops
+          cargo miri test --test cast
+
   x86-tests:
     name: "${{ matrix.target_feature }} on ${{ matrix.target }}"
     runs-on: ${{ matrix.os }}
--- a/crates/test_helpers/src/lib.rs
+++ b/crates/test_helpers/src/lib.rs
@@ -77,11 +77,21 @@ impl<T: core::fmt::Debug + DefaultStrategy, const LANES: usize> DefaultStrategy 
     }
 }
 
+#[cfg(not(miri))]
+fn make_runner() -> proptest::test_runner::TestRunner {
+    Default::default()
+}
+#[cfg(miri)]
+fn make_runner() -> proptest::test_runner::TestRunner {
+    // Only run a few tests on Miri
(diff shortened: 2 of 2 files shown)


calebzulawski (approved):

RalfJung:
I could also add a Miri job to CI? We probably want it to be 'allowed to fail' so that new intrinsics that Miri does not support yet can still easily be added, but it would be nice to somehow track whether the Miri state becomes worse.

RalfJung:
Not sure, I would guess around 20-30min for the entire test suite -- but I can't say yet because not all of it passes yet. ;)

.github/workflows/ci.yml
```diff
@@ -58,6 +58,28 @@ jobs:
...
+    name: "miri"
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install Miri
+        run: |
+          MIRI_NIGHTLY=nightly-$(curl -s https://rust-lang.github.io/rustup-components-history/x86_64-unknown-linux-gnu/miri)
+          echo "Installing latest nightly with Miri: $MIRI_NIGHTLY"
+          rustup default "$MIRI_NIGHTLY"
+          rustup component add miri
```
bjorn3, line 71: Would something like `rustup toolchain install nightly --component miri; rustup default nightly` work? I believe `rustup toolchain install` will fallback to older nightly versions if a component is unavailable.

RalfJung: Maybe? The custom Miri logic predates that rustup feature and it is hard to test.

I know that the script I used here definitely works. :)
Though I guess I could refrain from fixing the Miri toolstate today, and then we can test this tomorrow.

RalfJung: That is so much more complicated and verbose than the alternatives that I don't quite see why one would use this approach?

RalfJung: Yes that new snippet seems to work fine. :-)

RalfJung:
Ah, now Miri fails since it is still a day old and hence requires `const_fn_trait_bound`. ;) That should fix itself tomorrow though.

RalfJung:
```
error: unsupported operation: unimplemented intrinsic: simd_ceil
  --> /home/runner/work/portable-simd/portable-simd/crates/std_float/src/lib.rs:91:18
   |
91 |         unsafe { simd_ceil(self) }
   |                  ^^^^^^^^^^^^^^^ unimplemented intrinsic: simd_ceil
   |
```
Interesting, that intrinsic is not even declared in https://github.com/rust-lang/portable-simd/blob/master/crates/core_simd/src/intrinsics.rs. :thinking:

RalfJung:
But the tests that are now in there should work. It's a bit annoying to check this since "ignoring failures" means getting green checkmarks everywhere (with GHA there is no way to get a red checkmark for the job but still get a green checkmark for the PR as a whole) -- one has to look at the log to check if it actually failed. But that's the best we can do I guess.

workingjubilee:
Yes, that intrinsic is listed with the other "might libcall" floating point intrinsics.

https://github.com/rust-lang/portable-simd/blob/2d13059ae9162997c5b88997bda44443615f12ab/crates/std_float/src/lib.rs#L24-L44

PR #259, opened 2022-03-10
remove Miri from CI

RalfJung:
I will instead add testing at https://github.com/rust-lang/miri-test-libstd, where I will actually get notified when things fail.


workingjubilee:
`¯\_(ツ)_/¯`

PR #260, opened 2022-03-10
Added `.min` and `.max` for integers

### LINKED ISSUE #247: Confusing difference between float and integer in min and max
I tried this code:

```rust
let a = u32x4::from_array([0, 1, 0, 1]);
let b = u32x4::from_array([1, 0, 1, 0]);
print!("{:x?} ", a.min(b));

let a = f32x4::from_array([0., 1., 0., 1.]);
let b = f32x4::from_array([1., 0., 1., 0.]);
print!("{:x?}\n", a.min(b));
```

I expected to see this happen: `[0, 0, 0, 0] [0.0, 0.0, 0.0, 0.0]`.
`[0, 1, 0, 1] [0.0, 1.0, 0.0, 1.0]` would be a little surprising, because that isn't commonly available as native SIMD instruction.

Instead, this happened: `[0, 1, 0, 1] [0.0, 0.0, 0.0, 0.0]`

This is inconsistent. I expected either min or lanes_min (like in lanes_eq) to get the minimum for each lane.

Also, when you search `min(` in https://rust-lang.github.io/portable-simd/core_simd/simd/struct.Simd.html#method.min-1, where it says "Returns the minimum of each lane.", you have to scroll up quite a bit to see it is for floats only and the documentation of min and max for integers is very generic, doesn't explicitly mention that this isn't per lane.

jorgecarleitao:
This PR adds explicit implementations for `min` and `max` for integers.

Closes #247 and https://github.com/rust-lang/rust/issues/94682

### DIFF
--- a/crates/core_simd/src/comparisons.rs
+++ b/crates/core_simd/src/comparisons.rs
@@ -66,3 +66,37 @@ where
         unsafe { Mask::from_int_unchecked(intrinsics::simd_ge(self, other)) }
     }
 }
+
+macro_rules! impl_min_max_vector {
+    { $type:ty } => {
+        impl<const LANES: usize> Simd<$type, LANES>
+        where
+            LaneCount<LANES>: SupportedLaneCount,
+        {
+            /// Returns the lane-wise minimum with other
+            #[must_use = "method returns a new mask and does not mutate the original value"]
+            #[inline]
+            pub fn min(self, other: Self) -> Self {
+                self.lanes_gt(other).select(other, self)
+            }
+
+            /// Returns the lane-wise maximum with other
+            #[must_use = "method returns a new mask and does not mutate the original value"]
+            #[inline]
+            pub fn max(self, other: Self) -> Self {
+                self.lanes_lt(other).select(other, self)
+            }
+        }
+    }
+}
+
+impl_min_max_vector!(i8);
+impl_min_max_vector!(i16);
+impl_min_max_vector!(i32);
+impl_min_max_vector!(i64);
+impl_min_max_vector!(isize);
+impl_min_max_vector!(u8);
+impl_min_max_vector!(u16);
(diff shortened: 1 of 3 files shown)


crates/core_simd/src/comparisons.rs
```diff
@@ -66,3 +66,37 @@ where
...
 }
+
+macro_rules! impl_min_max_vector {
+    { $type:ty } => {
+        impl<const LANES: usize> Simd<$type, LANES>
+        where
+            LaneCount<LANES>: SupportedLaneCount,
+        {
+            /// Returns the lane-wise minimum with other
+            #[must_use = "method returns a new mask and does not mutate the original value"]
```
calebzulawski, line 77: This comment isn't quite right

calebzulawski (changes requested):

crates/core_simd/src/comparisons.rs
programmerjake, line 77: this must_use reason needs to be changed

calebzulawski:
Looks good to me other than that small comment

crates/core_simd/src/comparisons.rs
```diff
@@ -66,3 +66,37 @@ where
...
+            #[inline]
+            pub fn min(self, other: Self) -> Self {
+                self.lanes_gt(other).select(other, self)
+            }
+
+            /// Returns the lane-wise maximum with other
+            #[must_use = "method returns a new mask and does not mutate the original value"]
```
programmerjake, line 84: here too

jorgecarleitao:
Thank you both for the review. 🤦 Indeed. Pushed a fix to both of them.

programmerjake (approved):

calebzulawski (approved):

workingjubilee:
Thank you!

PR #265, opened 2022-03-13
Move comparisons to traits

calebzulawski:
A simpler variant of rust-lang/portable-simd#206.

* Comparisons are moved to `SimdPartialEq`, `SimdPartialOrd`, and `SimdOrd`.  The function names are prefixed with `simd_` to disambiguate from the regular `PartialEq` etc functions.  With the functions on traits instead of `Simd` directly, shadowing the function names doesn't work very well.
* Floating point `Ord`-like functions are put into a `SimdFloat` trait.  The intention is that eventually (some time after this PR) all floating point functions will be moved from `Simd` to `SimdFloat`, and the same goes for future `SimdInt`/`SimdUint` traits.


crates/core_simd/src/ord.rs
```diff
@@ -0,0 +1,222 @@
+use crate::simd::{intrinsics, LaneCount, Mask, Simd, SimdElement, SupportedLaneCount};
+
+/// Parallel `PartialOrd`.
+pub trait SimdPartialOrd {
```
programmerjake, line 4: `SimdPartialOrd` should probably imply `SimdPartialEq`.

crates/core_simd/src/ord.rs
```diff
@@ -0,0 +1,222 @@
...
+    /// The mask type returned by each comparison.
+    type Mask;
+
+    /// Test if each lane is less than the corresponding lane in `other`.
+    #[must_use = "method returns a new mask and does not mutate the original value"]
+    fn simd_lt(self, other: Self) -> Self::Mask;
```
programmerjake, line 10: I think we should name the new methods `lanes_*` rather than `simd_*` for consistency and leave it to the rename-to-`elementwise_*` PR to resolve that.

calebzulawski: Regarding "elementwise", I'm not sure functions themselves should necessarily have that prefix.  I was just thinking our terminology in the docs.  I figured `simd_` might be a good, terse way of indicating that you're not calling regular `Ord`.

crates/core_simd/src/ord.rs
calebzulawski: Good idea, this also lets me remove the duplicate `Mask`

workingjubilee:
Apologies, I am not trying to ignore this, I just was getting sidetracked by an Adventure or two.

crates/core_simd/src/ord.rs
programmerjake: i still stsnd by my consistency argument, if you want `simd_*` to be the prefix, then we should rename all `lanes_*` functions to `simd_*`

calebzulawski: These traits entail all of the functions with the `lanes_` prefix

programmerjake: (testing github bug with reviews, please ignore)

programmerjake: ah, ok.

programmerjake (approved):
lgtm

programmerjake:
we should post somewhere users will notice that this breaks API compatibility.

workingjubilee:
Okay. Having reviewed  this properly, I feel like the ergonomics of these ops could probably be improved further and I think we will eventually find we want to encode the idea of associated masking types in a separate trait, but having said that I see no reason not to merge this as it stands as it is a significant improvement.

PR #270, opened 2022-04-01
Add dotproduct assembly documentation and godbolt links

miguelraz:
Not yet finished but I wanted to save my work for a bit.

Adding a bunch of text to README.md, with some (may I say) nicely curated Rust godbolt links and displays.

`stdsimd` docs don't yet have a "voice/tone", let me know if it needs a course correction.


crates/core_simd/examples/README.md
```diff
@@ -0,0 +1,109 @@
...
+cargo run --example dot_product
+```
+
+and verify the code for `dot_product.rs` on your machine.
+
+### `dot_product.rs`
+
+This example code takes the dot product of two vectors. You are supposed to mulitply each pair of elements and add them all together.
+
+The easiest way to inspect the assembly of the `scalar` code versions (the non-SIMD versions) is to [click this link](https://rust.godbolt.org/z/xM9Mxb14n) for a *mise en place* of what is going on.
```
programmerjake, line 19: I think it would be better to avoid non-english phrases since not everyone knows French (i guess? idk what that phrase means).

crates/core_simd/examples/README.md
```diff
@@ -0,0 +1,109 @@
...
+    // TODO handle remainder when a.len() % 4 != 0
+    a.array_chunks::<4>()
+        .map(|&a| f32x4::from_array(a))
+        .zip(b.array_chunks::<4>().map(|&b| f32x4::from_array(b)))
+        .map(|(a, b)| (a * b).reduce_sum())
+        .sum()
+
+```
+
+1. SIMD comes in many flavors (instructions sets). These (like `sse`, `sse4.1`, `avx2`) describe the hardware capabilities of your current CPU. That is, if you don't have `avx512`, you physically do not have a SIMD vector that can hold 512 bytes at a time at most on your CPU.
```
programmerjake, line 57:
```suggestion
1. SIMD comes in many flavors (instructions sets). These (like `sse`, `sse4.1`, `avx2`) describe the hardware capabilities of your current CPU. That is, if you don't have `avx512`, you physically do not have any SIMD vector registers that can hold 512-bits at a time on your CPU. 
```

crates/core_simd/examples/README.md
```diff
@@ -0,0 +1,109 @@
...
+    a.array_chunks::<4>()
+        .map(|&a| f32x4::from_array(a))
+        .zip(b.array_chunks::<4>().map(|&b| f32x4::from_array(b)))
+        .map(|(a, b)| (a * b).reduce_sum())
+        .sum()
+
+```
+
+1. SIMD comes in many flavors (instructions sets). These (like `sse`, `sse4.1`, `avx2`) describe the hardware capabilities of your current CPU. That is, if you don't have `avx512`, you physically do not have a SIMD vector that can hold 512 bytes at a time at most on your CPU. 
+2. You can switch between different instruction sets by changing the `#![target-feature(...)]` macro above the function, as well as declaring it unsafe.
```
programmerjake, line 58:
```suggestion
2. You can switch between different instruction sets by both changing the `#![target-feature(...)]` macro above the function and declaring it unsafe.
```
declaring it unsafe by itself doesn't change the target features.

crates/core_simd/examples/README.md
```diff
@@ -0,0 +1,109 @@
...
+3. Inside Godbolt, you can hover over an instruction to display a tooltip of what it says. Try hovering your mouse over `mulps` and reading what it says.
```
programmerjake, line 59: I suggest phrasing in terms of "what the instruction does" rather than "what it says".

crates/core_simd/examples/README.md
```diff
@@ -0,0 +1,109 @@
...
+
+
+-----
+
+Now we will exploit the `mul_add` instruction. Open [this link to view the snippets side by side once again](https://rust.godbolt.org/z/vPTqG13vK). We've started off with a simple computation: adding and multiplying. Even though the arithmetic operations are not complicated, the performance payoff can come form knowing specific hardware capabilities like `mul_add`: in a single instruction, it can multiply 2 SIMD vectors and add them into a 3rd, which can cut swaths in the data movement overheads `xmm` registers can carry. Other instructions like inverse square roots are available (which are very popular for physics calculations), and it can get oodles more complex depending on the problem - there's published algorithms with `shuffles`, `swizzles` and `casts` for [decoding UTF8](https://arxiv.org/pdf/2010.03090.pdf), all in SIMD registers and with fancy table lookups. We won't talk about those here, but we just want to point out that firstly, reading the books can pay off drastically, and second, we're starting small to show the concepts, like using `mul_add` in the next snippet:
```
programmerjake, line 84: "can cut swaths in the data movement overheads `xmm` registers can carry" -- unclear, should be rephrased.

programmerjake (changes requested):

crates/core_simd/examples/README.md
```diff
@@ -0,0 +1,109 @@
...
+    // TODO handle remainder when a.len() % 4 != 0
+    a.array_chunks::<4>()
+        .map(|&a| f32x4::from_array(a))
+        .zip(b.array_chunks::<4>().map(|&b| f32x4::from_array(b)))
+        .fold(f32x4::splat(0.0), |acc, zipped| acc + zipped.0 * zipped.1)
+        .reduce_sum()
+}
+```
+
+In `dot_prod_simd_1`, we tried out the `fold` patter from our previous `scalar` code snippet examples. This pattern, when implemented via SIMD instructions naively, means that for every `f32x4` `element`-wise multiplication, we accumulate into a (initially `0` valued `f32x4` SIMD vector) and then finally do a `reduce_sum` at the end to get the final result. This
```
Urhengulas, line 79:
```suggestion
In `dot_prod_simd_1`, we tried out the `fold` pattern from our previous `scalar` code snippet examples. This pattern, when implemented via SIMD instructions naively, means that for every `f32x4` `element`-wise multiplication, we accumulate into a (initially `0` valued `f32x4` SIMD vector) and then finally do a `reduce_sum` at the end to get the final result. This
```

crates/core_simd/examples/README.md
Urhengulas, line 84: Probably this should be "can come from", not "can come form".
```suggestion
Now we will exploit the `mul_add` instruction. Open [this link to view the snippets side by side once again](https://rust.godbolt.org/z/vPTqG13vK). We've started off with a simple computation: adding and multiplying. Even though the arithmetic operations are not complicated, the performance payoff can come from knowing specific hardware capabilities like `mul_add`: in a single instruction, it can multiply 2 SIMD vectors and add them into a 3rd, which can cut swaths in the data movement overheads `xmm` registers can carry. Other instructions like inverse square roots are available (which are very popular for physics calculations), and it can get oodles more complex depending on the problem - there's published algorithms with `shuffles`, `swizzles` and `casts` for [decoding UTF8](https://arxiv.org/pdf/2010.03090.pdf), all in SIMD registers and with fancy table lookups. We won't talk about those here, but we just want to point out that firstly, reading the books can pay off drastically, and second, we're starting small to show the concepts, like using `mul_add` in the next snippet:
```

Urhengulas (commented):
I just spotted two potential typos while reading through your PR 😄

PR #272, opened 2022-04-03
Standardize documentation for SIMD vector and mask types

sstangl:
This is a documentation-only patch that standardizes the presentation of vector types like `u32x4` and mask types like `mask32x16`.

The reasoning behind the patch was as follows:
1. Standardized terminology should be preferred, so `element` instead of `value` and `SIMD vector` instead of `vector`. These terms appear in the rest of the documentation and tutorials.
2. Try to avoid situations where two numbers are next to each other. So `16 elements of 32 bits` instead of `16 32-bit elements`.
4. Try to anticipate what readers are looking for -- so state the full bit-width directly.

### Vector Types
- Before: Vector of 32 `i8` values
- After: A 256-bit SIMD vector with 32 elements of type `i8`.

### Mask Types
- Before: Vector of 16 16-bit masks
- After: A mask for SIMD vectors with 16 elements of 32 bits.

### DIFF
--- a/crates/core_simd/src/masks.rs
+++ b/crates/core_simd/src/masks.rs
@@ -507,58 +507,58 @@ where
     }
 }
 
-/// Vector of eight 8-bit masks
+/// A 64-bit SIMD vector mask for eight elements of 8 bits.
 pub type mask8x8 = Mask<i8, 8>;
 
-/// Vector of 16 8-bit masks
+/// A 128-bit SIMD vector mask for 16 elements of 8 bits.
 pub type mask8x16 = Mask<i8, 16>;
 
-/// Vector of 32 8-bit masks
+/// A 256-bit SIMD vector mask for 32 elements of 8 bits.
 pub type mask8x32 = Mask<i8, 32>;
 
-/// Vector of 16 8-bit masks
+/// A 512-bit SIMD vector mask for 64 elements of 8 bits.
 pub type mask8x64 = Mask<i8, 64>;
 
-/// Vector of four 16-bit masks
+/// A 64-bit SIMD vector mask for four elements of 16 bits.
 pub type mask16x4 = Mask<i16, 4>;
 
-/// Vector of eight 16-bit masks
+/// A 128-bit SIMD vector mask for eight elements of 16 bits.
 pub type mask16x8 = Mask<i16, 8>;
 
-/// Vector of 16 16-bit masks
+/// A 256-bit SIMD vector mask for 16 elements of 16 bits.
 pub type mask16x16 = Mask<i16, 16>;
 
-/// Vector of 32 16-bit masks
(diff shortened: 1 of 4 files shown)


calebzulawski (approved):
Seems like an improvement to me, thanks!

calebzulawski:
It's worth noting that the masks might not actually be the same size as their corresponding vector on some targets, but I'm not sure that the old comments were indicative of that either.

workingjubilee:
Correct, the Mask type is intended to denote a type that "matches with" a given vector element size and count. The exact repr that is used to store Masks is something the hardware decides.

crates/core_simd/src/masks.rs
```diff
@@ -507,58 +507,58 @@ where
     }
 }
 
-/// Vector of eight 8-bit masks
+/// A 64-bit SIMD vector mask for eight elements of 8 bits.
```
programmerjake, line 510: Masks have unspecified size (some architectures `mask16x8` is 128-bits, other architectures it is an 8-bit bitmask, we may also get architectures where it isn't either of those (e.g. masks could always have 8-bit elements so `mask8x8`, `mask16x8`, ..., and `mask64x8` would all be 64-bits)).

These comments need to not state a specific size for masks:
```suggestion
/// A SIMD vector mask of eight elements, for operating with SIMD vectors with 8 bit elements. 
```

programmerjake (changes requested):

crates/core_simd/src/masks.rs
sstangl: Ah yeah, that makes sense. Thanks everyone! I'll also update the `Mask` documentation to note that these are *boolean* masks -- I got tricked thinking of bitwise masking operations.

programmerjake:
> Correct, the Mask type is intended to denote a type that "matches with" a given vector element size and count. The exact repr that is used to store Masks is something the hardware decides... it's only "no less than 1 bit per element and no more than the size of the vector type it masks".

there isn't an upper limit on mask sizes, some architecture could say all masks have 64-bit elements, and it would still work correctly.

crates/core_simd/src/masks.rs
programmerjake: assuming you'll change the other comments too...

sstangl: Yeah, pondering the right wording. Don't worry :+1:

sstangl:
OK, updated with more accurate wording, thanks everyone. I had to be a *little* terse, because rustdoc's search feature has a fairly low maximum column count -- exceeding that makes all entries look roughly identical, with trailing ellipses.

I also added the line "Masks represent boolean inclusion/exclusion on a per-lane basis." to the `Mask` documentation.

programmerjake (approved):
lgtm

PR #274, opened 2022-04-06
Use SIMD equality for PartialEq on SIMD vectors

sstangl:
A small change implementing parallel `PartialEq` for `Simd` vectors, fixing a `TODO`. All `Simd` vectors by definition support `SimdPartialEq`, so the logic is copied.

I currently think that the call to `simd_reduce_all` should indeed be round-tripped through `Mask` to produce a proper integer vector, to handle odd architectures. I wasn't sure if `simd_reduce_all(simd_eq(x, y))` was valid, in particular because the definition in LLVM appears to be bitwise AND, and I wasn't clear on where the reduction to Rust `bool` takes place.

### DIFF
--- a/crates/core_simd/src/vector.rs
+++ b/crates/core_simd/src/vector.rs
@@ -429,8 +429,26 @@ where
 {
     #[inline]
     fn eq(&self, other: &Self) -> bool {
-        // TODO use SIMD equality
-        self.to_array() == other.to_array()
+        // Safety: All SIMD vectors are SimdPartialEq, and the comparison produces a valid mask.
+        let mask = unsafe {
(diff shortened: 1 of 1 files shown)


workingjubilee:
I believe currently `simd_eq` returns a vector of integers where the values are all -1 or 0. I suspect this is more correct code in any case, at least, relative to possible implementation variances in the future.

clarfonthey:
It looks like this is kind of the same #274? I personally would prefer the one that uses the safe methods, rather than the intrinsics directly, but they appear to do the same thing.

workingjubilee:
Do you mean #209? Yes, it effectively is.

Issue #228, opened 2022-01-17
Use a more trait-oriented design


workingjubilee:
We have coalesced consensus around a more direct application of using traits for the type-argument-specific functions on `std::simd::Simd` and finding other ways to resolve the too-many-imports problem, at least for now. While it **is** inconvenient to have to import such traits, "making imports more manageable" is something that should probably be addressed at a slightly higher design level than within this library's types (since it's of relevance to all of Rust).

( Obviously, if people really want these traits in scope always, they can nominate them for the prelude as-is, though a solution that is less dependent on "being std", like inherent traits, would be nice. )

Lokathor:
`use core::simd::prelude::*;` doesn't seem particular burdensome until then.

workingjubilee:
@Urgau [asked](https://github.com/rust-lang/portable-simd/issues/230#issue-1110455656):
> Hi, I was wondering why this crate choose to go with a **trait-oriented design** and all the complication with it **instead** of going with the **primitive/struct and macros route** like core is doing for `uX`, `iX` and `fX` ?

### Short Answer
We tried that and it proved fairly unwieldy and subpar.

### Long Answer
First and foremost: This is all still an experiment on nightly, so we remain entitled to rewrite everything. We will not exercise this entitlement frivolously, but some amount of tinkering is not frivolous (and, for instance, code using the type aliases has often remained stable in spite of massive internal changes).

But it turns out the way that the core primitive types work
- is not actually ideal for the needs of numeric code.
- does not "scale" to the nature of `core::simd` very well.

Please see https://github.com/rust-lang/rfcs/pull/2581, which proposes generic integers as a way to deal with some of the existing problems. In essence, it is not actually clear that having something like `num-traits` in `std` is a Bad Idea, and T-libs has expressed at least some positive sentiment towards it recently.

But we still must provide methods that are, at minimum:
- generic over all lane counts.
- generic over all numeric types.

This means we can either use `struct Simd<T, N>` as the "base container" or use multiple structs (as we experimented with in the past) and then provide a `trait Vector` which is implemented for all types.

The principle "entities shall not be multiplied beyond necessity" (also known as Occam's razor, and often misquoted as something like "the simplest solution is often the best") undergirds much of logic, math, and even user experience design. This is also referred to as "Don't Repeat Yourself" in programming. The problem is that using macros still constitutes a certain amount of "repeating yourself", including in the form of documentation. And that is, at least, my major concern:

[...]

Issue #15, opened 2020-09-30
What traits abstract over what?


workingjubilee:
We've discussed potentially implementing trait abstractions over SIMD vectors in the basic API because it's pretty unwieldy to handle every single possible TxW combo, and we've mentioned we will need to leave some traits in separate semi-official crates, e.g. a `simd-complex`, but we need to think about what abstractions we want to offer where.

Discussion partly here: https://rust-lang.zulipchat.com/#narrow/stream/257879-project-portable-simd/topic/Semi-official.20(non-std.3A.3Asimd).20crates

workingjubilee:
of particular note, what seems like a strong consensus:

> I'd be fine with just recording an official stance that "SIMD types should, by default, implement ops traits vertically", and then leaving more traits to be added to core::num later on. Perhaps a "we also propose the following traits but they can be accepted separately from the main SIMD work"
—@Lokathor

workingjubilee:
Somewhat obvious if one reviews the code so far, but so far we've been inclined towards a strong absence of traits in the public API, with the occasional attempt to use them cleverly for bounds. Basically there isn't a lot of reason in `std` to trait over u32 and u8 or u16 and i16, so we've been mostly also not. This may change, we'll see!

bjorn3:
Simd code would more likely to want to abstract over different lane sizes for optimal speed on different processors than scalar code would want to abstract over scalar size.

calebzulawski:
To add on to Jubilee's justification, I think it's not so much that there is no desire to abstract over SIMD vectors, but there isn't much need to do so *in core*.  I think this is similar to why `num` is a separate crate.  I envision a crate similar to `num` (or even part of it) for SIMD.

PR #278, opened 2022-04-15
Move element-specific functions to traits

### LINKED ISSUE #85: Improve rustdoc views
The implementation is somewhat hard to read when viewed with rustdoc, to a degree that isn't really acceptable for external users. Some of this can be improved just by rearranging things.

calebzulawski:
This serves a few purposes:
* Deduplicates functions that are reimplemented for each element type (the docs are dramatically better)
* Distinguishes between functions on particular element types and functions that work on any element type (also helps docs)
* Distinguishes between functions that are slightly different between different elements (like `abs`)

I believe this should close #85 and #228 (and maybe #15 as well)


crates/core_simd/src/elements/float.rs
```diff
@@ -0,0 +1,344 @@
...
+    ///
+    /// * `1.0` if the number is positive, `+0.0`, or `INFINITY`
+    /// * `-1.0` if the number is negative, `-0.0`, or `NEG_INFINITY`
+    /// * `NAN` if the number is `NAN`
+    #[must_use = "method returns a new vector and does not mutate the original value"]
+    fn signum(self) -> Self;
+
+    /// Returns each lane with the magnitude of `self` and the sign of `sign`.
+    ///
+    /// If any lane is a `NAN`, then a `NAN` with the sign of `sign` is returned.
```
programmerjake, line 86: hmm, wouldn't this be element-wise `copysign` rather than returning a vector of all NaNs if at least one input vector is a NaN?

programmerjake (approved):
Other than 1 nit, lgtm!

crates/core_simd/src/elements/float.rs
calebzulawski: This is just the existing documentation, but I updated it to be a little clearer.

programmerjake (approved):

calebzulawski:
Just upgraded the bound from `Sized` to `Copy` since some of the functions take `self` rather than `&self` (and clippy helpfully gave a warning)

programmerjake (approved):

calebzulawski:
@workingjubilee do you mind if I merge this?

PR #280, opened 2022-05-21
Document unsafety

### LINKED ISSUE #173: Document internal unsafety

calebzulawski:
Fixes #173.  Clippy no longer crashes on this 🙂

I also made the bitmask traits safe, since implementing them isn't really unsafe.  The bitmask conversion functions are themselves safe but contain documented unsafe blocks.

### DIFF
--- a/crates/core_simd/src/elements/float.rs
+++ b/crates/core_simd/src/elements/float.rs
@@ -202,17 +202,20 @@ macro_rules! impl_trait {
             #[inline]
             fn to_bits(self) -> Simd<$bits_ty, LANES> {
                 assert_eq!(core::mem::size_of::<Self>(), core::mem::size_of::<Self::Bits>());
+                // Safety: transmuting between vector types is safe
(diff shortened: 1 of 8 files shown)


programmerjake (approved):
lgtm

PR #281, opened 2022-06-02
Deduplicate to_int_unchecked

calebzulawski:
No need to implement on `f32` and `f64` specifically, `FloatToInt` handles this.


calebzulawski:
(I think this was the last function implemented on specific element types!)

programmerjake (approved):
lgtm

PR #282, opened 2022-06-06
Add splat_lane

calebzulawski:
Like splat, but from an existing vector.

### DIFF
--- a/crates/core_simd/src/swizzle.rs
+++ b/crates/core_simd/src/swizzle.rs
@@ -188,6 +188,21 @@ where
     T: SimdElement,
     LaneCount<LANES>: SupportedLaneCount,
 {
+    /// Splat a lane to the other lanes in the vector.
+    ///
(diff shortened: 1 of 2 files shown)


thomcc (commented):
I'm not sure this makes that much sense as a public method (although I'm not sure it's *not* worth having, it just seems a little specific of a shuffle), but I do think it's worth using inside `splat` if it's needed to fix https://rust-lang.zulipchat.com/#narrow/stream/257879-project-portable-simd/topic/Very.20bad.20.60u16x8.3A.3Asplat.60.20codegen.20on.20x86_64

programmerjake (commented):
idk that we should have `splat_lane` as a public API, considering how easy it is to do:
```rust
pub fn splat_lane_0(v: u16x8) -> u16x8
{
    simd_swizzle!(v, [0; 8])
}
```

other than that, lgtm.

calebzulawski:
I suppose it is pretty trivial with the macro, agreed.  Trivial enough to add in the future if requested

PR #287, opened 2022-06-23
Vectors of pointers

calebzulawski:
* Adds vectors of pointers
* Changes scatter/gather implementation (but doesn't introduce new pointer API, yet)
* Modifies `Simd::cast`, since all casts are no longer valid (e.g. casting `f32` to a pointer)


crates/core_simd/src/cast.rs
```diff
@@ -0,0 +1,129 @@
...
+macro_rules! into_pointer {
+    { $($type:ty),* } => {
+        $(
+        impl<T> SimdCast<$type> for *const T {
+            fn cast<const LANES: usize>(x: Simd<Self, LANES>) -> Simd<$type, LANES>
+            where
+                LaneCount<LANES>: SupportedLaneCount,
+            {
+                // Safety: transmuting isize to pointers is safe
+                let x: Simd<isize, LANES> = unsafe { core::mem::transmute_copy(&x) };
```
programmerjake, line 54: I would consider that questionably safe, pointer to integer transmutes are almost certainly UB:
https://github.com/rust-lang/unsafe-code-guidelines/issues/286

programmerjake: also, the safety comment is the wrong way around, you're transmuting pointers to integers here, which is probably UB.

crates/core_simd/src/elements/const_ptr.rs
```diff
@@ -0,0 +1,66 @@
...
+    #[inline]
+    fn addr(self) -> Self::Usize {
+        self.cast()
+    }
+
+    #[inline]
+    fn wrapping_add(self, count: Self::Usize) -> Self {
+        let addr = self.addr() + (count * Simd::splat(core::mem::size_of::<T>()));
+        // Safety: transmuting usize to pointers is safe, even if accessing those pointers isn't.
+        unsafe { core::mem::transmute_copy(&addr) }
```
programmerjake, line 64: this really should use either a compiler intrinsic for vector wrapping add, or use `Simd::cast`, once casting is fixed.

programmerjake: also, if `addr` is supposed to be like `<*const T>::addr`, that specifically returns an integer that can't be converted back to a pointer. so that side also needs to be `cast`

crates/core_simd/src/cast.rs
thomcc: It's no longer almost certainly UB, but it does lose provenance (and `as`/casting should not): https://doc.rust-lang.org/nightly/src/core/ptr/const_ptr.rs.html#183-185

crates/core_simd/src/elements/const_ptr.rs
```diff
@@ -0,0 +1,66 @@
...
+    /// Returns `true` for each lane that is null.
+    fn is_null(self) -> Self::Mask;
+
+    /// Changes constness without changing the type.
+    fn as_mut(self) -> Self::MutPtr;
+
+    /// Gets the "address" portion of the pointer.
+    ///
+    /// Equivalent to calling [`pointer::addr`] on each lane.
+    fn addr(self) -> Self::Usize;
```
programmerjake, line 24: the docs need some elaboration -- especially specifically mentioning that the result can't be cast back to a pointer and used.
https://github.com/rust-lang/rust/blob/10f4ce324baf7cfb7ce2b2096662b82b79204944/library/core/src/ptr/const_ptr.rs#L178

crates/core_simd/src/cast.rs
programmerjake: i meant its UB to use the result by converting it back to a pointer and expecting it to be valid.

thomcc: I see. That is still UB (for a while, the transmute itself was probably UB too).

calebzulawski: My understanding is that once you convert a pointer to an integer, it loses provenance no matter what, and you need to use `with_addr` to inherit an existing pointer's provenance.  So this particular line (converting a pointer to an integer) should be fine, I think.

crates/core_simd/src/elements/const_ptr.rs
calebzulawski: This is just a rewrite of what we already had.  Definitely needs to use an intrinsic to be correct.

crates/core_simd/src/elements/const_ptr.rs
calebzulawski: Thought it would be better to reference the underlying function it's equivalent to, especially considering it's likely to change sometime before stabilization

crates/core_simd/src/cast.rs
programmerjake: that's actually incorrect -- on most platforms using `as` to cast a pointer to an integer exposes the pointer such that `as` casts from integers back to pointers using that address produces the original pointer including provenance. transmute doesn't do that.

because these functions are named `cast` and intended to function like `as`, they need to have the same pointer exposing behavior as `as`. you don't get that with `transmute`

crates/core_simd/src/elements/const_ptr.rs
programmerjake: yes, but imho we need to point out that you can't use it like `ptr as usize` which allows getting `ptr` back by `as` casting the resulting integer back to a pointer. otherwise people who only read the docs of this function may assume it's `ptr as usize`, leading them to write code with UB.

crates/core_simd/tests/pointers.rs
```diff
@@ -0,0 +1,59 @@
...
+                test_helpers::test_unary_elementwise(
+                    &Simd::<*$constness (), LANES>::addr,
+                    &<*$constness ()>::addr,
+                    &|_| true,
+                );
+            }
+
+            fn wrapping_offset<const LANES: usize>() {
+                test_helpers::test_binary_elementwise(
+                    &Simd::<*$constness (), LANES>::wrapping_offset,
```
programmerjake, line 26: testing `wrapping_offset` on a `*mut ()` is kinda pointless since it always returns the input pointer, since `()` is a ZST. imho it would be better to test on `*mut u32` since it has both a nonzero pointee size (avoiding the output always being the same as the input), and a non-one pointee size (catching the potential bug where rustc forgot to multiply the offset by the pointee size)

crates/core_simd/src/cast.rs
calebzulawski: I realized my confusing when reviewing the docs for `addr`:
> This is equivalent to `self as usize`, which semantically discards provenance and address-space information. To properly restore that information, use [with_addr](https://doc.rust-lang.org/std/primitive.pointer.html#method.with_addr) or [map_addr](https://doc.rust-lang.org/std/primitive.pointer.html#method.map_addr).

The bit about `self as usize` is incorrect, then?

programmerjake: > I realized my confusing when reviewing the docs for `addr`:
> 
> > This is equivalent to `self as usize`, which semantically discards provenance and address-space information. To properly restore that information, use [with_addr](https://doc.rust-lang.org/std/primitive.pointer.html#method.with_addr) or [map_addr](https://doc.rust-lang.org/std/primitive.pointer.html#method.map_addr).
> 
> The bit about `self as usize` is incorrect, then?

yes, when referring to current `ptr as usize` semantics...that said, iirc they want to move to `as` being equivalent to `addr` rather than `expose_addr` at some point (requires edition boundary).

crates/core_simd/src/elements/const_ptr.rs
```diff
@@ -0,0 +1,118 @@
...
+        unimplemented!()
+        //self.cast()
+    }
+
+    #[inline]
+    fn addr(self) -> Self::Usize {
+        // Safety: Since `addr` discards provenance, this is safe.
+        unsafe { core::mem::transmute_copy(&self) }
+
+        //TODO switch to casts when available
```
programmerjake, line 89: actually, `<*const T>::addr` uses `transmute` right now, it'll eventually be its own intrinsic, not a cast.

calebzulawski: Interesting, on stable it's `as`, but `transmute` now on nightly.

programmerjake: iirc that's because `transmute` was recently changed so ptr -> int isn't UB.

programmerjake (commented):
lgtm, assuming this is changed once simd int <-> ptr casts land in rustc

calebzulawski:
New intrinsics merged, so this is ready to go.

workingjubilee:
Huzzah.

crates/core_simd/src/elements/const_ptr.rs
```diff
@@ -0,0 +1,139 @@
...
+    fn wrapping_offset(self, offset: Self::Isize) -> Self;
+
+    /// Calculates the offset from a pointer using wrapping arithmetic.
+    ///
+    /// Equivalent to calling [`pointer::wrapping_add`] on each lane.
+    fn wrapping_add(self, count: Self::Usize) -> Self;
+
+    /// Calculates the offset from a pointer using wrapping arithmetic.
+    ///
+    /// Equivalent to calling [`pointer::wrapping_add`] on each lane.
```
programmerjake, line 64:
```suggestion
    /// Equivalent to calling [`pointer::wrapping_sub`] on each lane.
```

crates/core_simd/src/elements/mut_ptr.rs
```diff
@@ -0,0 +1,134 @@
...
+    fn wrapping_offset(self, offset: Self::Isize) -> Self;
+
+    /// Calculates the offset from a pointer using wrapping arithmetic.
+    ///
+    /// Equivalent to calling [`pointer::wrapping_add`] on each lane.
+    fn wrapping_add(self, count: Self::Usize) -> Self;
+
+    /// Calculates the offset from a pointer using wrapping arithmetic.
+    ///
+    /// Equivalent to calling [`pointer::wrapping_add`] on each lane.
```
programmerjake, line 61:
```suggestion
    /// Equivalent to calling [`pointer::wrapping_sub`] on each lane.
```

programmerjake (approved):
once those two `wrapping_sub` doc comments are fixed, lgtm

programmerjake (approved):

workingjubilee:
Apologies, going to try to get to this this week.

calebzulawski:
Okay thanks!

crates/core_simd/src/cast.rs
```diff
@@ -0,0 +1,45 @@
...
+unsafe impl SimdCast for f32 {}
+// Safety: primitive number types can be cast to other primitive number types
+unsafe impl SimdCast for f64 {}
+
+/// Supporting trait for `Simd::cast_ptr`.  Typically doesn't need to be used directly.
+///
+/// # Safety
+/// Implementing this trait asserts that the type is a valid vector element for the `simd_cast_ptr`
+/// intrinsic.
+pub unsafe trait SimdCastPtr: SimdElement {}
```
workingjubilee, line 40:
```suggestion
/// Supporting trait for `Simd::cast_ptr`.  Typically doesn't need to be used directly.
///
/// # Safety
/// Implementing this trait asserts that the type is a valid vector element for the `simd_cast_ptr`
/// intrinsic.
pub unsafe trait SimdCastPtr: SimdElement {}
```

Is there a reason we can't just use the bounds that std uses to guarantee the Pointee has the "Metadata" of `()` (and is thus a thin pointer)?

crates/core_simd/src/cast.rs
```diff
@@ -0,0 +1,45 @@
...
+
+// Safety: pointers can be cast to other pointer types
+unsafe impl<T> SimdCastPtr for *const T {}
+// Safety: pointers can be cast to other pointer types
+unsafe impl<T> SimdCastPtr for *mut T {}
```
workingjubilee, line 45: This bound doesn't seem correct since I think this allows fat pointers.

crates/core_simd/src/eq.rs
```diff
@@ -71,3 +73,37 @@ macro_rules! impl_mask {
 }
 
 impl_mask! { i8, i16, i32, i64, isize }
+
+impl<T, const LANES: usize> SimdPartialEq for Simd<*const T, LANES>
```
workingjubilee, line 77: This seems okay. Are there any caveats, e.g. if the pointers get evaluated at const time, that we should document?

crates/core_simd/src/ord.rs
```diff
@@ -211,3 +213,101 @@ macro_rules! impl_mask {
...
+    }
+
+    #[inline]
+    fn simd_clamp(self, min: Self, max: Self) -> Self {
+        assert!(
+            min.simd_le(max).all(),
+            "each lane in `min` must be less than or equal to the corresponding lane in `max`",
+        );
+        self.simd_max(min).simd_min(max)
+    }
```
workingjubilee, line 312: The notion of clamping pointers seems somewhat dubious...?

crates/core_simd/src/vector.rs
```diff
@@ -242,11 +254,10 @@ where
     #[cfg_attr(miri, track_caller)] // even without panics, this helps for Miri backtraces
     pub unsafe fn to_int_unchecked<I>(self) -> Simd<I, LANES>
     where
-        T: core::convert::FloatToInt<I>,
-        I: SimdElement,
+        T: core::convert::FloatToInt<I> + SimdCast,
+        I: SimdCast,
```
workingjubilee, line 258: Huh, so this happens because `SimdElement` is now either `SimdCast` xor `SimdPtrCast`, right?

crates/test_helpers/src/lib.rs
```diff
@@ -135,21 +157,21 @@ pub fn test_unary_elementwise<Scalar, ScalarResult, Vector, VectorResult, const
...
-            }
-            result
-        };
+        let result_2: [ScalarResult; LANES] = x
+            .iter()
+            .copied()
+            .map(fs)
+            .collect::<Vec<_>>()
+            .try_into()
+            .unwrap();
```
workingjubilee, line 174: This does look a bit nicer.

crates/test_helpers/src/lib.rs
```diff
@@ -196,23 +218,24 @@ pub fn test_binary_elementwise<
...
-            }
-            result
-        };
+        let result_2: [ScalarResult; LANES] = x
+            .iter()
+            .copied()
+            .zip(y.iter().copied())
+            .map(|(x, y)| fs(x, y))
+            .collect::<Vec<_>>()
+            .try_into()
```
workingjubilee, line 237: Hm, maybe it would be nice if we could just `try_collect` here instead...? Is that hypothetically possible?

crates/test_helpers/src/lib.rs
```diff
@@ -196,23 +218,24 @@ pub fn test_binary_elementwise<
     fs: &dyn Fn(Scalar1, Scalar2) -> ScalarResult,
     check: &dyn Fn([Scalar1; LANES], [Scalar2; LANES]) -> bool,
 ) where
-    Scalar1: Copy + Default + core::fmt::Debug + DefaultStrategy,
-    Scalar2: Copy + Default + core::fmt::Debug + DefaultStrategy,
-    ScalarResult: Copy + Default + biteq::BitEq + core::fmt::Debug + DefaultStrategy,
```
workingjubilee, line 201: huzzah less bounds.

workingjubilee (commented):
Thank you for this! Sorry for being so distracted. Mostly things look good but I have a few questions?

crates/core_simd/src/vector.rs
calebzulawski: Yeah.  But really, it has nothing to do with `SimdPtrCast` (if rust ever added a numeric pointer type for some reason, it could implement both).  This could be thought of as `SimdNumeric`, but specific to casting.

crates/core_simd/src/cast.rs
calebzulawski: I did change `SimdElement` to only implement on thin pointers, for the time being.

crates/core_simd/src/cast.rs
calebzulawski: I changed it to require `T: Pointee<Metadata = ()>`

crates/core_simd/src/vector.rs
workingjubilee: Ahhh okay.

crates/core_simd/src/cast.rs
workingjubilee: I think that combined, those two changes are correct, and I agree that I don't see how the metadata casting would be valid otherwise if the metadata did not match, except for the special case of a cast that strips metadata (i.e. takes any `U` to `()`) (at which point we don't need that to be a "SIMD cast", really) (it's fine to not model every possible action, especially if the action is dubious).

crates/core_simd/src/ord.rs
calebzulawski: It is odd, but pointers already implement `Ord` 🤷‍♂️

crates/core_simd/src/eq.rs
calebzulawski: I'm not sure it has any caveats beyond what's already true for `const fn` or whatever.  Did you have anything specific in mind?

crates/core_simd/src/vector.rs
calebzulawski: I think I will change it to `T: SimdCast<U>`, to match the updated `SimdPtrCast` (and in case we want to use similar bounds in the future).

calebzulawski: Actually, I realized this would require a quadratic number of implementations... It's probably fine how it is.

crates/test_helpers/src/lib.rs
calebzulawski: I think so, I didn't know about `try_collect` until now.  I'm not sure it's worth bringing in an unstable feature just for that though.

workingjubilee: Yeah, that's fair. I am mostly curious if it's possible for users of `try_collect` to experiment with.

calebzulawski: I just realized, I think this is the opposite of `try_collect`.  The iterator items can't error, the conversion of the collection is fallable (`Vec` to array conversion)

workingjubilee: Oh okay! Whoops.

crates/core_simd/src/eq.rs
workingjubilee: I am thinking of the pointer indeterminacy problem during CTFE. Things that would share addresses during runtime may not during compile-time function execution, and vice versa.

calebzulawski: Honestly, not sure what the implications would be other than since the SIMD intrinsics aren't const (yet) so we know any pointers for now will be runtime only.  If you do something bad with CTFE before putting it in a vector, that's probably out of scope for now.

calebzulawski: I think it's fair to assume all vector operations should mirror scalar operations in that regard, whatever the implications are.

workingjubilee: Fair enough. I think it's mostly something we should keep in mind while exploring the API surface.

crates/core_simd/src/intrinsics.rs
```diff
@@ -151,4 +148,17 @@ extern "platform-intrinsic" {
...
+    pub(crate) fn simd_arith_offset<T, U>(ptr: T, offset: U) -> T;
+
+    /// equivalent to `T as U` semantics, specifically for pointers
+    pub(crate) fn simd_cast_ptr<T, U>(ptr: T) -> U;
+
+    /// expose a pointer as an address
+    pub(crate) fn simd_expose_addr<T, U>(ptr: T) -> U;
+
+    /// convert an exposed address back to a pointer
+    pub(crate) fn simd_from_exposed_addr<T, U>(addr: T) -> U;
```
RalfJung, line 163: It look like currently no tests actually call any of these intrinsics, or am I missing something?

That would explain why portable-simd tests still pass in Miri where these intrinsics are not implemented.

calebzulawski: Ah, looking at the tests, I did omit some of the very simple ones like this.  They shouldn't be hard to add.

PR #289, opened 2022-07-01
Strengthen warnings about relying on Mask layout

thomcc:
This makes it more clear that you can't rely on the layout of these, which seems worth doing given that the names vaguely suggest that you can (and the docs only clarify that you can't on `Mask` but not the `maskNxM` aliases).

Not tied to the specific wording here, and I'm not sure that it's actually worth clarifying that "yes, this includes `[iN; M]`", so I'm open to rewording. I do think we should have the caveat on the type aliases as well, though.

### DIFF
--- a/crates/core_simd/src/masks.rs
+++ b/crates/core_simd/src/masks.rs
@@ -83,7 +83,9 @@ impl_element! { isize }
 ///
 /// Masks represent boolean inclusion/exclusion on a per-lane basis.
 ///
-/// The layout of this type is unspecified.
+/// The layout of this type is unspecified, and may change between platforms
+/// and/or Rust versions, and code should not assume that it is equivalent to
+/// `[T; LANES]`.
 #[repr(transparent)]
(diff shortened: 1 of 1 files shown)


crates/core_simd/src/masks.rs
```diff
@@ -83,8 +83,9 @@ impl_element! { isize }
 ///
 /// Masks represent boolean inclusion/exclusion on a per-lane basis.
 ///
-/// The layout of this type is unspecified.
-#[repr(transparent)]
```
calebzulawski, line 87: I don't think you meant to delete this?

calebzulawski (changes requested):

crates/core_simd/src/masks.rs
thomcc: Uhhhh nice catch.

calebzulawski (approved):

workingjubilee:
Thank you for following up on this!

PR #299, opened 2022-07-29
Fix interleave/deinterleave for vectors with only one lane

### LINKED ISSUE #298: Using Simd interleave leads to wrong result for Lanes = 1
The following code returns a wrong result
```rust
#![feature(portable_simd)]

#[cfg(test)]
mod tests {
    use std::simd::Simd;
    
    #[test]
    fn test_interleave() {
        let a = Simd::<u8, 1>::from_array([1_u8; 1]);
        let b = Simd::<u8, 1>::from_array([2_u8; 1]);
        
        let (c, d) = a.interleave(b);
        
        assert_eq!(c.to_array(), [1_u8; 1]); // this works
        assert_eq!(d.to_array(), [2_u8; 1]); // this is wrong, because d is [1_u8; 1]
    }
}
```
I suspect that this is because `1 / 2 = 0`.
This is the playground link: https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=bffbd7488ab03295b9c7724a32a8511c

calebzulawski:
Fixes #298

### DIFF
--- a/crates/core_simd/src/swizzle.rs
+++ b/crates/core_simd/src/swizzle.rs
@@ -340,7 +340,7 @@ where
             let mut i = 0;
             while i < LANES {
                 // Treat the source as a concatenated vector
-                let src_index = i * 2 + if second { 1 } else { 0 };
+                let src_index = i * 2 + second as usize;
                 idx[i] = if src_index < LANES {
                     Which::First(src_index)
                 } else {
--- a/crates/core_simd/src/swizzle.rs
+++ b/crates/core_simd/src/swizzle.rs
@@ -265,13 +265,10 @@ where
 
     /// Interleave two vectors.
     ///
-    /// Produces two vectors with lanes taken alternately from `self` and `other`.
+    /// The resulting vectors contain lanes taken alternatively from `self` and `other`, first
+    /// filling the first result, and then the second.
     ///
-    /// The first result contains the first `LANES / 2` lanes from `self` and `other`,
-    /// alternating, starting with the first lane of `self`.
-    ///
-    /// The second result contains the last `LANES / 2` lanes from `self` and `other`,
-    /// alternating, starting with the lane `LANES / 2` from the start of `self`.
+    /// The reverse of this operation is [`Simd::deinterleave`].
     ///
     /// ```
     /// #![feature(portable_simd)]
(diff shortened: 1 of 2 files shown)


programmerjake (commented):
actually, the original code is broken for lane counts that aren't a multiple of 2, i'd prefer if we fix all odd lane counts (since we'll want to relax the power-of-2 restriction later), rather than just introduce a special case for 1 lane.

calebzulawski:
Good point.  Though, I'm now realizing that interleaving odd-length vectors on instruction sets that only have power-of-two vectors doesn't work out quite right.  Interleaving `[0, 1, 2, x]` and `[3, 4, 5, x]` (vectors of 3 stuffed into hardware vectors of 4) wants to make `[0, 3, 1, 4]` and `[2, 5, x, x]`.  It's not really clear to me what non-multiples-of-two should look like.  `[0, 3, 1]` and `[4, 2, 5]`?  I guess that's interleaved, but I kinda doubt that matches an efficient shuffle on any architecture.

programmerjake:
> Good point. Though, I'm now realizing that interleaving odd-length vectors on instruction sets that only have power-of-two vectors doesn't work out quite right. Interleaving `[0, 1, 2, x]` and `[3, 4, 5, x]` (vectors of 3 stuffed into hardware vectors of 4) wants to make `[0, 3, 1, 4]` and `[2, 5, x, x]`.

imho it doesn't matter that much how it's represented when extended by llvm to a larger vector, you have the exact same style of issue for 4-lane vectors when the minimum hardware vector length is 8 or 16 for that type. What we care about is the user-visible API.

> It's not really clear to me what odd length interleave should look like. `[0, 3, 1]` and `[4, 2, 5]`?

That's what I'd go for...

> I guess that's interleaved, but I kinda doubt that matches an efficient shuffle on any architecture.

the bottom half matches a power-of-2 interleave.
the top half matches a power-of-2 interleave, followed by a slide-up and insert missing lanes...or just a general swizzle.

imho we need it for consistency, not because it's necessarily more efficient.

programmerjake:
see also: https://rust-lang.zulipchat.com/#narrow/stream/257879-project-portable-simd/topic/experimental.20feature.20to.20allow.20non-power-of-2.20vector.20lengths

calebzulawski:
Pretty sure this is fixed for other odd lengths, not sure without merging the other PR and writing new tests.

programmerjake (approved):
lgtm. we'd probably want to add other tests for other odd lane counts, that can be done as a followup PR after #300 is merged.

PR #310, opened 2022-10-16
Update readme

calebzulawski:
Updated the readme to better reflect the current status of `std::simd`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,20 +24,10 @@ or by setting up `rustup default nightly` or else with `cargo +nightly {build,te
 ```bash
 cargo new hellosimd
 ```
-to create a new crate. Edit `hellosimd/Cargo.toml` to be 
-```toml
-[package]
-name = "hellosimd"
-version = "0.1.0"
-edition = "2018"
-[dependencies]
-core_simd = { git = "https://github.com/rust-lang/portable-simd" }
-```
-
-and finally write this in `src/main.rs`:
+to create a new crate. Finally write this in `src/main.rs`:
 ```rust
 #![feature(portable_simd)]
-use core_simd::*;
+use std::simd::f32x4;
 fn main() {
     let a = f32x4::splat(10.0);
     let b = f32x4::from_array([1.0, 2.0, 3.0, 4.0]);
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -24,45 +24,33 @@ or by setting up `rustup default nightly` or else with `cargo +nightly {build,te
...
-* **Signed Integers:** `i8`, `i16`, `i32`, `i64`, `i128`, `isize`
-* **Unsigned Integers:** `u8`, `u16`, `u32`, `u64`, `u128`, `usize`
-* **Masks:** `mask8`, `mask16`, `mask32`, `mask64`, `mask128`, `masksize`
+* **Signed Integers:** `i8`, `i16`, `i32`, `i64`, `isize` (`i128` excluded)
+* **Unsigned Integers:** `u8`, `u16`, `u32`, `u64`, `usize` (`u128` excluded)
+* **Masks:** 8-bit, 16-bit, 32-bit, 64-bit, and `usize`-sized masks
 
 Floating point, signed integers, and unsigned integers are the [primitive types](https://doc.rust-lang.org/core/primitive/index.html) you're already used to.
-The `mask` types are "truthy" values, but they use the number of bits in their name instead of just 1 bit like a normal `bool` uses.
+The mask types are "truthy" values, like `bool`, but have an unspecified layout in the vector type and cannot be constructed outside of a vector.
```
programmerjake, line 53:
```suggestion
The mask types have elements that are "truthy" values, like `bool`, but have an unspecified layout because different architectures prefer different layouts for mask types.
```
"unspecified layout in vector type" doesn't make sense to me. also `Mask` is constructable just like `Vec` is: you just call the constructor functions.

programmerjake:
> Sorry if this has been addressed, but does the unspecified layout affect the veracity of the docs for to_int():
> 
> https://rust-lang.github.io/portable-simd/core_simd/struct.Mask.html#method.to_int

no, `m.to_int()` is entirely equivalent to `m.select(Simd::splat(-1), Simd::splat(0))`, layout doesn't affect that -- just like the `Vec` struct's layout being unspecified doesn't mean you can't use `Vec<bool>`

sammysheep:
> > Sorry if this has been addressed, but does the unspecified layout affect the veracity of the docs for to_int():
> > https://rust-lang.github.io/portable-simd/core_simd/struct.Mask.html#method.to_int
> 
> no, `m.to_int()` is entirely equivalent to `m.select(Simd::splat(-1), Simd::splat(0))`, layout doesn't affect that -- just like the `Vec` struct's layout being unspecified doesn't mean you can't use `Vec<bool>`

Phew, thanks for the clarification and all that's been accomplished too!

README.md
calebzulawski: Yeah this is better. I meant the individual element is not individually constructible, but that's probably pedantic anyway.  It's effectively `bool`.

programmerjake (approved):

PR #311, opened 2022-10-17
Specify aliases in one place, and make it more uniform which are defined

### LINKED ISSUE #301: missing mask8x4
Is there a reason there's no `mask8x4` exposed? `i8x4` is and the API supports it.

I'm also curious why there's no `i/u8x2`

calebzulawski:
Fixes #301.

This expands the type aliases to cover all sizes from 1 to 64 elements, to keep it easy to audit.


programmerjake (approved):

sammysheep:
Knit: based on what is being aliased
https://rust-lang.github.io/portable-simd/core_simd/struct.Simd.html

Change:
`            pub type $alias = $crate::simd::Simd<$element, $elements>;`
To:
`            pub type $alias = $crate::simd::Simd<$simd_element, $lanes>;`

I must admit that what you've done matches the current alias description better, but reading `$element` next to `$elements` confused this reader at the first.

May just be my issue. Thanks again!

calebzulawski:
Good point, made the names something a little easier to distinguish.

PR #314, opened 2022-11-11
impl TryFrom<&[T]> for Simd

### LINKED ISSUE #194: impl `TryFrom<&[T]>` for Simd?
now that we have added `Simd::from_slice` (see #177), which panics when the len is not enough to construct the Simd; I think we may need to consider adding a `TryFrom` impl to let the users handle errors in case the conversion fails.

calebzulawski:
Closes #194

### DIFF
--- a/crates/core_simd/src/vector.rs
+++ b/crates/core_simd/src/vector.rs
@@ -650,6 +650,30 @@ where
     }
 }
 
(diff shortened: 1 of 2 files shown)


programmerjake:
maybe add some tests for inputs with length `<`, `==`, and `>` the lane count?

programmerjake (approved):

PR #318, opened 2022-11-28
Avoid a scalar loop in `Simd::from_slice`

thomcc:
Discussed on Zulip. This also bit @the8472 when converting the SIMD string search patch to `std::simd` IIRC.

CC @rust-lang/wg-const-eval in case anybody cares that we're using `#![feature(const_ptr_read)]` here (Rationale: This whole API is unstable anyway, and this can be implemented without it, just it doesn't get vectorized).

### DIFF
--- a/crates/core_simd/src/lib.rs
+++ b/crates/core_simd/src/lib.rs
@@ -1,5 +1,6 @@
 #![no_std]
 #![feature(
+    const_ptr_read,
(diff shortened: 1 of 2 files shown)


calebzulawski (approved):

oli-obk:
that feature is fine to use

Issue #319, opened 2022-11-30
non-power-of-2 Simd types have wrong size


programmerjake:
I tried this code:
https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=f4ee5c373508bcb37be6cda3bb627b87
```rust
#![feature(repr_simd)]
use core::alloc::Layout;
#[repr(simd)]
struct Simd<T, const N: usize>(pub [T; N]);

fn main() {
    assert_eq!(
        Layout::new::<Simd<i8, 3>>(),
        Layout::from_size_align(3, 1).unwrap()
    );
}
```

I expected to see this happen: run with no errors because `Simd<T, N>` types should be the same size as the corresponding array type `[T; N]` (something we unofficially agreed on iirc, but is not formally decided)

Instead, this happened:
```
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `Layout { size: 4, align: 4 (1 << 2) }`,
 right: `Layout { size: 3, align: 1 (1 << 0) }`', src/main.rs:7:5
```

### Meta

`rustc --version --verbose`:
```
<current nightly as of issue creation -- probably 2022-11-29 nightly>
```

`crate version in Cargo.toml`:
N/A

programmerjake:
see also #63

programmerjake:
see some previous discussion: https://rust-lang.zulipchat.com/#narrow/stream/257879-project-portable-simd/topic/Bytemuckable.20Simd/near/311734698

programmerjake:
imho we should define layout such that `sizeof(Simd<T, N>) == sizeof([T; N])` and `alignof(Simd<T, N>) == gcd(alignof(T), llvm_alignof(Simd<T, N>)) >= alignof(T)`
or, if we don't want to depend on llvm, we could pick the largest power of two `<= sizeof([T; N])` such that no padding is required.

programmerjake:
Note the changes to vector layout by @RalfJung in https://github.com/rust-lang/rust/commit/891a4da4f3cf99d5ee6413d9a6948031306179af are the exact opposite of what we probably want: that commit adds padding until size is divisible by alignment, what we probably want is to decrease alignment enough that we can enforce vectors are the exact same size as the corresponding array (usually means no padding, there could still be padding if the element type contains padding, portable-simd will presumably enforce that those element types aren't allowed but `repr(simd)` might allow them in the future)

programmerjake:
[...]

PR #331, opened 2023-02-19
Add copy_to_slice

### LINKED ISSUE #324: explicit simd store -- opposite of Simd::from_slice
currently we have to either cast the dest pointer to `Simd<T, N>` and use `write_unaligned` or use an array-typed store, this is somewhat likely to cause issues at some point. plus, having a function on `Simd` is nicer than having to figure out to use `*dest = *val.as_array()`.

calebzulawski:
Closes #324

### DIFF
--- a/crates/core_simd/src/vector.rs
+++ b/crates/core_simd/src/vector.rs
@@ -174,6 +174,7 @@ where
             slice.len() >= LANES,
             "slice length must be at least the number of lanes"
         );
+        assert!(core::mem::size_of::<Self>() == LANES * core::mem::size_of::<T>());
         // Safety:
         // - We've checked the length is sufficient.
         // - `T` and `Simd<T, N>` are Copy types.
@@ -203,6 +204,7 @@ where
             slice.len() >= LANES,
             "slice length must be at least the number of lanes"
         );
+        assert!(core::mem::size_of::<Self>() == LANES * core::mem::size_of::<T>());
         // Safety:
         // - We've checked the length is sufficient
         // - `T` and `Simd<T, N>` are Copy types.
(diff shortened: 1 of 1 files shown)


crates/core_simd/src/vector.rs
```diff
@@ -180,6 +180,35 @@ where
...
+    /// ```
+    pub fn copy_to_slice(self, slice: &mut [T]) {
+        assert!(
+            slice.len() >= LANES,
+            "slice length must be at least the number of lanes"
+        );
+        // Safety:
+        // - We've checked the length is sufficient
+        // - `T` and `Simd<T, N>` are Copy types.
+        unsafe { slice.as_mut_ptr().cast::<Self>().write_unaligned(self) }
```
programmerjake, line 209: what happens if the simd type has padding (e.g. for non-power-of-2 types where #319 isn't fixed yet)?

imho we should add an assert that the simd type and corresponding array type have the same size, otherwise `write_unaligned` is permitted to overwrite after-the-end elements with `undef` padding -- unsound. panicking is better than UB and the assert will never fire once #319 is fixed.

programmerjake (changes requested):

crates/core_simd/src/vector.rs
programmerjake: also, you might want an explicit `Copy` bound since we might add `Simd<&mut _, _>` support -- which isn't Copy.

calebzulawski: Good point about the padding.  I think I'll leave out the `Copy` bound for because basically every one of `Simd`'s associated functions fail if the element isn't `Copy`, that would probably need a much more significant overhaul.

programmerjake (approved):

workingjubilee (approved):
I kinda want the name to be something like `copy_into`, but maybe that makes it too generic-sounding??? Otherwise seems cool though!

PR #333, opened 2023-03-08
Document alternatives to Mask<T, LANES> that guarantee layout. Fixes #332

### LINKED ISSUE #332: Support avx512 bitmasks with dynamic feature detection
The choice between `full_masks.rs` and `bitmask.rs` seems to be made at compile time (https://github.com/rust-lang/portable-simd/blob/9bd30e77b3a3c699af102ebb3df0f6110f8aa02e/crates/core_simd/src/masks.rs#L5) based on compile-time presence of the `avx512f` target feature.

When compiling for general x86-64 targets but doing runtime feature detection for `avx512f`, the current approach will generate suboptimal code for avx512, because it will use the `full_masks.rs` implementation instead of the `bitmask.rs` implementation.

I'd like to be able to use bitmasks in avx512 even when using runtime feature detection. It's probably out of scope (and, I think, undesirable) for `std::simd` to be responsible for runtime feature detection, so one option would be for `std::simd` to be refactored to expose _both_ `Mask` and `Bitmask` as two separate types -- perhaps both implementing the same set of traits. This way, callers could choose between a full-mask implementation and a bitmask implementation based on whatever logic they choose, including e.g. runtime feature detection.

reinerp:
Intended to fix #332.

### DIFF
--- a/crates/core_simd/src/masks.rs
+++ b/crates/core_simd/src/masks.rs
@@ -89,9 +89,10 @@ impl_element! { isize }
 /// and/or Rust versions, and code should not assume that it is equivalent to
 /// `[T; LANES]`.
 ///
-/// For a type with layout guaranteed equivalent to `[T; LANES]`, use
-/// `Simd<T, LANES>`. For a type with layout guaranteed to use 1 bit per
-/// lane (padded up to the next integer type), use [`ToBitMask::BitMask`].
+/// For a type with size guaranteed equivalent to `[T; LANES]`, use
+/// `Simd<T, LANES>`. For a type with size guaranteed to use 1 bit per
+/// lane (padded up to the next integer type or), use [`ToBitMask::BitMask`]
+/// or (with crate feature `generic_const_exprs`) `ToBitMaskArray::BitMaskArray`.
 #[repr(transparent)]
 pub struct Mask<T, const LANES: usize>(mask_impl::Mask<T, LANES>)
 where
--- a/crates/core_simd/src/masks.rs
+++ b/crates/core_simd/src/masks.rs
@@ -88,6 +88,10 @@ impl_element! { isize }
 /// The layout of this type is unspecified, and may change between platforms
 /// and/or Rust versions, and code should not assume that it is equivalent to
 /// `[T; LANES]`.
+///
+/// For a type with layout guaranteed equivalent to `[T; LANES]`, use 
+/// `SIMD<T, LANES>`. For a type with layout guaranteed to use 1 bit per
+/// lane (padded up to full bytes), use `LANES::BitMask`.
 #[repr(transparent)]
 pub struct Mask<T, const LANES: usize>(mask_impl::Mask<T, LANES>)
 where
--- a/crates/core_simd/src/masks.rs
+++ b/crates/core_simd/src/masks.rs
@@ -89,7 +89,7 @@ impl_element! { isize }
 /// and/or Rust versions, and code should not assume that it is equivalent to
 /// `[T; LANES]`.
 ///
-/// For a type with size guaranteed equivalent to `[T; LANES]`, use
+/// For a type with size guaranteed equal to `[T; LANES]`, use
 /// `Simd<T, LANES>`. For a type with size guaranteed to use 1 bit per
 /// lane (padded up to the next integer type or), use [`ToBitMask::BitMask`]
 /// or (with crate feature `generic_const_exprs`) `ToBitMaskArray::BitMaskArray`.
--- a/crates/core_simd/src/masks.rs
+++ b/crates/core_simd/src/masks.rs
@@ -90,7 +90,7 @@ impl_element! { isize }
 /// `[T; LANES]`.
 ///
 /// For a type with layout guaranteed equivalent to `[T; LANES]`, use
-/// `SIMD<T, LANES>`. For a type with layout guaranteed to use 1 bit per
+/// `Simd<T, LANES>`. For a type with layout guaranteed to use 1 bit per
 /// lane (padded up to full bytes), use `LANES::BitMask`.
 #[repr(transparent)]
 pub struct Mask<T, const LANES: usize>(mask_impl::Mask<T, LANES>)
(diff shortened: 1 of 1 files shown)


crates/core_simd/src/masks.rs
```diff
@@ -88,6 +88,10 @@ impl_element! { isize }
 /// The layout of this type is unspecified, and may change between platforms
 /// and/or Rust versions, and code should not assume that it is equivalent to
 /// `[T; LANES]`.
+///
+/// For a type with layout guaranteed equivalent to `[T; LANES]`, use
+/// `SIMD<T, LANES>`. For a type with layout guaranteed to use 1 bit per
```
workingjubilee, line 93:
```suggestion
/// `Simd<T, LANES>`. For a type with layout guaranteed to use 1 bit per
```

crates/core_simd/src/masks.rs
```diff
@@ -88,6 +88,10 @@ impl_element! { isize }
...
+/// lane (padded up to full bytes), use `LANES::BitMask`.
```
workingjubilee, line 94: So, the exact number of bytes is an implementation detail (we've used `u64` before and we might choose it again... basically we want to retain the freedom to use an integer of a reasonable size, OR to use a byte array, depending on what strikes the best ergonomics/codegen balance). Right now, this seems like it might suggest it uses exactly the next byte in size. I am not sure what would be most clear to express our desired reservation while providing you the minimum of guarantees you want (or if you even consider that acceptable).

workingjubilee (changes requested):

workingjubilee:
I've been working on a generic integers RFC with the intention that in the future we can simply use those, but it might still be that rounding up in some cases is preferred, unfortunately.

programmerjake:
> 

true, e.g. on SVP64 all mask vectors being a single `u64` internally (when `LANES <= 64`) is likely most efficient

crates/core_simd/src/masks.rs
```diff
@@ -88,6 +88,10 @@ impl_element! { isize }
 /// The layout of this type is unspecified, and may change between platforms
 /// and/or Rust versions, and code should not assume that it is equivalent to
 /// `[T; LANES]`.
+///
+/// For a type with layout guaranteed equivalent to `[T; LANES]`, use
+/// `Simd<T, LANES>`. For a type with layout guaranteed to use 1 bit per
+/// lane (padded up to full bytes), use `LANES::BitMask`.
```
calebzulawski, line 94: The type is actually `ToBitMask::BitMask`: https://doc.rust-lang.org/std/simd/trait.ToBitMask.html#associatedtype.BitMask.  It's not padded up to the next byte, but the next integer.

calebzulawski (changes requested):

workingjubilee:
And in the original AVX512 case mentioned, the plausible masks use hardware representation of a `u16` and a `u64` (even if they are logically much smaller at times), and anything else is a peculiar choice.

reinerp:
Thanks everyone for the helpful comments.

Per Caleb's comment https://github.com/rust-lang/portable-simd/issues/332#issuecomment-1462898103 this guarantee is much less important to me than it was previously. Even so, my suspicion is that it's worth still providing some pointers in the documentation.

Indeed, `ToBitMask::BitMask` is the type I meant to refer to. The layout of that type appears to be guaranteed by this library, in that it is `u8`/`u16`/`u32`/`u64` rather than an opaque wrapper type around one of those integer types. It seems to me that changing those types to something wider would constitute a breaking change in the library. Likewise, changing those types when compiling for different targets would likely break portability. Am I understanding this right?

So, I *think* the right approach is to document `ToBitMask::BitMask` as "1 bit per lane, padded up to the next integer type", but I'm not confident. Going with this assumption for now, I've updated the PR to state this.

Let me know what you think.

workingjubilee (commented):
Oh is it? Okay! It's... been a while since I've read things and I've lately just had my head down on trying to figure out how we'll make dynamic byte swizzles work and catching up with other stuff.

As a nightly feature, the API is still unstable, but we do try to keep things broadly consistent. But yes, you're right, changing them overly much from target to target when compiling things would at the very least probably make things a bit confusing and thus also make it easy for people to make mistakes.

programmerjake:
do note that `Simd<T, N>` isn't guaranteed layout equivalent to `[T; N]` -- `Simd` often has greater alignment and may currently have some padding. #319 is trying to change that so `Simd` will never have padding (beyond what `T` itself has) and still may have greater alignment.

programmerjake:
also, `ToBitmask` won't necessarily be implemented for all lane counts, e.g. there currently is no integer type with >128 bits so `Simd<T, 129>` likely won't impl `ToBitmask`.

if you want a bitmask type that always exists no matter the lane count, use `ToBitMaskArray`, though that depends on the portable-simd crate's `generic_const_exprs` cfg feature, which isn't currently enabled by `std`.

programmerjake:
> do note that `Simd<T, N>` isn't guaranteed layout equivalent to `[T; N]` -- `Simd` often has greater alignment and may currently have some padding.

this means that "size guaranteed equal to `[T; LANES]`, use `Simd<T, LANES>`" is currently incorrect since `size_of::<Simd<T, N>>() >= size_of::<[T; N]>()`, not `size_of::<Simd<T, N>>() == size_of::<[T; N]>()`. this is what I created #319 to fix.

crates/core_simd/src/masks.rs
```diff
@@ -88,6 +88,11 @@ impl_element! { isize }
 /// The layout of this type is unspecified, and may change between platforms
 /// and/or Rust versions, and code should not assume that it is equivalent to
 /// `[T; LANES]`.
+///
+/// For a type with size guaranteed equal to `[T; LANES]`, use
```
workingjubilee, line 92:
```suggestion
/// For a type with size guaranteed equal-or-greater-than `[T; LANES]`, use
```

programmerjake:
```suggestion
/// For a type equivalent to `[T; LANES]` (with possible trailing padding for alignment purposes), use
```
maybe equivalent is too strong? ideas?

programmerjake: > ```
> /// For a type with size guaranteed equal-or-greater-than `[T; LANES]`, use
> ```

imho that loses most meaning, e.g. `[u8; 5000]` fits but is unlikely to be desired.

workingjubilee: "equivalent" is indeed too strong.

PR #334, opened 2023-03-12
Implement dynamic byte-swizzle prototype

workingjubilee:
This is meant to be an example that is used to test a Rust intrinsic against, which will replace it. The interface is fairly direct and doesn't address more nuanced or interesting permutations one can do, nevermind on types other than bytes.

The ultimate goal is for direct LLVM support for this.

### DIFF
--- a/crates/core_simd/src/swizzle_dyn.rs
+++ b/crates/core_simd/src/swizzle_dyn.rs
@@ -8,6 +8,11 @@ where
     /// Swizzle a vector of bytes according to the index vector.
     /// Indices within range select the appropriate byte.
     /// Indices "out of bounds" instead select 0.
+    ///
+    /// Note that the current implementation is selected during build-time
+    /// of the standard library, so `cargo build -Zbuild-std` may be necessary
+    /// to unlock better performance, especially for larger vectors.
+    /// A planned compiler improvement will enable using `#[target_feature]` instead.
     #[inline]
     pub fn swizzle_dyn(self, idxs: Simd<u8, N>) -> Self {
         #![allow(unused_imports, unused_unsafe)]
@@ -74,14 +79,22 @@ unsafe fn avx2_pshufb(bytes: Simd<u8, 32>, idxs: Simd<u8, 32>) -> Simd<u8, 32> {
     let high = mid + mid;
     // SAFETY: Caller promised AVX2
     unsafe {
-        // This is ordering sensitive: permutes fight for the same ports
-        // and LLVM will order these instructions how you put them here.
-        // Try break up the permutes a bit accordingly.
+        // This is ordering sensitive, and LLVM will order these how you put them.
+        // Most AVX2 impls use ~5 "ports", and only 1 or 2 are capable of permutes.
+        // But the "compose" step will lower to ops that can also use at least 1 other port.
+        // So this tries to break up permutes so composition flows through "open" ports.
+        // Comparative benches should be done on multiple AVX2 CPUs before reordering this
+
         let hihi = avx2_cross_shuffle::<0x11>(bytes.into(), bytes.into());
-        let hi_shuf = Simd::from(avx2_half_pshufb(hihi, idxs.into()));
+        let hi_shuf = Simd::from(avx2_half_pshufb(
+            hihi,        // duplicate the vector's top half
+            idxs.into(), // so that using only 4 bits of an index still picks bytes 16-31
+        ));
+        // A zero-fill during the compose step gives the "all-Neon-like" OOB-is-0 semantics
         let compose = idxs.simd_lt(high).select(hi_shuf, Simd::splat(0));
         let lolo = avx2_cross_shuffle::<0x00>(bytes.into(), bytes.into());
         let lo_shuf = Simd::from(avx2_half_pshufb(lolo, idxs.into()));
+        // Repeat, then pick indices < 16, overwriting indices 0-15 from previous compose step
         let compose = idxs.simd_lt(mid).select(lo_shuf, compose);
         compose
     }
--- a/crates/core_simd/src/swizzle_dyn.rs
+++ b/crates/core_simd/src/swizzle_dyn.rs
@@ -1,0 +1,142 @@
+use crate::simd::{LaneCount, Simd, SupportedLaneCount};
+use core::mem;
+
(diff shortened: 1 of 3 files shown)


workingjubilee:
The API isn't perfect as we will want to figure out something so that, if we do indeed want to support this function beyond an N of 128~256, we can have this use 16-bit indices instead like RVV allows.

However, that's not necessarily the biggest concern, and this is too important as a functionality to worry about "what if it needs two functions, one for each index type?" Then it needs two functions! We will endure.

calebzulawski:
Are 16 bit indices as universal as 8 bit?

crates/core_simd/src/swizzle_dyn.rs
```diff
@@ -0,0 +1,142 @@
+use crate::simd::{LaneCount, Simd, SupportedLaneCount};
+use core::mem;
+
+impl<const N: usize> Simd<u8, N>
+where
+    LaneCount<N>: SupportedLaneCount,
+{
+    /// Swizzle a vector of bytes according to the index vector.
+    /// Indices within range select the appropriate byte.
+    /// Indices "out of bounds" instead select 0.
```
calebzulawski, line 10: Maybe add a note that this really needs `build-std` to work correctly

workingjubilee: Well, it kinda doesn't, does it? The vectors are generic, so this will get instantiated at compile time. What it needs is to be combined with `target_feature` configuration, either dynamic multiversioning or compile-time versioning or whatnot.

workingjubilee:
Hmm. Honestly, 16-bit indices are _only_ used, afaik, by RISC-V's Vector extension.

crates/core_simd/src/swizzle_dyn.rs
calebzulawski: The `cfg`s will depend on the features std is built with, unfortunately

workingjubilee: Oooh good point hmmmmmm...

...I guess I'll make this dynamically multiversioned lol.

workingjubilee: That part, at least, will be fixed upon promoting this into an intrinsic.

programmerjake:
> Hmm. Honestly, 16-bit indices are _only_ used, afaik, by RISC-V's Vector extension. So I think saying "nah, use a target intrinsic for that" would be fair. It's mostly "if we find a way to seamlessly transition to larger indices, that would be cool".

afaict avx512 supports 16-bit indexes for [`vpermi2w` and related operations](https://www.felixcloutier.com/x86/vpermi2w:vpermi2d:vpermi2q:vpermi2ps:vpermi2pd).

SimpleV supports 8/16/32/64-bit indexes iirc.

crates/core_simd/src/swizzle_dyn.rs
```diff
@@ -0,0 +1,142 @@
...
+    use x86::_mm256_shuffle_epi8 as avx2_half_pshufb;
+    let mid = Simd::splat(16u8);
+    let high = mid + mid;
+    // SAFETY: Caller promised AVX2
+    unsafe {
+        // This is ordering sensitive: permutes fight for the same ports
+        // and LLVM will order these instructions how you put them here.
+        // Try break up the permutes a bit accordingly.
+        let hihi = avx2_cross_shuffle::<0x11>(bytes.into(), bytes.into());
+        let hi_shuf = Simd::from(avx2_half_pshufb(hihi, idxs.into()));
```
jhorstmann, line 81: Very nice, the logic looks good to me. The `idxs.into()` might warrant a comment that `shufb` only uses the lower 4 bits of indices, and so selecting index 16 would be the same as selecting index 0 of the high half.

workingjubilee:
The AVX512 version doesn't really matter to the abstract operation we're defining because that instruction uses indices that have the same size as the type, _because it overwrites the index vector with the results_ (subject to the mask). And that's to be expected for AVX512F because without the AVX512BW extension, you can't do byte-level operations at all. But we don't really care about that because destructive update on the indices is a pretty unusual pattern, a form using the instruction can be yielded by combining it with a masked store, and what is actually relevant for what we're doing is whether a u8 will be big enough.

crates/core_simd/src/swizzle_dyn.rs
```diff
@@ -0,0 +1,155 @@
...
+    use x86::_mm256_shuffle_epi8 as avx2_half_pshufb;
+    let mid = Simd::splat(16u8);
+    let high = mid + mid;
+    // SAFETY: Caller promised AVX2
+    unsafe {
+        // This is ordering sensitive, and LLVM will order these how you put them.
+        // Most AVX2 impls use ~5 "ports", and only 1 or 2 are capable of permutes.
+        // But the "compose" step will lower to ops that can also use at least 1 other port.
+        // So this tries to break up permutes so composition flows through "open" ports.
+        // Comparative benches should be done on multiple AVX2 CPUs before reordering this
```
workingjubilee, line 86: Having all this commentary here isn't strictly necessary but I'm going to transplant more-or-less the same remarks into rustc (and maybe into LLVM???) later, so writing this down matters.

crates/core_simd/src/swizzle_dyn.rs
```diff
@@ -0,0 +1,155 @@
...
+
+        let hihi = avx2_cross_shuffle::<0x11>(bytes.into(), bytes.into());
+        let hi_shuf = Simd::from(avx2_half_pshufb(
+            hihi,        // duplicate the vector's top half
+            idxs.into(), // so that using only 4 bits of an index still picks bytes 16-31
```
workingjubilee, line 91: I think this clarifies what @jhorstmann was remarking on.

workingjubilee (commented):
As funny as it would be to package a CPUID implementation to handle the which-AVX-version stuff, I think I am gonna skip it for now. I could have completely skipped implementing this "up here" at the "tip", but I wanted to have full testing in our suite against proptest, first, which is very good at finding counterfactuals.

Dr-Emann:
Is there any tracking for making this an intrinsic to avoid having to use `build-std` to get reasonable performance on e.g. x86_64 since std isn't usually built with any of the features to enable the simd versions?

Issue #277, opened 2022-04-14
Concat and split operations


clarfonthey:
I've recently found myself coding up shorthand methods for operations like:

```rust
fn concat(lhs: f64x4, rhs: f64x4) -> f64x8;
fn split(val: f64x8) -> (f64x4, f64x4);
```

which are shorthands for their various swizzle representations. I tried messing around with the code for the crate to make these work but got stuck with dependencies on `generic_const_exprs` and couldn't find a way to make it work.

Since the `simd_swizzle` macro can be a bit overwhelming, it would be nice to have more shorthands like these, but I'm not sure exactly how it would fit within the existing design plans, so, I figured I'd just start a discussion here.

One big problem with these methods under the existing, general implementation is checking that a size is actually supported: for example, `f64x16` doesn't exist but `f32x16` does.

programmerjake:
> One big problem with these methods under the existing, general implementation is checking that a size is actually supported: for example, `f64x16` doesn't exist but `f32x16` does.

Wether or not an element count is supported is independent of the element type, we intentionally decided on that to make generic code easier to write/use. Just the `f64x16` type alias is missing, `Simd<f64, 16>` works:
https://play.rust-lang.org/?version=nightly&mode=release&edition=2021&gist=93d6280b54e054de94bbf17c9c1227d4

clarfonthey:
Oh, interesting. Well, regardless, there is an upper limit to the element size, e.g. `Simd<T, 128>` isn't allowed. Unless the goal is to support all powers-of-two `usize`, which... I'm assuming isn't the case, given the `SupportedLaneCount` trait.

calebzulawski:
Not that I necessarily recommend it, but this would be possible to implement with a trait instead of generic const expressions.  Either way, a function like this is probably useful.

clarfonthey:
So, I *did* try and make it work with an extension trait called `DoubleableLaneCount` and due to limitations in the existing const generics impl, associated constants sometimes aren't allowed in generic const position. Could work on a trait that explicitly returns the `Simd` type, however, but I didn't see any existing ones.

PR #335, opened 2023-03-13
Implement common swizzle operations.

reinerp:
`concat` and `slice` address #277.

`general_reverse` is useful for horizontal reductions, bitonic sorting, and many similar cases.

### DIFF
--- a/crates/core_simd/src/swizzle.rs
+++ b/crates/core_simd/src/swizzle.rs
@@ -364,4 +364,110 @@ where
 
         (Even::swizzle2(self, other), Odd::swizzle2(self, other))
     }
+
+    /// Takes a slice of a vector to produce a shorter vector.
+    ///
+    /// This is equivalent to computing `&self[OFFSET..OFFSET+LEN]` on
+    /// the underlying array.
+    /// 
+    /// ```
+    /// # #![feature(portable_simd)]
+    /// # use core::simd::Simd;
+    /// let x = Simd::from_array([0, 1, 2, 3, 4, 5, 6, 7]);
+    /// let y = x.slice::<2, 4>();
+    /// assert_eq!(y.to_array(), [2, 3, 4, 5]);
+    /// ```
+    /// 
+    /// Will be rejected at compile time if `OFFSET + LEN > LANES`.
+    #[inline]
+    #[must_use = "method returns a new vector and does not mutate the original inputs"]
+    pub fn slice<const OFFSET: usize, const LEN: usize>(self) -> Simd<T, LEN> 
+        where LaneCount<LEN>: SupportedLaneCount {
+        const fn slice_index<const LEN: usize>(offset: usize, lanes: usize) -> [usize; LEN] {
+            assert!(offset + LEN <= lanes, "slice out of bounds");
+            let mut index = [0; LEN];
+            let mut i = 0;
+            while i < LEN {
+                index[i] = i + offset;
+                i += 1;
+            }
+            index
+        }
+        struct Slice<const OFFSET: usize>;
+        impl<const OFFSET: usize, const LEN: usize, const LANES: usize> Swizzle<LANES, LEN> for Slice<OFFSET> {
+            const INDEX: [usize; LEN] = slice_index::<LEN>(OFFSET, LANES);
+        }
+        Slice::<OFFSET>::swizzle(self)
+    }
+
+    /// Concatenates two vectors of equal length.
+    /// 
+    /// Due to limitations in const generics, the length of the resulting vector cannot be inferred
+    /// from the input vectors. You must specify it explicitly.
+    /// 
+    /// ```
+    /// # #![feature(portable_simd)]
+    /// # use core::simd::Simd;
+    /// let x = Simd::from_array([0, 1, 2, 3]);
+    /// let y = Simd::from_array([4, 5, 6, 7]);
+    /// let z = x.concat_to::<8>(y);
+    /// assert_eq!(z.to_array(), [0, 1, 2, 3, 4, 5, 6, 7]);
+    /// ```
+    /// 
+    /// Will be rejected at compile time if `LANES * 2 != DOUBLE_LANES`.
+    pub fn concat_to<const DOUBLE_LANES: usize>(self, other: Self) -> Simd<T, DOUBLE_LANES>
+        where LaneCount<DOUBLE_LANES>: SupportedLaneCount
+    {
+        const fn concat_index<const DOUBLE_LANES: usize>(lanes: usize) -> [Which; DOUBLE_LANES] {
+            assert!(lanes * 2 == DOUBLE_LANES);
+            let mut index = [Which::First(0); DOUBLE_LANES];
+            let mut i = 0;
+            while i < lanes {
+                index[i] = Which::First(i);
+                index[i + lanes] = Which::Second(i);
+                i += 1;
+            }
+            index
+        }
+        struct Concat;
+        impl<const LANES: usize, const DOUBLE_LANES: usize> Swizzle2<LANES, DOUBLE_LANES> for Concat {
+            const INDEX: [Which; DOUBLE_LANES] = concat_index::<DOUBLE_LANES>(LANES);
+        }
+        Concat::swizzle2(self, other)
+    }
+
+    /// For each lane `i`, swaps it with lane `i ^ SWAP_MASK`.
+    /// 
+    /// Also known as `grev` in the RISC-V Bitmanip specification, this is a powerful
+    /// swizzle operation that can implement many common patterns as special cases.
+    /// 
+    /// ```
+    /// # #![feature(portable_simd)]
+    /// # use core::simd::Simd;
+    /// let x = Simd::from_array([0, 1, 2, 3, 4, 5, 6, 7]);
+    /// // Swap adjacent lanes:
+    /// assert_eq!(x.general_reverse::<1>().to_array(), [1, 0, 3, 2, 5, 4, 7, 6]);
+    /// // Swap lanes separated by distance 2:
+    /// assert_eq!(x.general_reverse::<2>().to_array(), [2, 3, 0, 1, 6, 7, 4, 5]);
+    /// // Swap lanes separated by distance 4:
+    /// assert_eq!(x.general_reverse::<4>().to_array(), [4, 5, 6, 7, 0, 1, 2, 3]);
+    /// // Reverse lanes, within each 4-lane group:
+    /// assert_eq!(x.general_reverse::<3>().to_array(), [3, 2, 1, 0, 7, 6, 5, 4]);
+    /// ```
+    pub fn general_reverse<const SWAP_MASK: usize>(self) -> Self {
+        const fn general_reverse_index<const LANES: usize>(swap_mask: usize) -> [usize; LANES] {
+            let mut index = [0; LANES];
+            let mut i = 0;
+            while i < LANES {
+                index[i] = i ^ swap_mask;
+                i += 1;
+            }
+            index
+        }
+        struct GeneralReverse<const DISTANCE: usize>;
+        impl<const LANES: usize, const DISTANCE: usize> Swizzle<LANES, LANES> for GeneralReverse<DISTANCE> {
+            const INDEX: [usize; LANES] = general_reverse_index::<LANES>(DISTANCE);
+        }
+        GeneralReverse::<SWAP_MASK>::swizzle(self)
+    }
 }
--- a/crates/core_simd/src/swizzle.rs
+++ b/crates/core_simd/src/swizzle.rs
@@ -369,7 +369,7 @@ where
     ///
     /// This is equivalent to computing `&self[OFFSET..OFFSET+LEN]` on
     /// the underlying array.
-    /// 
+    ///
     /// ```
     /// # #![feature(portable_simd)]
     /// # use core::simd::Simd;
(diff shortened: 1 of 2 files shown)


crates/core_simd/src/swizzle.rs
```diff
@@ -364,4 +364,131 @@ where
...
+    ///
+    /// This is equivalent to computing `&self[OFFSET..OFFSET+LEN]` on
+    /// the underlying array.
+    ///
+    /// ```
+    /// # #![feature(portable_simd)]
+    /// # use core::simd::Simd;
+    /// let x = Simd::from_array([0, 1, 2, 3, 4, 5, 6, 7]);
+    /// let y = x.slice::<2, 4>();
+    /// assert_eq!(y.to_array(), [2, 3, 4, 5]);
```
workingjubilee, line 378: The example is good on its own, but I'd like to know: what sort of algorithms are you going to use this one in? I'd like to see this one in action. For the benefit of deciding what this should look like in the future, not necessarily for inclusion in the docs.

crates/core_simd/src/swizzle.rs
```diff
@@ -364,4 +364,131 @@ where
...
+    /// let x = Simd::from_array([0, 1, 2, 3]);
+    /// let y = Simd::from_array([4, 5, 6, 7]);
+    /// let z = x.concat_to::<8>(y);
+    /// assert_eq!(z.to_array(), [0, 1, 2, 3, 4, 5, 6, 7]);
+    /// ```
+    ///
+    /// Will be rejected at compile time if `LANES * 2 != DOUBLE_LANES`.
+    #[inline]
+    #[must_use = "method returns a new vector and does not mutate the original inputs"]
+    pub fn concat_to<const DOUBLE_LANES: usize>(self, other: Self) -> Simd<T, DOUBLE_LANES>
```
workingjubilee, line 424: `concat_to`? Why not just `concat`?

crates/core_simd/src/swizzle.rs
```diff
@@ -364,4 +364,131 @@ where
...
+    /// # use core::simd::Simd;
+    /// let x = Simd::from_array([0u32, 1, 2, 3, 4, 5, 6, 7]);
+    /// let x = x + x.general_reverse::<1>();
+    /// let x = x + x.general_reverse::<2>();
+    /// let x = x + x.general_reverse::<4>();
+    /// assert_eq!(x.to_array(), [28, 28, 28, 28, 28, 28, 28, 28]);
+    /// ```
+    #[inline]
+    #[must_use = "method returns a new vector and does not mutate the original inputs"]
+    pub fn general_reverse<const SWAP_MASK: usize>(self) -> Self {
```
programmerjake, line 478: imho this should include something like `lanewise` in it's name, since I expect Rust to gain a bitwise `grev` at some point and we'd want the bitwise simd and scalar integer operations to have matching names (probably `gen_rev` or `generalized_reverse` or `grev`?).

crates/core_simd/src/swizzle.rs
```diff
@@ -364,4 +364,131 @@ where
...
+        struct Concat;
+        impl<const LANES: usize, const DOUBLE_LANES: usize> Swizzle2<LANES, DOUBLE_LANES> for Concat {
+            const INDEX: [Which; DOUBLE_LANES] = concat_index::<DOUBLE_LANES>(LANES);
+        }
+        Concat::swizzle2(self, other)
+    }
+
+    /// For each lane `i`, swaps it with lane `i ^ SWAP_MASK`.
+    ///
+    /// Also known as `grev` in the RISC-V Bitmanip specification, this is a powerful
```
programmerjake, line 448: you may want to mention that `grev` in RISC-V B is bitwise, not lanewise.

crates/core_simd/src/swizzle.rs
```diff
@@ -364,4 +364,131 @@ where
...
     }
+
+    /// Takes a slice of a vector to produce a shorter vector.
+    ///
+    /// This is equivalent to computing `&self[OFFSET..OFFSET+LEN]` on
+    /// the underlying array.
+    ///
+    /// ```
+    /// # #![feature(portable_simd)]
+    /// # use core::simd::Simd;
```
workingjubilee, line 375: This needs to look more like this:
https://github.com/rust-lang/portable-simd/blob/8dcb4d5ef745dcbe62c77f25415c21fcbe9f165c/crates/core_simd/src/elements/int.rs#L18-L21

crates/core_simd/src/swizzle.rs
workingjubilee: For something that effectively implements a very specific xor-striding pattern, `general_reverse` is a non-descript name.

programmerjake: well, `grev` *is* what that particular bitwise op has become named, thanks to RISC-V afaict.

programmerjake: grev -- general bit reverse

workingjubilee: Just because an ISA has picked a terrible name doesn't mean we need to copy it.

crates/core_simd/src/swizzle.rs
programmerjake, line 424: imho `concat` should support concatenating vectors of arbitrary length, e.g. `Simd::concat::<15>(my_i32x5, my_i32x10)`

crates/core_simd/src/swizzle.rs
```diff
@@ -364,4 +364,131 @@ where
...
+    ///
+    /// Will be rejected at compile time if `LANES * 2 != DOUBLE_LANES`.
+    #[inline]
+    #[must_use = "method returns a new vector and does not mutate the original inputs"]
+    pub fn concat_to<const DOUBLE_LANES: usize>(self, other: Self) -> Simd<T, DOUBLE_LANES>
+    where
+        LaneCount<DOUBLE_LANES>: SupportedLaneCount,
+    {
+        const fn concat_index<const DOUBLE_LANES: usize>(lanes: usize) -> [Which; DOUBLE_LANES] {
+            assert!(lanes * 2 == DOUBLE_LANES);
```
workingjubilee, line 429: Just checking: Are you aware that when you put something in a `const fn`, and then the `const fn` is evaluated at runtime, that this means an `assert!` in it will not be evaluated at compilation time, but rather will be evaluated at runtime?

crates/core_simd/src/swizzle.rs
programmerjake: well, it gets the name because it's a generalization of bitwise reverse and all byte swaps for reversing endian -- e.g. `grev(v, 0x3F)` is 64-bit bitwise reverse, `grev(v, 0x38)` reverses all bytes in a 64-bit value, `grev(v, 0x18)` reverses bytes in each 32-bit part, `grev(v, 0x8)` reverses bytes in each 16-bit part, etc.

imho `grev` is the name people use even if it wasn't the best originally, so we should use it for consistency and searchability.

crates/core_simd/src/swizzle.rs
programmerjake: untested code for concatenating arbitrary vector lengths:
```rust
impl<T, const LANES: usize> Simd<T, LANES>
where
    T: SimdElement,
{
    /// use like `a.concat::<15, _>(b)`
    pub fn concat<const OUTPUT_LANES: usize, const LANES2: usize>(self, other: Simd<T, LANES2>) -> Simd<T, OUTPUT_LANES>
    where
        LaneCount<OUTPUT_LANES>: SupportedLaneCount,
        LaneCount<LANES2>: SupportedLaneCount,
    {
        struct Extend;
        impl<const I: usize, const O: usize> Swizzle<I, O> for Extend {
            const INDEX: [usize; O] = {
                assert!(I <= O);
                let mut retval = [0; O];
                let mut i = 0;
                while i < I {
                    retval[i] = i;
                }
                retval
            };
        }
        struct Concat<const A: usize, const B: usize, const Y: usize>;
        impl<const A: usize, const B: usize, const Y: usize> Swizzle2<Y, Y> for Concat<A, B, Y> {
            const INDEX: [Which; Y] = {
                assert!(A + B == Y, "concat: OUTPUT_LANES must be the sum of all input lane counts");
                let mut retval = [Which::First(0); Y];
                let mut i = 0;
                while i < Y {
                    if i < A {
                        retval[i] = Which::First(i);
                    } else {
                        retval[i] = Which::Second(i - A);
                    }
                }
                retval
            };
        }
        Concat::<LANES, LANES2, OUTPUT_LANES>::swizzle2(Extend::swizzle(self), Extend::swizzle(other))
    }
}
```

crates/core_simd/src/swizzle.rs
programmerjake: it's only used at compile time, so that doesn't matter here.

workingjubilee: Ah, right, I see now, it happens in the associated constant. Then that requires monomorphization which puts it at risk of depending on trait and const evaluation details. Hmm. I do want to allow us to use certain post-monomorphization errors but I currently don't fully understand the evaluation patterns that rustc will do to intuit in what cases this will/will not happen.

crates/core_simd/src/swizzle.rs
workingjubilee: Sure, but does it work with type inference?

reinerp (commented):
Thanks for the very helpful comments! I have switched from `slice` to a `split` API, and have made a few minor cleanups. Tests should pass now, I believe.

crates/core_simd/src/swizzle.rs
programmerjake: > Sure, but does it work with type inference?

about as well as the size-doubling `concat` imho -- you have to specify the output size.

programmerjake: > Given how `Swizzle2` works right now, arbitrary concats are not possible.

they're totally possible, just not using only one `Swizzle2` operation. the example code I posted uses separate swizzles to extend the inputs to the same length so `Swizzle2` can be used to combine them.

> The `Swizzle2` trait assumes that both of the two input vectors have the same length. We could potentially change that if you like, but I don't know all of the consequences of doing so.

`swizzle2` matches LLVM IR's `shuffle` instruction, so if we extend it to accept inputs of different length we would have to generate more complex IR -- basically matching the `concat` demo code I posted.

crates/core_simd/src/swizzle.rs
programmerjake: `swizzle_to_xor_indexes`?

crates/core_simd/src/swizzle.rs
reinerp: I see. I misread your demo code. Thanks, done!

The new variant is slightly worse for type inference, because you can't write `x.concat::<4, _>(y)` without `generic_arg_infer` (https://github.com/rust-lang/rust/issues/85077); you have to write `x.concat::<4, 2>(y)` or whatever. That said, you can still constrain the `OUTPUT_LANES` by context, e.g. by giving a type signature to the result: `let z: u32x4 = x.concat(y);`.

crates/core_simd/tests/swizzle.rs
```diff
@@ -74,3 +74,66 @@ fn interleave_one() {
...
+fn concat_equal_width() {
+    let x = Simd::from_array([0, 1, 2, 3]);
+    let y = Simd::from_array([4, 5, 6, 7]);
+    let z = x.concat(y);
+    assert_eq!(z.to_array(), [0, 1, 2, 3, 4, 5, 6, 7]);
+}
+
+#[test]
+#[cfg_attr(target_arch = "wasm32", wasm_bindgen_test)]
+fn concat_different_width() {
```
programmerjake, line 98:
```suggestion
#[cfg(feature = "all_lane_counts")]
fn concat_different_width() {
```

crates/core_simd/src/swizzle.rs
workingjubilee: Oh hm.

crates/core_simd/src/swizzle.rs
```diff
@@ -364,4 +364,176 @@ where
...
+    /// # #![feature(portable_simd)]
+    /// # #[cfg(feature = "as_crate")] use core_simd::simd::Simd;
+    /// # #[cfg(not(feature = "as_crate"))] use core::simd::Simd;
+    /// let x = Simd::from_array([0, 1, 2, 3]);
+    /// let y = Simd::from_array([4, 5, 6, 7]);
+    /// let z = x.concat(y);
+    /// assert_eq!(z.to_array(), [0, 1, 2, 3, 4, 5, 6, 7]);
+    /// ```
+    ///
+    /// Will be rejected at compile time if `LANES * 2 != DOUBLE_LANES`.
```
pthariensflame, line 430: This assertion is too strict, and also redundant with the one above when relaxed.  It’s also not implemented (only the one above is).  It should just be removed from the docs.

crates/core_simd/src/swizzle.rs
pthariensflame: `reverse_pow2_lane_groups`?  Conceptually this operation performs the reversal of blocks of `k` lanes within `n`-lane groups, where `k` and `n` are both powers of two, `k ≤ n ≤ LANES`, and the choice of `k` and `n` is determined uniquely up to operation uniqueness by choosing where index `0` will be swizzled to (it’ll be exactly `SWAP_MASK`).

programmerjake: > `reverse_pow2_lane_groups`?

afaict `grev` is actually more powerful than that, it can do any arbitrary combination of those `k`-`n` reversals for arbitrary `k` and `n`.

e.g. `grev(v, 5)` is equivalent to `simd_swizzle!(v, [5, 4, 7, 6, 1, 0, 3, 2])` which is a combination of reversing adjacent pairs (blocks of 1) and swapping blocks of 4.

reinerp (commented):
I think all comments are now addressed. Please take another look.

reinerp:
Hi folks, I believe I have responded to all review comments so far.

My sense is there are some areas where opinions differ between people on this thread on the best path forward, namely:
1. Naming of `swizzle_to_xor_indices`.
2. Whether, and in what form, to include `concat`/`slice`, given the limited type inference support due to lack of `generic_const_exprs`.

My sense is that this PR is stuck in review, pending consensus on these decisions. I'm not sure what process your project uses to make decisions in such cases. Could you please advise on the process? I am very excited about the `portable-simd` work in general, and this PR in particular, and am happy to help if I can.

calebzulawski:
I've definitely been neglecting this PR.  Some thoughts:
* Regarding concat and split, there's a `generic_const_exprs` cargo feature that you can put these functions behind.  While this prevents them from making it to the nightly compiler, I think I'd prefer implementing it this way.  It makes it clearer which functions we're waiting on and allows us to test the intended implementation.
* I find the grev function confusing and I'm not sure if this is something general enough to implement in the standard library.  Maybe it is, I'm just not sure.
* Something with `swap_lanes` is probably best, but `swap_lanes_xor` is still a little confusing (since it's the index that's being xor'd, not the lane)

PR #336, opened 2023-03-26
Skip building wasm-bindgen-test on non-wasm targets

bjorn3:
This reduces compilation time


bjorn3:
For a 64 core system this reduces compilation time from 22s to 16s. For a 2 core + HT system this reduces compilation time from 50s to 40s.

calebzulawski:
Thanks!

PR #337, opened 2023-04-06
Remove formats `[T; N]` does not impl

workingjubilee:
This is inconsistent with how arrays and slices format, and it also causes unnecessary code bloat in libcore. We can revisit this if people ever agree on doing these formatters for the other slice-y types.

### DIFF
--- a/crates/core_simd/src/fmt.rs
+++ b/crates/core_simd/src/fmt.rs
@@ -1,39 +1,23 @@
 use crate::simd::{LaneCount, Simd, SimdElement, SupportedLaneCount};
 use core::fmt;
 
-macro_rules! impl_fmt_trait {
-    { $($trait:ident,)* } => {
-        $(
-            impl<T, const LANES: usize> fmt::$trait for Simd<T, LANES>
-            where
-                LaneCount<LANES>: SupportedLaneCount,
-                T: SimdElement + fmt::$trait,
-            {
-                fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-                    #[repr(transparent)]
-                    struct Wrapper<'a, T: fmt::$trait>(&'a T);
+impl<T, const LANES: usize> fmt::Debug for Simd<T, LANES>
+where
+    LaneCount<LANES>: SupportedLaneCount,
+    T: SimdElement + fmt::Debug,
+{
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        #[repr(transparent)]
+        struct Wrapper<'a, T: fmt::Debug>(&'a T);
 
-                    impl<T: fmt::$trait> fmt::Debug for Wrapper<'_, T> {
-                        fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-                            self.0.fmt(f)
-                        }
-                    }
-
-                    f.debug_list()
-                        .entries(self.as_array().iter().map(|x| Wrapper(x)))
-                        .finish()
-                }
+        impl<T: fmt::Debug> fmt::Debug for Wrapper<'_, T> {
+            fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+                self.0.fmt(f)
             }
-        )*
-    }
-}
+        }
 
-impl_fmt_trait! {
-    Debug,
-    Binary,
-    LowerExp,
-    UpperExp,
-    Octal,
-    LowerHex,
-    UpperHex,
+        f.debug_list()
+            .entries(self.as_array().iter().map(|x| Wrapper(x)))
+            .finish()
+    }
 }
--- a/crates/core_simd/src/fmt.rs
+++ b/crates/core_simd/src/fmt.rs
@@ -14,6 +14,7 @@ where
     /// let floats = Simd::<f32, 4>::splat(-1.0);
     /// assert_eq!(format!("{:?}", [-1.0; 4]), format!("{:?}", floats));
     /// ```
+    #[inline]
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         <[T] as fmt::Debug>::fmt(self.as_array(), f)
     }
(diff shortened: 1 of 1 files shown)


crates/core_simd/src/fmt.rs
```diff
@@ -1,39 +1,23 @@
...
-                    #[repr(transparent)]
-                    struct Wrapper<'a, T: fmt::$trait>(&'a T);
+impl<T, const LANES: usize> fmt::Debug for Simd<T, LANES>
+where
+    LaneCount<LANES>: SupportedLaneCount,
+    T: SimdElement + fmt::Debug,
+{
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        #[repr(transparent)]
+        struct Wrapper<'a, T: fmt::Debug>(&'a T);
```
Sp00ph, line 11: Can't all of this just be replaced by `<[T]>::fmt(self.as_array(), f)`?

programmerjake (approved):

Lokathor:
I'm officially strongly against this, simd (and arrays) should implement the formatters of their element type, but fine.

crates/core_simd/src/fmt.rs
calebzulawski: What about simply `Debug::fmt`?

crates/core_simd/src/fmt.rs
```diff
@@ -1,39 +1,20 @@
...
+{
+    /// A `Simd<T, N>` has a debug format like the one for `[T]`:
+    /// ```
+    /// # #![feature(portable_simd)]
+    /// # #[cfg(feature = "as_crate")] use core_simd::simd::Simd;
+    /// # #[cfg(not(feature = "as_crate"))] use core::simd::Simd;
+    /// let floats = Simd::<f32, 4>::splat(-1.0);
+    /// assert_eq!(format!("{:?}", [-1.0; 4]), format!("{:?}", floats));
+    /// ```
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
```
calebzulawski, line 17: Since this is just passed to array's implementation, I think this function should be inline

calebzulawski (changes requested):

workingjubilee:
@Lokathor I wouldn't mind having formatters for arrays like that! It's just not clear we should be ad-libbing this here.

calebzulawski:
As long as it works with `x?` etc I think it's sufficient to match arrays for now

crates/core_simd/src/fmt.rs
workingjubilee: @bjorn3 @Sp00ph What's your estimation of the binary size impact on sth like a `libcore.so` of doing that, here?

crates/core_simd/src/fmt.rs
workingjubilee: @calebzulawski My hope is that forwarding to `[T]`'s formatter reduces overall footprint slightly, and that it should be fine since a test validates the equality. Maybe I will be proven wrong.

crates/core_simd/src/fmt.rs
nbdd0121: Actually this PR doesn't change binary size of `libcore.o` at all, because it's `core::core_arch::simd::` that's using space, not portable simd 😅

nbdd0121: (But I still think it's a good cleanup and it's good to match array in terms of formatting)

Sp00ph: > Actually this PR doesn't change binary size of `libcore.o` at all, because it's `core::core_arch::simd::` that's using space, not portable simd 😅

`core::arch::simd` uses `portable_simd` all over the place internally though.

> @Sp00ph What's your estimation of the binary size impact on sth like a `libcore.so` of doing that, here?

In my testing the .rlib shrunk by about 65K. Not sure how well that translates to the size of the `.so` though.

workingjubilee: > Actually this PR doesn't change binary size of libcore.o at all, because it's `core::core_arch::simd::` that's using space, not portable simd :sweat_smile:

oh right, lemme see...

> core::arch::simd uses portable_simd all over the place internally though.

Eh not quite.

nbdd0121: `core::core_arch::simd` is from `stdarch` crate, not `portable-simd`.

When I say "doesn't change binary size at all", it really means it! I looked at the `size core.o` and the output is exactly the same.

crates/core_simd/src/fmt.rs
calebzulawski: Makes sense! Maybe `as_slice` would be a little more clear, in that case

Sp00ph: There is no `Simd::as_slice()`

crates/core_simd/src/fmt.rs
calebzulawski: My argument for `inline` has less to do with binary size and more that this function simply defers to another function, so it's expected that this will inline (whether or not slice fmt is marked `inline` is another story)

crates/core_simd/src/fmt.rs
calebzulawski: Whoops, true

crates/core_simd/src/fmt.rs
workingjubilee: Hm. Are you accounting for the bottom-up nature of inlining, and thus at the time this function has an inlining decision made on it, it may already contain all the fmt code instantiated inside it, thus a hint to strengthen the inlining would reduce the likelihood of what we want (a single fmt instance)?

calebzulawski: I know it's not quite this simple, but if the slice fmt function is inlined into this function, it should probably be inlined into the caller.  If it isn't, I suspect this function will be inlined regardless of what attribute we put on it.

workingjubilee: Fair enough!

calebzulawski (approved):

PR #338, opened 2023-04-10
lane -> element for `core::simd::Simd`

workingjubilee:
A while ago we began saying T, N instead of T, LANES in reference to Simd. At some point that leaked in to us checking in code with const N: usize. After a while, we had a discussion and agreed that "lanes", while common, is unnecessary jargon for Rust learners who aren't familiar with SIMD, and is fully interchangeable with terms for arrays like element and index.

But we never acted on that. Let's update the main type's docs, at least. VL has been proposed here because Simd appears beside [T; N] often, so getting ahead on assigning a name to Simd's specific N may help? The example tweaks enable removing a slated-for-removal nightly fn.

I have the rest of this as it would go across the rest of the module's API docs waiting in the wings, but I figured that any kibitzing over how this should be done should be handled on Simd's docs because it's pretty procedurally applied elsewhere. There's basically "how we talk about Simd" and "how we talk about Mask" as the two major points of interest, so once both of those are done, and they can be mostly-separate discussions, it's robotic.

### DIFF
--- a/crates/core_simd/src/vector.rs
+++ b/crates/core_simd/src/vector.rs
@@ -709,7 +709,7 @@ where
             Mask::from_int_unchecked(tfvec)
         };
 
-        // Two vectors are equal if they are elementwise equal
+        // Two vectors are equal if all elements are equal when compared elementwise
         mask.all()
     }
 
@@ -722,7 +722,7 @@ where
             Mask::from_int_unchecked(tfvec)
         };
 
-        // Two vectors are non-equal if they are elementwise non-equal
+        // Two vectors are non-equal if any elements are non-equal when compared elementwise
         mask.any()
     }
 }
--- a/crates/core_simd/src/vector.rs
+++ b/crates/core_simd/src/vector.rs
@@ -3,48 +3,55 @@ use crate::simd::{
     SimdPartialOrd, SupportedLaneCount, Swizzle,
 };
 
-/// A SIMD vector of `LANES` elements of type `T`. `Simd<T, N>` has the same shape as [`[T; N]`](array), but operates like `T`.
+/// A SIMD vector with the shape of `[T; N]` but the operations of `T`.
 ///
-/// Two vectors of the same type and length will, by convention, support the operators (+, *, etc.) that `T` does.
-/// These take the lanes at each index on the left-hand side and right-hand side, perform the operation,
-/// and return the result in the same lane in a vector of equal size. For a given operator, this is equivalent to zipping
-/// the two arrays together and mapping the operator over each lane.
+/// `Simd<T, N>` supports the operators (+, *, etc.) that `T` does in "elementwise" fashion.
+/// These take the element at each index from the left-hand side and right-hand side,
+/// perform the operation, then return the result in the same index in a vector of equal size.
+/// In other words, an elementwise operation is equivalent to a zip, then map.
 ///
 /// ```rust
-/// # #![feature(array_zip, portable_simd)]
+/// # #![feature(portable_simd)]
 /// # use core::simd::{Simd};
-/// let a0: [i32; 4] = [-2, 0, 2, 4];
-/// let a1 = [10, 9, 8, 7];
-/// let zm_add = a0.zip(a1).map(|(lhs, rhs)| lhs + rhs);
-/// let zm_mul = a0.zip(a1).map(|(lhs, rhs)| lhs * rhs);
+/// # use core::array;
+/// let a: [i32; 4] = [-2, 0, 2, 4];
+/// let b = [10, 9, 8, 7];
+/// let sum = array::from_fn(|i| a[i] + b[i]);
+/// let prod = array::from_fn(|i| a[i] * b[i]);
 ///
 /// // `Simd<T, N>` implements `From<[T; N]>
-/// let (v0, v1) = (Simd::from(a0), Simd::from(a1));
+/// let (v, w) = (Simd::from(a), Simd::from(b));
 /// // Which means arrays implement `Into<Simd<T, N>>`.
-/// assert_eq!(v0 + v1, zm_add.into());
-/// assert_eq!(v0 * v1, zm_mul.into());
+/// assert_eq!(v + w, sum.into());
+/// assert_eq!(v * w, prod.into());
 /// ```
 ///
-/// `Simd` with integers has the quirk that these operations are also inherently wrapping, as if `T` was [`Wrapping<T>`].
+///
+/// `Simd` with integer elements treats operators as wrapping, as if `T` was [`Wrapping<T>`].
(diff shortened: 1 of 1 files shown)


workingjubilee:
Also yes, I realize it's slightly confusing to not change the actual function names where applicable, yet, but we were already inconsistent. I feel hashing out a bit of how the documentation should look is a necessary first step to updating any of the function names still not yet renamed.

crates/core_simd/src/vector.rs
```diff
@@ -3,70 +3,92 @@ use crate::simd::{
     SimdPartialOrd, SupportedLaneCount, Swizzle,
 };
 
-/// A SIMD vector of `LANES` elements of type `T`. `Simd<T, N>` has the same shape as [`[T; N]`](array), but operates like `T`.
+/// A SIMD vector with the shape of `[T; VL]` but the operations of `T`.
```
calebzulawski, line 6: While I understand the choice of `VL`, I think it's still slightly jargon-ish. I would prefer `N` because it doesn't mean anything, just as `[T; AL]` would be slightly unnecessary.

workingjubilee: I'm not overly attached to it, no. It does make me wonder why the `std` doesn't refer to arrays as `[T; L]` in general, though.

calebzulawski: I think it's helpful that one might see `N` and just think "number". And at worst, it's just another letter

scottmcm: I think it's because one commonly talks of "n elements" and "O(n)" and such, and nobody ever wants to say "O(l)" because fonts, so avoids `L` too.

crates/core_simd/src/vector.rs
```diff
@@ -673,18 +692,18 @@ where
     fn ne(&self, other: &Self) -> bool {
         // Safety: All SIMD vectors are SimdPartialEq, and the comparison produces a valid mask.
         let mask = unsafe {
-            let tfvec: Simd<<T as SimdElement>::Mask, LANES> = intrinsics::simd_ne(*self, *other);
+            let tfvec: Simd<<T as SimdElement>::Mask, N> = intrinsics::simd_ne(*self, *other);
             Mask::from_int_unchecked(tfvec)
         };
 
-        // Two vectors are non-equal if any lane tested true for vertical non-equality.
+        // Two vectors are non-equal if they are elementwise non-equal
```
programmerjake, line 699:
```suggestion
        // Two vectors are non-equal if any elements are non-equal when compared elementwise
```
`elementwise non-equal` implies to me producing a vector of results, not a scalar

crates/core_simd/src/vector.rs
```diff
@@ -651,20 +670,20 @@ where
...
     fn eq(&self, other: &Self) -> bool {
         // Safety: All SIMD vectors are SimdPartialEq, and the comparison produces a valid mask.
         let mask = unsafe {
-            let tfvec: Simd<<T as SimdElement>::Mask, LANES> = intrinsics::simd_eq(*self, *other);
+            let tfvec: Simd<<T as SimdElement>::Mask, N> = intrinsics::simd_eq(*self, *other);
             Mask::from_int_unchecked(tfvec)
         };
 
-        // Two vectors are equal if all lanes tested true for vertical equality.
+        // Two vectors are equal if they are elementwise equal
```
programmerjake, line 686:
```suggestion
        // Two vectors are equal if all elements are equal when compared elementwise
```
`elementwise equal` implies to me producing a vector of results, not a scalar

programmerjake (commented):
other than those minor issues, lgtm!

programmerjake (approved):

crates/core_simd/src/vector.rs
```diff
@@ -3,48 +3,66 @@ use crate::simd::{
...
 ///
-/// Two vectors of the same type and length will, by convention, support the operators (+, *, etc.) that `T` does.
-/// These take the lanes at each index on the left-hand side and right-hand side, perform the operation,
-/// and return the result in the same lane in a vector of equal size. For a given operator, this is equivalent to zipping
-/// the two arrays together and mapping the operator over each lane.
+/// `Simd<T, N>` supports the operators (+, *, etc.) that `T` does in "elementwise" fashion.
+/// These take the element at each index from the left-hand side and right-hand side,
+/// perform the operation, then return the result in the same index in a vector of equal size.
+/// In other words, an elementwise operation is equivalent to a zip, then map.
 ///
```
Lokathor, line 12: I'd add a single-line paragraph after this paragraph but before the code example that's something like

> The major difference between `Simd<T, N>` and `[T; N]` is that SIMD data generally has a higher required alignment.

That helps quickly answer the common question of "why are we making a separate type at all? why can't we just add this directly on arrays?"

programmerjake: may also want to mention that arrays and vectors are different types in most compilers (e.g. LLVM IR) and using array types often won't produce simd instructions.

scottmcm: "often won't" might overstate it  -- we have tests confirming that they often *will*

https://github.com/rust-lang/rust/blob/a73288371e3fa0a610fbc11e7e8418017bdfde42/tests/codegen/array-map.rs#L9-L27

programmerjake: well, i built a more complex test (simplified core of a 3D triangle rasterizer) and i'm rather impressed by how good llvm has gotten since I tried a similar algorithm with clang and llvm 4:
https://rust.godbolt.org/z/dqq164sbx

programmerjake: though it still needed some adjustment to make it vectorize (llvm won't convert a loop into a masked store), hence i think arrays still sometimes don't optimize as well as explicit simd ops.

scottmcm: Certainly, but that's not because of the *type*, that's because of the operations.  If one doesn't use a masked load operation with a SIMD type it's probably not going to optimize to one either.

crates/core_simd/src/vector.rs
```diff
@@ -3,48 +3,71 @@ use crate::simd::{
...
+///
+/// More complex elementwise operations may join vectors with masks which "enable" or "disable"
+/// appropriate elements. These often involve a `Simd<T, N>` where `T` is `usize`, `*const T`,
+/// or `*mut T`, and "enabling" means to use the index or pointer.
+/// The resulting operation is like a mutating iteration:
+///
+/// ```rust
+/// # #![feature(portable_simd)]
+/// # use core::simd::Simd;
+/// let source: Vec<i32> = (0..100).into_iter().collect();
```
workingjubilee, line 42: This example is _clearly_ unfinished, I am removing it for now since it's entirely greenfield.

PR #339, opened 2023-04-22
Don't use direct field access in `Simd` functions

Sp00ph:
See https://rust-lang.zulipchat.com/#narrow/stream/257879-project-portable-simd/topic/splat.20no.20longer.20compiles.20for.20release.20builds/near/351949106

### DIFF
--- a/crates/core_simd/src/lib.rs
+++ b/crates/core_simd/src/lib.rs
@@ -1,6 +1,8 @@
 #![no_std]
 #![feature(
     const_ptr_read,
+    const_refs_to_cell,
+    const_transmute_copy,
     convert_float_to_int,
     decl_macro,
     intra_doc_pointers,
--- a/crates/core_simd/src/lib.rs
+++ b/crates/core_simd/src/lib.rs
@@ -2,7 +2,6 @@
 #![feature(
     const_ptr_read,
     const_refs_to_cell,
-    const_transmute_copy,
     convert_float_to_int,
     decl_macro,
     intra_doc_pointers,
--- a/crates/core_simd/src/vector.rs
+++ b/crates/core_simd/src/vector.rs
@@ -135,22 +135,32 @@ where
     /// assert_eq!(v.as_array(), &[0, 1, 2, 3]);
     /// ```
     pub const fn as_array(&self) -> &[T; LANES] {
-        &self.0
+        // SAFETY: Transmuting between `Simd<T, LANES>` and `[T; LANES]`
+        // is always valid and `Simd<T, LANES>` never has a lower alignment
+        // than `[T; LANES]`.
+        unsafe { &*(self as *const Self as *const [T; LANES]) }
     }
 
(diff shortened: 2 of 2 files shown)


crates/core_simd/src/vector.rs
```diff
@@ -135,22 +135,26 @@ where
     /// assert_eq!(v.as_array(), &[0, 1, 2, 3]);
     /// ```
     pub const fn as_array(&self) -> &[T; LANES] {
-        &self.0
+        // SAFETY: `[T; LANES]` and `Simd<T, LANES>` have the same layout.
```
calebzulawski, line 138: Since layout includes alignment, I'm not sure this is quite accurate

crates/core_simd/src/lib.rs
```diff
@@ -1,10 +1,13 @@
 #![no_std]
 #![feature(
     const_ptr_read,
+    const_refs_to_cell,
+    const_transmute_copy,
     convert_float_to_int,
     decl_macro,
     intra_doc_pointers,
     platform_intrinsics,
+    ptr_from_ref,
```
calebzulawski, line 10: `core` doesn't appear to have this feature enabled, so I would use `as` instead

crates/core_simd/src/vector.rs
Sp00ph: Would something like "`Simd<T, LANES>` can always be transmuted to `[T; LANES]`" be fine with you as the safety comment?

calebzulawski: Yep, or maybe something like "`Simd<T, LANES>` is an over-aligned array"

calebzulawski (approved):

workingjubilee (changes requested):
Please document why we *have* to do it this way, so no one tries to be "helpful" and open a PR "fixing" this.

Sp00ph:
Where do you think would be the best place to put such a disclaimer? Should I just put it on the `Simd` struct directly, since it's not specific to the changed functions? Or maybe both on the struct and on the changed functions?

workingjubilee:
Let's do this belt-and-suspenders style, yeah. Double up. It can be terse but it should be in both places.

workingjubilee (approved):

Sp00ph:
I expanded the comment a bit to also include that you can't directly construct the type either, as I don't think it's obvious that that counts as a field access (and also had a typo in the previous commit message).

calebzulawski:
I think you can remove the transmute feature now

Sp00ph:
This is the second time in a row now that the `x86_64-apple-darwin` test failed because it wasn't able to reach `https://static.rust-lang.org/dist/channel-rust-nightly.toml.sha256`. Looks like something's wrong with that machine.

jhorstmann:
Sounds like this could already be the fix to what I just reported in rust-lang/rust#110722

PR #340, opened 2023-04-23
Syncing fixes

workingjubilee:
I accidentally over-replaced in my last merge. And there's a few other small compilation-breaking fixes that can be done before merging, so I will.

### DIFF
--- a/crates/core_simd/src/vector.rs
+++ b/crates/core_simd/src/vector.rs
@@ -2,6 +2,7 @@ use crate::simd::{
     intrinsics, LaneCount, Mask, MaskElement, SimdCast, SimdCastPtr, SimdConstPtr, SimdMutPtr,
     SimdPartialOrd, SupportedLaneCount, Swizzle,
 };
+use core::convert::{TryFrom, TryInto};
 
 /// A SIMD vector with the shape of `[T; N]` but the operations of `T`.
 ///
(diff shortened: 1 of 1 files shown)


crates/core_simd/src/vector.rs
```diff
@@ -863,7 +864,7 @@ where
 {
     type Error = core::array::TryFromSliceError;
 
-    fn try_from(slice: &mut [T]) -> Result<Self, Self::Error> {
+    fn try_from(slice: &mut [T]) -> Result<Self, core::array::TryFromSliceError> {
```
programmerjake, line 867: why not just leave it as `Self::Error` -- it's shorter and imho better (sorry icr the word I wanted insead of better -- basically imho more aligned with norms of rust code)

calebzulawski: idiomatic? :)

programmerjake: not what I was thinking of, but good enough!

workingjubilee: @programmerjake because `x check` screams at me when I do.

programmerjake: oh, ok....

programmerjake (approved):

workingjubilee:
GitHub Actions is being tetchy with not finishing out the tests, with a bunch of failures to run the test at all, but I'm passing the main repo's tests and that's good enough for me, honestly.

calebzulawski (approved):

Issue #341, opened 2023-04-23
don't use `read_unaligned` to read from an array casted to `*const Simd`


programmerjake:
We should create a `simd_read_array` intrinsic because for `sizeof(Simd<T, N>) > sizeof([T; N])` (which can happen until #319 is fixed) `read_unaligned` is probably UB due to being able to read the bytes beyond the end of the input array -- the padding in the `Simd<T, N>`.

We need an intrinsic rather than just using `memcpy` because the intrinsic will generate llvm's `load` instruction with vector type (llvm guarantees vector `load` won't read padding), whereas `memcpy` may end up using less efficient array-typed loads which sometimes use scalar code.

https://rust-lang.zulipchat.com/#narrow/stream/257879-project-portable-simd/topic/splat.20no.20longer.20compiles.20for.20release.20builds/near/352101044

scottmcm:
EDIT: this comment is no longer relevant as that copy-via-vectors work got removed.

I think that with https://github.com/rust-lang/rust/pull/111999 it might be fine to do this as an array copy?  This:
```rust
const N: usize = 16;
pub fn simd_from_array(a: [u32; N]) -> Simd<u32, N> {
    let mut v = Simd::default();
    *v.as_mut_array() = a;
    v
}
```
Generates a vector load/store now: <https://rust.godbolt.org/z/hYv4T17TP>
```llvm
define void @_ZN7example15simd_from_array17h5fc2848bbb0cc1d0E(ptr noalias nocapture noundef writeonly sret(<16 x i32>) align 64 dereferenceable(64) %_0, ptr noalias nocapture noundef readonly align 4 dereferenceable(64) %a) unnamed_addr #0 {
  %v.0.copyload = load <16 x i32>, ptr %a, align 4
  store <16 x i32> %v.0.copyload, ptr %_0, align 64
  ret void
}
```
(Though it'll fallback to a safe `memcpy` for non-pot sizes because <https://github.com/rust-lang/rust/pull/115236> found that non-pot vector loads & stores go very poorly on many platforms.)

scottmcm:
It looks like https://github.com/rust-lang/portable-simd/issues/319 is fixed; do we still need this?

[...]

PR #342, opened 2023-04-23
Fix {to,from}_array UB when repr(simd) produces padding

calebzulawski:
Addresses #341 (but doesn't introduce an intrinsic).

### DIFF
--- a/crates/core_simd/src/vector.rs
+++ b/crates/core_simd/src/vector.rs
@@ -184,12 +184,15 @@ where
     /// # Safety
     /// Reading `ptr` must be safe, as if by `<*const [T; N]>::read_unaligned`.
     const unsafe fn load(ptr: *const [T; N]) -> Self {
-        let mut tmp = core::mem::MaybeUninit::uninit();
+        // There are potentially simpler ways to write this function, but this should result in
+        // LLVM `load <N x T>`
+
+        let mut tmp = core::mem::MaybeUninit::<Self>::uninit();
         // SAFETY: `Simd<T, N>` always contains `N` elements of type `T`.  It may have padding
         // which does not need to be initialized.  The safety of reading `ptr` is ensured by the
         // caller.
         unsafe {
-            core::ptr::copy_nonoverlapping(ptr, tmp.as_mut_ptr() as *mut _, 1);
+            core::ptr::copy_nonoverlapping(ptr, tmp.as_mut_ptr().cast(), 1);
             tmp.assume_init()
         }
     }
@@ -201,9 +204,14 @@ where
     /// # Safety
     /// Writing to `ptr` must be safe, as if by `<*mut [T; N]>::write_unaligned`.
     const unsafe fn store(self, ptr: *mut [T; N]) {
+        // There are potentially simpler ways to write this function, but this should result in
+        // LLVM `store <N x T>`
+
+        // Creating a temporary helps LLVM turn the memcpy into a store.
+        let tmp = self;
         // SAFETY: `Simd<T, N>` always contains `N` elements of type `T`.  The safety of writing
         // `ptr` is ensured by the caller.
-        unsafe { core::ptr::copy_nonoverlapping(self.as_array(), ptr, 1) }
+        unsafe { core::ptr::copy_nonoverlapping(tmp.as_array(), ptr, 1) }
     }
 
     /// Converts an array to a SIMD vector.
--- a/crates/core_simd/src/vector.rs
+++ b/crates/core_simd/src/vector.rs
@@ -176,34 +176,62 @@ where
         unsafe { &mut *(self as *mut Self as *mut [T; N]) }
     }
 
+    /// Load a vector from an array of `T`.
+    ///
+    /// This function is necessary since `repr(simd)` has padding for non-power-of-2 vectors (at the time of writing).
+    /// With padding, `read_unaligned` will read past the end of an array of N elements.
+    ///
+    /// # Safety
+    /// Reading `ptr` must be safe, as if by `<*const [T; N]>::read_unaligned`.
+    const unsafe fn load(ptr: *const [T; N]) -> Self {
+        let mut tmp = core::mem::MaybeUninit::uninit();
+        // SAFETY: `Simd<T, N>` always contains `N` elements of type `T`.  It may have padding
+        // which does not need to be initialized.  The safety of reading `ptr` is ensured by the
+        // caller.
+        unsafe {
+            core::ptr::copy_nonoverlapping(ptr, tmp.as_mut_ptr() as *mut _, 1);
+            tmp.assume_init()
+        }
+    }
+
+    /// Store a vector to an array of `T`.
+    ///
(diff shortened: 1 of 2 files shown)


crates/core_simd/src/vector.rs
```diff
@@ -176,34 +176,62 @@ where
...
-        // is always valid. No need to use `read_unaligned` here, since
-        // the vector never has a lower alignment than the array.
+        let mut tmp = core::mem::MaybeUninit::uninit();
+        // SAFETY: writing to `tmp` is safe and initializes it.
         //
-        // FIXME: We currently use a pointer read instead of `transmute_copy` because
-        // it results in better codegen with optimizations disabled, but we should
-        // probably just use `transmute` once that works on const generic types.
+        // FIXME: We currently use a pointer store instead of `transmute_copy` because `repr(simd)`
+        // results in padding for non-power-of-2 vectors (so vectors are larger than arrays).
```
scottmcm, line 227: If vectors are larger than arrays and more aligned (or the same size and same align), then I think the `read` is always fine for this?

Because if not, we'll have to remove <https://doc.rust-lang.org/std/simd/struct.Simd.html#method.as_mut_array>.

(Which, comes to think of it, means that we can always implement `to_array` as `*self.as_array()`, since we currently require `T: Copy`.)

crates/core_simd/src/vector.rs
```diff
@@ -176,34 +176,62 @@ where
...
+    ///
+    /// # Safety
+    /// Reading `ptr` must be safe, as if by `<*const [T; N]>::read_unaligned`.
+    const unsafe fn load(ptr: *const [T; N]) -> Self {
+        let mut tmp = core::mem::MaybeUninit::uninit();
+        // SAFETY: `Simd<T, N>` always contains `N` elements of type `T`.  It may have padding
+        // which does not need to be initialized.  The safety of reading `ptr` is ensured by the
+        // caller.
+        unsafe {
+            core::ptr::copy_nonoverlapping(ptr, tmp.as_mut_ptr() as *mut _, 1);
```
scottmcm, line 192: ymmv nit: If you're going to `as` with an `_`, might as well use `cast`:
```suggestion
            core::ptr::copy_nonoverlapping(ptr, tmp.as_mut_ptr().cast(), 1);
```

calebzulawski: This needs `::<Self>` on `MaybeUninit` to avoid a type inference issue:
```
error[E0699]: the type of this value must be known to call a method on a raw pointer on it
   --> crates/core_simd/src/vector.rs:192:66
    |
192 |             core::ptr::copy_nonoverlapping(ptr, tmp.as_mut_ptr().cast(), 1);
    |
    ```

File: crates/core_simd/src/vector.rs (hunk 1 above)
Comment (calebzulawski) [reply crates/core_simd/src/vector.rs]: Good point, I changed it to `*self.as_array()`

File: crates/core_simd/src/vector.rs (hunk 2 above)
Comment (scottmcm) [reply crates/core_simd/src/vector.rs]: Ah, I see.

(TBH, I like the `::<Self>` since in unsafe code my usual "it's fine, if it infers weird rust will tell me" is less reliable.)

**Review by scottmcm — APPROVED** [review]
I'm not a reviewer here, but FWIW this looks good to me.

File: crates/core_simd/src/vector.rs (hunk 2 above)
Comment (calebzulawski) [reply crates/core_simd/src/vector.rs]: Agreed, I thought it was more clear in the end. I always forget about `cast` too...

File: crates/core_simd/src/vector.rs [hunk 3]
```diff
@@ -176,34 +176,62 @@ where
...
+        }
+    }
+
+    /// Store a vector to an array of `T`.
+    ///
+    /// See `load` as to why this function is necessary.
+    ///
+    /// # Safety
+    /// Writing to `ptr` must be safe, as if by `<*mut [T; N]>::write_unaligned`.
+    const unsafe fn store(self, ptr: *mut [T; N]) {
```
programmerjake: I think you'll want to explicitly copy `self` to a temporary first, and then memcpy from the temporary, since afaict llvm optimizes that to a vector load/store due to the vector insns generated by the temporary, whereas just calling memcpy generates only a memcpy so llvm doesn't see any vector operations and generates possibly less-efficient code.

compare `store` vs. `store2` -- `store2` has the temporary.
https://godbolt.org/z/jWsc6ezP4

crates/core_simd/src/vector.rs
```diff
@@ -220,18 +208,7 @@ where
...
-        // FIXME: We currently use a pointer store instead of `transmute_copy` because `repr(simd)`
-        // results in padding for non-power-of-2 vectors (so vectors are larger than arrays).
-        //
-        // NOTE: This deliberately doesn't just use `self.0`, see the comment
-        // on the struct definition for details.
-        unsafe {
-            self.store(tmp.as_mut_ptr());
-            tmp.assume_init()
-        }
+        *self.as_array()
```
programmerjake, line 211: this doesn't generate a vector `store` instruction so llvm may produce less optimal code.

crates/core_simd/src/vector.rs
scottmcm: Isn't it *intentional*, though, that it's not a vector read?  Because I read the OP as saying that that would generate an over-long read.

programmerjake: no, because LLVM IR defines vector `load`/`store` instructions to never read/write the padding, as i explained on Zulip. `read_unaligned` has no such guarantee

programmerjake: so we *intentionally* want a llvm `load`/`store` instruction with `align <elem-align>` not full-simd-type-align

calebzulawski: It's not written as a vector store, because that would be wrong (writing past the end of the array), but it should lower as one.  I added the temporary with a comment as to why it's necessary.

programmerjake (approved):

crates/core_simd/src/vector.rs
```diff
@@ -176,34 +176,70 @@ where
...
+    /// See `load` as to why this function is necessary.
+    ///
+    /// # Safety
+    /// Writing to `ptr` must be safe, as if by `<*mut [T; N]>::write_unaligned`.
+    const unsafe fn store(self, ptr: *mut [T; N]) {
+        // There are potentially simpler ways to write this function, but this should result in
+        // LLVM `store <N x T>`
+
+        // Creating a temporary helps LLVM turn the memcpy into a store.
+        let tmp = self;
```
programmerjake, line 211: once this is merged into the main `rust` repo, we'll probably want to add codegen tests to ensure `load`/`store` actually optimize to a llvm vector `load`/`store` instruction.

programmerjake: created #343

crates/core_simd/src/vector.rs
scottmcm: Oh, I see, I was mixing up the LLVM type and the Rust type.

PR #347, opened 2023-05-07
Add `#[inline]` to functions which were missing it, and `#[track_caller]` to ones with runtime panics from user input

thomcc:
Rationale for inlining is basically twofold. First: even though `#[inline]` doesn't help the LLVM inliner for generics, it apparently can help the MIR inliner (and the lack of it has been commented on before by @saethlin, at least).

Second, and perhaps more controversially, this library depends on inlining in a way that basically no other part of the stdlib does -- failing to inline most of these functions kind of defeats the point of the design we've chosen.

Concretely, the `simd_foo` intrinsics this library will codegen to take advantage of whatever vector operations are allowed in the function where they finally get codegenned (e.g. after all inlining occurs).

This means that if we define our high-level wrappers for some `simd_foo` intrinsic and the wrapper gets called from a downstream users `#[target_feature(enable="foo")]` function (for a relevant feature), then failing to inline that wrapper into the call-site is squandering the approach we've gained with by using the `simd_foo` intrinsic design.

Arguably, this justifies use of `#[inline(always)]`. I am in favor of that (it would also could help `core::simd`-using code have improved performance in unoptimized builds compared to `core::arch` use (which is absurdly slow without opts). That said, I didn't bother since that can be done later, and might be better as just some extra tuning on the MIR inliner. Not to mention, we support simd ops on unreasonably huge vectors and it seems completely unreasonable to use `#[inline(always)]` on functions of `Simd<f64, 4096>` or whatever.

And I also added a clippy warning for missing inline on public functions (although private functions that are called from public functions should get ideally get `#[inline]` too in most cases).

---

The rationale for `track_caller` should be obvious (much easier for users to track down the issue). While track_caller can have a very small perf impact, it does not matter here (the overhead it adds is one extra pointer argument passed into the function, but we expect these to be inlined, so that argument doesn't need to be passed).

### DIFF
--- a/crates/core_simd/src/vector.rs
+++ b/crates/core_simd/src/vector.rs
@@ -266,6 +266,7 @@ where
     /// let v = u32x4::from_slice(&source);
     /// assert_eq!(v.as_array(), &[1, 2, 3, 4]);
     /// ```
+    #[must_use]
     #[inline]
     #[track_caller]
     pub const fn from_slice(slice: &[T]) -> Self {
--- a/crates/core_simd/src/vector.rs
+++ b/crates/core_simd/src/vector.rs
@@ -122,6 +122,7 @@ where
     /// let v = u32x4::splat(0);
     /// assert_eq!(v.lanes(), 4);
     /// ```
+    #[inline]
     pub const fn lanes(&self) -> usize {
         Self::LANES
     }
@@ -136,6 +137,7 @@ where
     /// let v = u32x4::splat(8);
     /// assert_eq!(v.as_array(), &[8, 8, 8, 8]);
     /// ```
+    #[inline]
     pub fn splat(value: T) -> Self {
         // This is preferred over `[value; N]`, since it's explicitly a splat:
         // https://github.com/rust-lang/rust/issues/97804
@@ -156,6 +158,7 @@ where
     /// let v: u64x4 = Simd::from_array([0, 1, 2, 3]);
     /// assert_eq!(v.as_array(), &[0, 1, 2, 3]);
     /// ```
+    #[inline]
     pub const fn as_array(&self) -> &[T; N] {
         // SAFETY: `Simd<T, N>` is just an overaligned `[T; N]` with
         // potential padding at the end, so pointer casting to a
@@ -167,6 +170,7 @@ where
     }
 
     /// Returns a mutable array reference containing the entire SIMD vector.
+    #[inline]
     pub fn as_mut_array(&mut self) -> &mut [T; N] {
         // SAFETY: `Simd<T, N>` is just an overaligned `[T; N]` with
         // potential padding at the end, so pointer casting to a
@@ -184,6 +188,7 @@ where
     ///
     /// # Safety
     /// Reading `ptr` must be safe, as if by `<*const [T; N]>::read_unaligned`.
+    #[inline]
     const unsafe fn load(ptr: *const [T; N]) -> Self {
         // There are potentially simpler ways to write this function, but this should result in
         // LLVM `load <N x T>`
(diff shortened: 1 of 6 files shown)


saethlin:
FYI since https://github.com/rust-lang/rust/pull/109247, MIR inlining is _permitted_ without `#[inline]`, but the attribute still doubles(!) the inline cost threshold. The MIR inliner also now ignores the size of locals (https://github.com/rust-lang/rust/pull/110705), so `Simd<f64, 4096>` does not become a bad inlining candidate on account of its size. It will still be a bad inlining candidate if its size results in a lot of statements.

thomcc:
Yeah that makes sense. Even if it's not a hard block for the MIR inliner now, these are things that *really really* should be inlined.

calebzulawski (approved):

calebzulawski:
I think this is uncontroversial and mostly oversights. That lint is helpful

andy-thomason:
Very sensible. It is a shame that rust does not inline methods by default like C++ does.

saethlin:
> It is a shame that rust does not inline methods by default like C++ does.

By default, Rust has exactly the same inlining behaviors as C++ does. Removing indirect calls from final codegen is not what is being discussed here.

crates/core_simd/src/vector.rs
```diff
@@ -257,7 +265,8 @@ where
     /// let v = u32x4::from_slice(&source);
     /// assert_eq!(v.as_array(), &[1, 2, 3, 4]);
     /// ```
-    #[must_use]
```
programmerjake, line 260: why delete `#[must_use]`? imho it's appropriate here...

andy-thomason:
> By default, Rust has exactly the same inlining behaviors as C++ does.

Experienced C++ users will generally build the entire project as a single compilation unit
which makes significant improvements to build time and code performance by enabling
IR reductions early in the lowering process. For example, I did a little example for the LLVM
group where I built LLVM from scratch nearly a thousand times faster than the default
build as less than ten CUs. The resulting code was also faster and smaller. With modern
computers, you could build LLVM in a single CU.

If we do a default build with Rust (i.e. large numbers of CUs), we get the same slow compilation time
and loss of performance from both lack of inlining and the inability of the compiler to do global optimisation
by introspection.

Of course inline semantics are more than just a hint to the compiler. In C++ they promote functions to
their own sections and enable the linker to remove duplicates.

I know that calling on all functions to be inlinable will fall on deaf ears, but it would be a simple experiment
to try. You never know, you might like the result.

crates/core_simd/src/vector.rs
thomcc: Wasn't intentional.

bjorn3:
In C++ I think it is more that building as a single unit allows not parsing and typechecking headers again for every source file. Rustc already doesn't do this. As for the backend side, rustc in fact intentionally splits a single crate into multiple codegen units to improve compilation time by allowing parallelism. Cross crate inlining without `#[inline]` is supported to through LTO, but this is disabled by default because it is much slower. There is fat LTO which is effectively merging all codegen units of all crates into a single module and then optimizing it as a whole, which is really slow. And there is thin LTO, where only a summary is extracted from each module before optimizing each codegen unit in parallel using the combination of all summaries. This is much faster, but still slower than not doing any LTO. For release builds we default to thin local LTO which is a variant on thin LTO where only the codegen units within a single crate are taken into account. In the past we used a single codegen unit in release mode, but thin local LTO is much faster while barely regressing performance.

> Of course inline semantics are more than just a hint to the compiler. In C++ they promote functions to
their own sections and enable the linker to remove duplicates.

In rust we put every function in their own section and tell LLVM that it is allowed to merge functions. Unless you enable ICF for the linker this doesn't allow merging identical functions on the linker side, but LLVM does merge functions within a codegen unit and the linker will remove all unused functions because we use `--gc-sections`.

PR #353, opened 2023-06-07
Sync upstream

calebzulawski:
Reverse sync from rust-lang/rust


calebzulawski:
@workingjubilee have you ever done this reverse sync?

calebzulawski:
I added a message whenever a PR is opened against our code in rust-lang/rust, hopefully some of these smaller changes will happen here in the future.

programmerjake (approved):

workingjubilee:
I have, which was why I hadn't opened this PR yet. I was planning to investigate why the reverse sync brought in old changes that should have already been synced, and if it would be prevented in the future.

calebzulawski:
Ok, up to you if I abandon this, I'm guessing it has something to do with commit order.  I wanted to give it a try so I know how to do it in the future.

workingjubilee:
I'm quite appreciative that you did it, tbh! I will try to figure out swiftly if there is something amiss with the subtree sync that means we shouldn't land this. Probably not, we probably should, just would be nice to remove all doubt.

workingjubilee:
I finally ran the multi-iteration shell-game to confirm this will work, sorry for the delay!

PR #354, opened 2023-07-07
Disable MIPS jobs in CI

workingjubilee:
We can't simply expect these components available on nightly, currently.


workingjubilee:
Grah, sorry.

PR #359, opened 2023-07-27
Define portability

### LINKED ISSUE #227: Define "portable" and elaborate design constraints
While we have discussed what "portable" means in the past and have something of an image of that, at least partly based on our preexisting design constraints informing us heavily on what "portability" means, it seems prudent to define it in as formal-ish a sense as possible... though somewhat more like elaborating tradeoffs. And where SIMD is concerned, "portability" also means "with performance".

This has been directly raised (and thus relevant side-conversation also appears in) at least these issues:
- https://github.com/rust-lang/portable-simd/issues/225

It also is arguably the source of the entire now-infamous mask operation design challenge. This will be essential to our RFC, but is also important for user concerns, since some will have specific ideas already in mind.

Excerpted from [Zulip conversation](https://rust-lang.zulipchat.com/#narrow/stream/257879-project-portable-simd):
- workingjubilee:
  > I think [portable-simd#225](https://github.com/rust-lang/portable-simd/issues/225) has posed a fair question in that we want to like, kinda semi-formally define what we mean by "portable".
  > Or at least, explain all our constraints in a as concise a manner as possible.
- calebzulawski:
  > Agreed.
  > We should probably formalize it somewhere, though I assume our answer is going to be something like "can be implemented reasonably on nearly all architectures and offers good performance for that particular operation as far as that architecture goes"
  > I think the point you made nails it, that just because there's an instruction it's not necessarily fast anyway
- programmerjake: 
  > i'd pick something like "provide all reasonable primitive operations that a simd programmer would want in a mostly architecture independent way"
- calebzulawski:
  > Yep. I think we should also specify that for some architectures there might be entirely different ways to organize algorithms that avoid particular operations, and that's considered an optimization, not a flaw of portability
  > "it will do exactly what you tell it, as best as it can"
- miguelraz:
  > So, a compiler.

calebzulawski:
It's been a while, so I think we understand the constraints enough to answer this.  Closes #227.

### DIFF
--- a/crates/core_simd/src/core_simd_docs.md
+++ b/crates/core_simd/src/core_simd_docs.md
@@ -2,3 +2,36 @@ Portable SIMD module.
 
 This module offers a portable abstraction for SIMD operations
 that is not bound to any particular hardware architecture.
+
+# What is "portable"?
+
+This module provides a SIMD implementation that is fast and predictable on any target.
+
+### Portable SIMD works on every target
+
+Unlike target-specific SIMD in `std::arch`, portable SIMD compiles for every target.
+In this regard, it is just like "regular" Rust.
+
+### Portable SIMD is consistent between targets
+
+A program using portable SIMD can expect identical behavior on any target.
+In most regards, [`Simd<T, N>`] can be thought of as a parallelized `[T; N]` and operates like a sequence of `T`.
+
+This has one notable exception: a handful of older architectures (e.g. `armv7` and `powerpc`) flush [subnormal](`f32::is_subnormal`) `f32` values to zero.
+On these architectures, subnormal `f32` input values are replaced with zeros, and any operation producing subnormal `f32` values produces zeros instead.
+This doesn't affect most architectures or programs.
+
+### Operations use the best instructions available
+
+Operations provided by this module compile to the best available SIMD instructions.
+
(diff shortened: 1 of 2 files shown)


crates/core_simd/src/core_simd_docs.md
```diff
@@ -2,3 +2,36 @@ Portable SIMD module.
...
+### Operations use the best instructions available
+
+Operations provided by this module compile to the best available SIMD instructions.
+
+Portable SIMD is not a low-level vendor library, and operations in portable SIMD _do not_ necessarily map to a single instruction.
+Instead, they map to a reasonable implementation of the operation for the target.
+
+Consistency between targets is not compromised to use faster or fewer instructions.
+In some cases, `std::arch` will provide a faster function that has slightly different behavior than the `std::simd` equivalent.
+For example, [`_mm_min_ps`](`core::arch::x86_64::_mm_min_ps`) can be slightly faster than [`SimdFloat::simd_min`], but does not conform to the IEEE standard also used by [`f32::min`].
```
programmerjake, line 33: maybe add a footnote that you can generate `minps` with portable-simd by using `a.simd_lt(b).select(a, b)` and similarly for `maxps`

programmerjake (approved):

PR #360, opened 2023-07-27
Add scalar shifts

### LINKED ISSUE #225: Shl and Shr implementations should accept a (polymorphic?) scalar argument rather than a vector
The operation "shift with a different offset in each lane" is not portable: it's supported by NEON, unsupported by SSE 4.1 and Wasm, and only partially-supported by AVX2. It would probably make more sense for types like `Simd<i32, LANES>` to implement `Shl<i32>`, rather than `Shl<Simd<i32, LANES>>`.

Scalar integer types currently implement `Shl` and `Shr` for [all other integer types](https://doc.rust-lang.org/std/ops/trait.Shl.html#implementors), so that might be worth considering. One downside is that it would make the `struct Simd` rustdoc page even longer.

calebzulawski:
I think shifts are different than `Add`, `Mul`, etc, and warrant these auto-splatting trait impls.  Unlike arithmetic, it's not clear to users that the correct LLVM IR is a splatted shift, and they expect something that looks more like a unary op.  Closes #225.

### DIFF
--- a/crates/core_simd/src/ops.rs
+++ b/crates/core_simd/src/ops.rs
@@ -6,6 +6,7 @@ use core::ops::{Shl, Shr};
 
 mod assign;
 mod deref;
+mod shift_scalar;
 mod unary;
 
 impl<I, T, const LANES: usize> core::ops::Index<I> for Simd<T, LANES>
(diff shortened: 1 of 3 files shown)


programmerjake (approved):

programmerjake:
maybe add a smoke test?

programmerjake (approved):

thomcc (approved):
Yes, I agree that shifts are fundamentally different than add/mul in this respect.

workingjubilee:
I still want all of them tbh! I agree these are less likely to break stuff due to bad type inference though, since shifts are already weird about the shamt operand.

PR #361, opened 2023-07-27
Add wrapping negation for unsigned integers

### LINKED ISSUE #330: Missing Neg for unsigned types
`Simd` misses impls of `Neg` for unsigned integer types.

Since `Simd` uses wrapping arithmetic, negation for unsigned values has a good definition: `N - val % N` where `N = MAX + 1`. See e.g. [`u8::wrapping_neg`](https://doc.rust-lang.org/stable/std/primitive.u8.html#method.wrapping_neg).

And, yes, I want to use this (calculation of rejection threshold for Lemire's method of sampling unbiased integers uniformly from a range).

calebzulawski:
A compromise that closes #330.

One could argue that since `Simd` is always wrapping, it should implement `Neg`, however this is potentially confusing when this operation effectively always wraps.  It could also be argued that regular unsigned integers wrap in release mode--`Simd` is simply always in release mode :)

A compromise is to add an explicit `wrapping_neg` functions.

### DIFF
--- a/crates/core_simd/src/elements/uint.rs
+++ b/crates/core_simd/src/elements/uint.rs
@@ -16,6 +16,12 @@ pub trait SimdUint: Copy + Sealed {
     #[must_use]
     fn cast<T: SimdCast>(self) -> Self::Cast<T>;
 
+    /// Wrapping negation.
+    ///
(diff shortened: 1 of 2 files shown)


programmerjake (approved):
`wrapping_neg` seems fine to me, replicating the unsigned integers' API where it makes sense is a good idea.

PR #362, opened 2023-07-30
Add various integer ops

calebzulawski:
~Pending rust-lang/rust#114266 and an `i16xN::trailing_zeros` bug that I hope is fixed by LLVM 15 17~

Good to go.

### DIFF
--- a/crates/core_simd/src/elements/int.rs
+++ b/crates/core_simd/src/elements/int.rs
@@ -213,7 +213,7 @@ pub trait SimdInt: Copy + Sealed {
 }
 
 macro_rules! impl_trait {
-    { $($ty:ident ($unsigned:ident)),* } => {
+    { $($ty:ty),* } => {
         $(
         impl<const LANES: usize> Sealed for Simd<$ty, LANES>
         where
@@ -353,18 +353,16 @@ macro_rules! impl_trait {
 
             #[inline]
             fn leading_ones(self) -> Self {
-                use crate::simd::SimdUint;
-                self.cast::<$unsigned>().leading_ones().cast()
+                (!self).leading_zeros()
             }
 
             #[inline]
             fn trailing_ones(self) -> Self {
-                use crate::simd::SimdUint;
-                self.cast::<$unsigned>().trailing_ones().cast()
+                (!self).trailing_zeros()
             }
         }
         )*
     }
 }
 
-impl_trait! { i8 (u8), i16 (u16), i32 (u32), i64 (u64), isize (usize) }
+impl_trait! { i8, i16, i32, i64, isize }
--- a/crates/core_simd/src/elements/int.rs
+++ b/crates/core_simd/src/elements/int.rs
@@ -191,10 +191,29 @@ pub trait SimdInt: Copy + Sealed {
 
     /// Returns the cumulative bitwise "xor" across the lanes of the vector.
     fn reduce_xor(self) -> Self::Scalar;
+
+    /// Reverses the byte order of each element.
+    fn swap_bytes(self) -> Self;
+
+    /// Reverses the order of bits in each elemnent.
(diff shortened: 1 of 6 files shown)


workingjubilee:
LLVM 15 is already our minimum.

calebzulawski:
Whoops, I meant 17

workingjubilee:
can you describe the bug?

calebzulawski:
I haven't looked at the codegen yet, but it reports that 4095 has 3 trailing zeros

calebzulawski:
Aha, actually it's a bug in my test (I compare against `leading_zeros`). I'm disturbed that it only failed for `i16`.

crates/core_simd/src/elements/int.rs
```diff
@@ -307,9 +326,45 @@ macro_rules! impl_trait {
...
+            #[inline]
+            fn leading_ones(self) -> Self {
+                use crate::simd::SimdUint;
+                self.cast::<$unsigned>().leading_ones().cast()
+            }
+
+            #[inline]
+            fn trailing_ones(self) -> Self {
+                use crate::simd::SimdUint;
+                self.cast::<$unsigned>().trailing_ones().cast()
```
programmerjake, line 363: why not just use:
```
(!self).trailing_zeros()
```

sammysheep:
Why is it called `swap_bytes` and not `rev_bytes` or some such thing? Say for `u32` elements, is it swapping the high and low 16 bits, or are you reversing the byte order over all 4 bytes?

Sorry if this was stated elsewhere.

calebzulawski:
That is the name used by scalars for the same operation: https://doc.rust-lang.org/std/primitive.i32.html#method.swap_bytes

programmerjake (approved):

crates/core_simd/src/elements/int.rs
```diff
@@ -191,6 +191,25 @@ pub trait SimdInt: Copy + Sealed {
...
+
+    /// Reverses the byte order of each element.
+    fn swap_bytes(self) -> Self;
+
+    /// Reverses the order of bits in each elemnent.
+    /// The least significant bit becomes the most significant bit, second least-significant bit becomes second most-significant bit, etc.
+    fn reverse_bits(self) -> Self;
+
+    /// Returns the number of leading zeros in the binary representation of each element.
+    fn leading_zeros(self) -> Self;
```
calebzulawski, line 203: Do we think it makes sense for these functions to return `Self` (a signed int) or should it have a more complex signature and return the equivalent unsigned type?

programmerjake: which exact functions are you referring to? for `leading_zeros`, any integer type with >= 3 bits will never wrap no matter which signedness, so we have no problem until we get generic-integer element types (if that ever happens...). I think `reverse_bits` should return `Self` for consistency with `i32::reverse_bits`

programmerjake: for `leading_zeros` it might be nice to return the corresponding unsigned type with the same bit-width, since that's consistent with the integer types, and also since that works for all bit lengths including 0, 1, and 2-bits.

calebzulawski: I just mean the counting functions, like `leading_zeros`.  I agree that `Self` will always fit the result, I meant more if it's inconsistent (with scalars) to return a signed type, instead of the unsigned int with the same size.  This is definitely the simpler option

programmerjake: ah, ok. well:
> for leading_zeros it might be nice to return the corresponding unsigned type with the same bit-width, since that's consistent with the integer types, and also since that works for all bit lengths including 0, 1, and 2-bits.

plus the bit-counting functions are imo better thought of as unsigned rather than signed integer operations

PR #363, opened 2023-08-26
Commit Cargo.lock

bjorn3:
As of rust-lang/cargo#8728 it is now recommended to always check in Cargo.lock. This will help with reproducability and will avoid the need for cg_clif's build system to keep it's own copy of Cargo.lock for vendoring. It will also allow tidy to run on the portable-simd workspace.


bjorn3:
CI failure seems unrelated to this PR:

```
   = note: /usr/bin/ld: /rust/lib/rustlib/x86_64-unknown-linux-gnu/lib/libstd-19964a87984cce8a.rlib(std-19964a87984cce8a.std.f6dab3b182b9be26-cgu.0.rcgu.o): TLS transition from R_X86_64_TLSLD to R_X86_64_TPOFF32 against `_ZN3std9panicking11panic_count17LOCAL_PANIC_COUNT7__getit3VAL17h178ac08236b8843dE.0' at 0x5af in section `.text._ZN3std2rt19lang_start_internal17h1dd1442605b126fbE' failed
          /rust/lib/rustlib/x86_64-unknown-linux-gnu/lib/libstd-19964a87984cce8a.rlib: error adding symbols: Bad value
          collect2: error: ld returned 1 exit status
```

calebzulawski:
It's only `cross` jobs, so I'm not concerned at all

PR #367, opened 2023-10-01
Simplify Swizzle trait

calebzulawski:
This is a possible answer to the questions about swizzles raised in #364.  This API isn't necessary *perfect*, but it's now a single unobtrusive trait that's relatively easy to use.

* `Swizzle` is no longer parameterized by the input length
* replaced `Swizzle2` and `Which` by `Swizzle::concat_swizzle`
* added mask swizzles to resolve #268 
* replaced various supporting traits and functions with inline `const {}` blocks

PR #368, opened 2023-10-02
Remove `generic_const_exprs`

calebzulawski:
Inspired by the errors in #367 (caused by rust-lang/rust#116320) I've removed all instances of `generic_const_exprs`.

For `ToBitMaskArray`, this was straightforward, simply use an associated type instead of an associated const.  For the various `to_bytes` functions, this was more complicated.  I now implemented a `ToBytes` trait over every individual vector type.  In the future when const generics are more powerful, I'd hope this can be implemented generically.

This effectively upgrades these functions to nightly, as they're currently disabled on core.

### DIFF
--- a/crates/core_simd/src/masks/to_bitmask.rs
+++ b/crates/core_simd/src/masks/to_bitmask.rs
@@ -30,19 +30,18 @@ pub trait ToBitMask: Sealed {
 /// Converts masks to and from byte array bitmasks.
 ///
 /// Each bit of the bitmask corresponds to a mask lane, starting with the LSB of the first byte.
-#[cfg(feature = "generic_const_exprs")]
 pub trait ToBitMaskArray: Sealed {
-    /// The length of the bitmask array.
-    const BYTES: usize;
+    /// The bitmask array.
+    type BitMaskArray;
 
     /// Converts a mask to a bitmask.
-    fn to_bitmask_array(self) -> [u8; Self::BYTES];
+    fn to_bitmask_array(self) -> Self::BitMaskArray;
 
     /// Converts a bitmask to a mask.
-    fn from_bitmask_array(bitmask: [u8; Self::BYTES]) -> Self;
+    fn from_bitmask_array(bitmask: Self::BitMaskArray) -> Self;
 }
 
-macro_rules! impl_integer_intrinsic {
+macro_rules! impl_integer {
     { $(impl ToBitMask<BitMask=$int:ty> for Mask<_, $lanes:literal>)* } => {
         $(
         impl<T: MaskElement> ToBitMask for Mask<T, $lanes> {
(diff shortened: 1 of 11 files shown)


crates/core_simd/src/masks/to_bitmask.rs
```diff
@@ -30,19 +30,18 @@ pub trait ToBitMask: Sealed {
 /// Converts masks to and from byte array bitmasks.
 ///
 /// Each bit of the bitmask corresponds to a mask lane, starting with the LSB of the first byte.
-#[cfg(feature = "generic_const_exprs")]
 pub trait ToBitMaskArray: Sealed {
-    /// The length of the bitmask array.
-    const BYTES: usize;
+    /// The bitmask array.
+    type BitMaskArray;
```
programmerjake, line 35: this should have some useful bounds, such as `Copy + Unpin + Send + Sync + 'static + AsRef<[u8]> + AsMut<[u8]> + BorrowMut<[u8]>` and maybe more

crates/core_simd/src/to_bytes.rs
```diff
@@ -1,72 +1,126 @@
...
+    use super::*;
+    pub trait Sealed {}
+    impl<T: SimdElement, const N: usize> Sealed for Simd<T, N> where LaneCount<N>: SupportedLaneCount {}
+}
+use sealed::Sealed;
+
+/// Convert SIMD vectors to vectors of bytes
+pub trait ToBytes: Sealed {
+    /// This type, reinterpreted as bytes.
+    type Bytes;
```
programmerjake, line 13: could use some bounds here too.

crates/core_simd/src/to_bytes.rs
```diff
@@ -1,72 +1,126 @@
...
+        impl_to_bytes! { $ty, $size * 8 }
+        impl_to_bytes! { $ty, $size * 16 }
+        impl_to_bytes! { $ty, $size * 32 }
+        impl_to_bytes! { $ty, $size * 64 }
+    };
+
+    // multiply element size by number of elements
+    { $ty:tt, 1 * $elems:literal } => { impl_to_bytes! { @impl [$ty; $elems], $elems } };
+    { $ty:tt, $size:literal * 1 } => { impl_to_bytes! { @impl [$ty; 1], $size } };
+    { $ty:tt, 2 * 2  } => { impl_to_bytes! { @impl [$ty; 2], 4  } };
```
programmerjake, line 58: you don't have to do these multiplications by hand, rustc works just fine with e.g. `AType<{4 * 8}>`, all it needs is that the expression not contain any const generic arguments.

calebzulawski: Ah, you're right. Didn't know you could do that

crates/core_simd/src/masks/to_bitmask.rs
calebzulawski: Would adding more bounds in the future be backwards compatible?

workingjubilee: No, bounds can only be relaxed forward-compatibly, and only sometimes.

calebzulawski: Is that still true when the trait is sealed?

workingjubilee: ...that does change things a bit, since the major compat hazard is people referencing the type generically.

programmerjake (approved):

PR #370, opened 2023-10-14
Eliminate use of `#[cfg_attr(not(doc), repr(transparent))]`

dtolnay:
Context: https://github.com/rust-lang/rust/pull/116743. This PR is extracted from there by advice of https://github.com/rust-lang/rust/pull/116743#issuecomment-1763178813.


calebzulawski:
Thank you for bringing the PR here 🙏

PR #371, opened 2023-10-21
Rename "lane" to "element"

calebzulawski:
In an attempt to use uniform language, I've changed most instances of "lane" to "element"

This is a huge PR, but to sum it up, there are some interface changes:
* Rename `Simd::LANES` to `Simd::LEN`
* Rename `Simd::lanes()` to `Simd::len()`
* Rename `Simd::rotate_lanes_{left,right}` to `Simd::rotate_elements_{left,right}`

Additionally, I updated documentation, and changed all generic parameters `LANES` to `N`.

I didn't touch `LaneCount` and `SupportedLaneCount`, but I figure those should be changed to be consistent.  `SimdLen` and `SupportedSimdLen`?

### DIFF
--- a/crates/core_simd/src/masks.rs
+++ b/crates/core_simd/src/masks.rs
@@ -1,4 +1,4 @@
-//! Types and traits associated with masking lanes of vectors.
+//! Types and traits associated with masking elements of vectors.
 //! Types representing
 #![allow(non_camel_case_types)]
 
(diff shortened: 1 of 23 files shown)


crates/core_simd/src/masks.rs
```diff
@@ -189,75 +188,75 @@ where
...
+    pub fn cast<U: MaskElement>(self) -> Mask<U, N> {
         Mask(self.0.convert())
     }
 
-    /// Tests the value of the specified lane.
+    /// Tests the value of the specified element.
     ///
     /// # Safety
-    /// `lane` must be less than `LANES`.
+    /// `element` must be less than `self.len()`.
```
programmerjake, line 205: you named the argument `index`
```suggestion
    /// `index` must be less than `self.len()`.
```

Issue #16, opened 2020-10-01
Implement SIMD-specific functions


calebzulawski:
An incomplete list (motivated by a reddit comment)
- [ ] scatter/gather
- [ ] nontemporal load/store
- [ ] lookup tables (possibly related to #11)

Lokathor:
i saw the non-temporal stuff in std::arch when i wrote `safe_arch` but couldn't understand what was going on enough to confidently make a safety claim, so i skipped it.

workingjubilee:
Temporal hinting lets you indicate loads and stores that should not be cached, because you are only reading once and storing once to that location, so you have no need in the rest of the algorithm to frequently read and update it. Basically, it's a way to indicate "please do not shatter my cache with these particular read/writes, I am using that for the ENTIRE REST of my algorithm."

workingjubilee:
I do not see nontemporal loads and stores as "SIMD-specific" but I will acknowledge their heightened usage in SIMD. Obviously scatter/gather is!

programmerjake:
other load/stores that should be implemented are masked load/stores, strided load/stores (often faster than gather/scatter), and combinations thereof.

There's also vector reductions, e.g. reduce-add for things like dot-product.

workingjubilee:
> Emits a !nontemporal store according to LLVM (see their docs). Probably will never become stable.

I am feeling pretty doubtful about exposing an interface to this in std::simd.

workingjubilee:
In #45 we found out that nontemporal stores are not really supported in the codegen backends, not LLVM and as far as I'm aware not in cranelift either, and also that there's no plans to expose this as a scalar API yet, so I took it off the list. That still leaves all the other stuff, of course.

farnoy:
How difficult would it be to expose masked loads & stores? Gather/scatter are already supported, while being more complex. Both are backed by LLVM intrinsics.

I tried working around it by using those gather operations, but the only optimization LLVM does on them it seems, is recognizing a broadcast/splat operation:

```rs
u8x8::gather_select_unchecked(
    bytestring,
    Mask::splat(true),
    usizex8::from_array([
        0, 0, 0, 0, 0, 0, 0, 0
    ]),
    u8x8::splat(0),
);
```

This translates to a single broadcast instruction on AVX, but any divergence in the indices causes inefficient emulation, like when the indices are 0 through 7. Even when specifically compiling for a `target-cpu` that supports AVX-512 masked loads.

calebzulawski:
[...]

PR #374, opened 2023-11-16
Add support for masked loads & stores

farnoy:
Resolves the masked loads/stores checklist item from https://github.com/rust-lang/portable-simd/issues/16

Depends on rustc exposing new intrinsics - https://github.com/rust-lang/rust/pull/117953

### DIFF
--- a/crates/core_simd/src/vector.rs
+++ b/crates/core_simd/src/vector.rs
@@ -311,6 +311,42 @@ where
         unsafe { self.store(slice.as_mut_ptr().cast()) }
     }
 
+    #[must_use]
+    #[inline]
+    pub fn masked_load_or(slice: &[T], or: Self) -> Self {
+        Self::masked_load_select(slice, Mask::splat(true), or)
+    }
+
+    #[must_use]
+    #[inline]
+    pub fn masked_load_select(slice: &[T], enable: Mask<isize, N>, or: Self) -> Self {
+        let ptr = slice.as_ptr();
+        let idxs = Simd::<usize, N>::from_slice(&[
+            0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23,
+            24, 25, 26, 27, 28, 29, 30, 31, 32, 33, 34, 35, 36, 37, 38, 39, 40, 41, 42, 43, 44, 45,
+            46, 47, 48, 49, 50, 51, 52, 53, 54, 55, 56, 57, 58, 59, 60, 61, 62, 63, 64,
+        ]);
+        let enable: Mask<isize, N> = enable & idxs.simd_lt(Simd::splat(slice.len()));
+        unsafe { Self::masked_load_select_ptr(ptr, enable, or) }
+    }
+
+    #[must_use]
+    #[inline]
+    pub unsafe fn masked_load_select_unchecked(
+        slice: &[T],
+        enable: Mask<isize, N>,
+        or: Self,
+    ) -> Self {
+        let ptr = slice.as_ptr();
+        unsafe { Self::masked_load_select_ptr(ptr, enable, or) }
+    }
+
+    #[must_use]
+    #[inline]
+    pub unsafe fn masked_load_select_ptr(ptr: *const T, enable: Mask<isize, N>, or: Self) -> Self {
+        unsafe { intrinsics::simd_masked_load(or, ptr, enable.to_int()) }
+    }
+
     /// Reads from potentially discontiguous indices in `slice` to construct a SIMD vector.
     /// If an index is out-of-bounds, the element is instead selected from the `or` vector.
     ///
@@ -487,6 +523,29 @@ where
     ) -> Self {
         // Safety: The caller is responsible for upholding all invariants
         unsafe { intrinsics::simd_gather(or, source, enable.to_int()) }
+    }
+
+    #[inline]
+    pub fn masked_store(self, slice: &mut [T], enable: Mask<isize, N>) {
+        let ptr = slice.as_mut_ptr();
+        let idxs = Simd::<usize, N>::from_slice(&[
+            0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23,
+            24, 25, 26, 27, 28, 29, 30, 31, 32, 33, 34, 35, 36, 37, 38, 39, 40, 41, 42, 43, 44, 45,
+            46, 47, 48, 49, 50, 51, 52, 53, 54, 55, 56, 57, 58, 59, 60, 61, 62, 63, 64,
+        ]);
+        let enable: Mask<isize, N> = enable & idxs.simd_lt(Simd::splat(slice.len()));
+        unsafe { self.masked_store_ptr(ptr, enable) }
+    }
+
+    #[inline]
+    pub unsafe fn masked_store_unchecked(self, slice: &mut [T], enable: Mask<isize, N>) {
+        let ptr = slice.as_mut_ptr();
+        unsafe { self.masked_store_ptr(ptr, enable) }
+    }
+
+    #[inline]
+    pub unsafe fn masked_store_ptr(self, ptr: *mut T, enable: Mask<isize, N>) {
+        unsafe { intrinsics::simd_masked_store(self, ptr, enable.to_int()) }
     }
 
     /// Writes the values in a SIMD vector to potentially discontiguous indices in `slice`.
--- a/crates/core_simd/src/intrinsics.rs
+++ b/crates/core_simd/src/intrinsics.rs
@@ -107,6 +107,15 @@ extern "platform-intrinsic" {
     /// like gather, but more spicy, as it writes instead of reads
     pub(crate) fn simd_scatter<T, U, V>(val: T, ptr: U, mask: V);
 
+    /// like a loop of reads offset from the same pointer
+    /// val: vector of values to select if a lane is masked
+    /// ptr: vector of pointers to read from
+    /// mask: a "wide" mask of integers, selects as if simd_select(mask, read(ptr), val)
+    /// note, the LLVM intrinsic accepts a mask vector of `<N x i1>`
+    pub(crate) fn simd_masked_load<T, U, V>(val: T, ptr: U, mask: V) -> T;
+    /// like masked_load, but more spicy, as it writes instead of reads
+    pub(crate) fn simd_masked_store<T, U, V>(val: T, ptr: U, mask: V);
+
     // {s,u}add.sat
     pub(crate) fn simd_saturating_add<T>(x: T, y: T) -> T;
 


crates/core_simd/src/vector.rs
```diff
@@ -311,6 +311,42 @@ where
...
+    #[inline]
+    pub fn masked_load_or(slice: &[T], or: Self) -> Self {
+        Self::masked_load_select(slice, Mask::splat(true), or)
+    }
+
+    #[must_use]
+    #[inline]
+    pub fn masked_load_select(slice: &[T], enable: Mask<isize, N>, or: Self) -> Self {
+        let ptr = slice.as_ptr();
+        let idxs = Simd::<usize, N>::from_slice(&[
```
farnoy, line 324: I don't like this. The `rotate_lanes_left` code does a clever thing with a constexpr and maybe that should be preferred?

But ideally, I'd like to have something like `Simd<T, N>::lane_indices() -> Simd<T, N>` (generating 0 through N elements) built-in as it's a useful function to have in user code.

crates/core_simd/src/vector.rs
```diff
@@ -311,6 +311,42 @@ where
...
 
+    #[must_use]
+    #[inline]
+    pub fn masked_load_or(slice: &[T], or: Self) -> Self {
+        Self::masked_load_select(slice, Mask::splat(true), or)
+    }
+
+    #[must_use]
+    #[inline]
+    pub fn masked_load_select(slice: &[T], enable: Mask<isize, N>, or: Self) -> Self {
```
farnoy, line 322: `isize` here is wrong I think. I got it from `gather` but it made sense there as it was masking pointers. I think we want `Mask<<T as SimdElement>::Mask, N>`

calebzulawski: I agree--the mask should either match the element type, or possibly accept any element.

crates/core_simd/src/vector.rs
```diff
@@ -311,6 +312,57 @@ where
...
+        Mask<<T as SimdElement>::Mask, N>: ToBitMask + ToBitMaskArray,
+    {
+        enable &= {
+            let mask = bzhi_u64(u64::MAX, core::cmp::min(N, slice.len()) as u32);
+            let mask_bytes: [u8; 8] = unsafe { core::mem::transmute(mask) };
+            let mut in_bounds_arr = Mask::splat(true).to_bitmask_array();
+            let len = in_bounds_arr.as_ref().len();
+            in_bounds_arr.as_mut().copy_from_slice(&mask_bytes[..len]);
+            Mask::from_bitmask_array(in_bounds_arr)
+        };
```
farnoy, line 341: The previous version was generating suboptimal code. This implementation is still a bit gnarly because I wasn't able to easily cast a `u64` into a `Mask<_, N>` easily. This is because `ToBitmask` uses u8/u16/u32/u64. I don't think we can pull in `num_traits` for an easy solution to this?

This version generates code that matches my handwritten AVX-512 intrinsics.

crates/core_simd/src/vector.rs
```diff
@@ -311,6 +312,57 @@ where
...
+        unsafe { Self::masked_load_select_ptr(ptr, enable, or) }
+    }
+
+    #[must_use]
+    #[inline]
+    pub unsafe fn masked_load_select_ptr(
+        ptr: *const T,
+        enable: Mask<<T as SimdElement>::Mask, N>,
+        or: Self,
+    ) -> Self {
```
calebzulawski, line 362: I think all of the `masked_loads` should take `self` instead of `or`

crates/core_simd/src/vector.rs
calebzulawski: I made some (not yet merged) changes to bitmasks in #375 that may help you (with that change, it's possible to always return a `u64`), but I'm suspicious about this implementation, since it's not particularly portable.  Bitmasks are really only efficient on x86, and most x86 for the time being will not be AVX-512.  I'm confident this codegen will be horrendous on aarch64 :)

I think you should use the old version of the bounds checking, and conditionally check for x86 and N < 64 to use bitmasks.  I would actually split off the x86-optimized enable-adjusting code to its own function.

crates/core_simd/src/vector.rs
```diff
@@ -311,6 +312,57 @@ where
...
     }
 
+    #[must_use]
+    #[inline]
+    pub fn masked_load_or(slice: &[T], or: Self) -> Self
+    where
+        Mask<<T as SimdElement>::Mask, N>: ToBitMask + ToBitMaskArray,
+    {
+        Self::masked_load_select(slice, Mask::splat(true), or)
+    }
```
calebzulawski, line 322: I don't think this function does anything?

calebzulawski: I suppose it's masking the slice length, but I'm not sure I would even call this a masked load.  Also, I'm not sure this is the best way to load from slices on most architectures.

crates/core_simd/src/vector.rs
```diff
@@ -311,6 +312,57 @@ where
...
+
+    #[must_use]
+    #[inline]
+    pub fn masked_load_select(
```
calebzulawski, line 326: I would call this just `masked_load`, or `load_select`.  IMO `masked` and `select` mean the same thing.

crates/core_simd/src/vector.rs
farnoy: It's supposed to enable us to write extremely nice loops like this:

```rust
let mut accum = Simd::<u8, N>::default();
for i in (0..data.len()).step_by(N) {
    accum ^= Simd::masked_load_or(&data[i..], Simd::default());
}
```

No epilogues or scalar fallbacks needed. This could be even shorter since `SimdElement` implies `Default`?

But it's only a reality on AVX-512 for now on account of the codegen

crates/core_simd/src/vector.rs
farnoy: I'll check it out. So far, the bitmasks haven't been a problem for me because of other codegen issues.

I hoped that LLVM implements the fallback by breaking the vector in half, log2 on every step. I was imagining something like this:

```rust
fn load_select<T, const N: usize> (slice: &[T], enable: Mask<T::Mask, N>, or: Simd<T, N>) -> Simd<T, N> where ... {
    if likely(enable.all() && in_bounds(N-1)) {
      Self::from_slice_unchecked(slice)
    } else {
        let low = Simd::<T, N/2>::load_select(slice, enable.low_half());
        let high = Simd::<T, N/2>::load_select(&slice[N..], enable.high_half());
        merge(low, high)
    }
}
```

But instead it generates code that's more like this:

```rust
fn load_select<T, const N: usize> (slice: &[T], enable: Mask<T::Mask, N>, or: Simd<T, N>) -> Simd<T, N> where ... {
    let mut elems = [T::default(); N];
    for lane_ix in 0..N {
        if in_bounds(lane_ix) && enable.test(lane_ix) {
            elems[lane_ix] = slice[lane_ix];
        }
    }
    Self::from_array(elems)
}
```

If I wanted to implement this for other architectures efficiently, could I do it with some form of recursive swizzle & macros? I think the potential for writing simpler code while retaining efficiency is there. The intended use case is to have the `likely(enable.all() && in_bounds(N-1))` branch be always taken, except for the last iteration. I think that would lead to good codegen, since you don't need to duplicate loop bodies if you wanted an epilogue, and it should also be perfectly predictable until the last iteration?

This could still be inefficient in certain cases:
* Even for AVX-512, not needing to do bitmask operations in all but one iteration could free up the core. In my experience, you can probably spare scalar resources for this so it doesn't seem like a big deal
* The caller depends on the mask for their computation after `load_select`. This could be a problem if they compute the mask differently than this library, and the compiler doesn't notice. Basically computing the masks twice, in different ways

crates/core_simd/src/vector.rs
farnoy: I disagree. Something weird about typing `default.load_select(&data[i..])`.

As per other comments, I think I'm going to add this function as well:

```rust
pub fn load_select(slice: &[T]) -> Self {
    Self::load_select_or(slice, Self::default())
}
```

This would match `Simd::from_array()` and `Simd::from_slice()` nicely

crates/core_simd/src/vector.rs
calebzulawski: I would at least rename it--it might be implemented via masked loads, but API-wise I would consider it something else entirely.  This is perhaps `Simd::from_slice`.

crates/core_simd/src/vector.rs
calebzulawski: I don't think const generics are powerful enough to allow anything recursive, yet.  I think you're being too tricky for the compiler--if you'd like to skip the checks, you need to add the explicit `if enable.all()` branch.

calebzulawski: This was obviously easier for me to implement, since I didn't write it generically, but does this look right?
https://rust.godbolt.org/z/TvYqPh4js

First, I couldn't convince the compiler to emit bzhi so I had to resort to asm.  But even then--the naive approach seems more efficient to me.

crates/core_simd/src/vector.rs
calebzulawski: Actually, I remember suggesting this for gather, and no one seemed to like that either. Fine with me

crates/core_simd/src/vector.rs
farnoy: Check out my comment on StackOverflow here: https://stackoverflow.com/questions/75179720/how-to-get-rust-compiler-to-emit-bzhi-instruction-without-resorting-to-platform#comment136619035_75179720

I found two ways to emit bzhi but neither are production-ready IMO.

I will benchmark this some more. I like your constexpr array solution. When I did something similar before (but using a `&'static [usize; 64]`), it loaded the entire slice into multiple vector registers before shuffling.

Your solution is precise and indeed looks clean here, including for `-C target-feature=+sse4.1,+avx2`  version. `mm_maskload` comes from AVX/AVX2. Not sure about SSE & NEON but it's promising.

I don't think it's valid to do `len as i32`, you'd need something like `i32::try_from(len).unwrap_or(i32::MAX)` to do a saturating cast. Otherwise passing in a slice of length `i32::MAX as usize + 1` could make the bounds check reject valid elements, right? This applies to smaller masks too, like i16 and i8.

As for const generics and `N/2`, I'll try adding just a single branch `if likely(enable.all() && in_bounds(N-1))` to see if there's a perf hit on AVX-512, because it's definitely going to be a big win for other targets.

calebzulawski: This is silly, but maybe you can rely on the optimizer here: https://rust.godbolt.org/z/TxPYhx6K4

You would could add that `index` function to one of the `Sealed` traits.

farnoy:
@calebzulawski I pushed out a new version with the changes we discussed.

I've also added temporary toggles for easier benchmarking (`USE_BRANCH` and `USE_BITMASK`).

Here's some data on a simple benchmark reading from one u8 vector and zero-extending it to u16 to write back to another vector. The write buffer is small (1300B) and fully cached. The benchmark is DRAM-bound. This is on Zen 4 (7950X3D) with `-C target-cpu=znver4`

1. `USE_BRANCH=true,USE_BITMASK=true` at ~45GB/s here's the [hot loop](https://uica.uops.info/?code=.LBB17_8%3A%0D%0A%20%20%20%20%20%20%20%20dec%20rax%0D%0A%0D%0A%20%20%20%20%20%20%20%20add%20rcx%2C%2032%0D%0A%20%20%20%20%20%20%20%20add%20r9%2C%20-32%0D%0A%20%20%20%20%20%20%20%20test%20rax%2C%20rax%0D%0A%0D%0A%20%20%20%20%20%20%20%20je%20.LBB17_9%0D%0A%0D%0A.LBB17_4%3A%0D%0A%20%20%20%20%20%20%20%20cmp%20rcx%2C%20rsi%0D%0A%20%20%20%20%20%20%20%20ja%20.LBB17_10%0D%0A%0D%0A%20%20%20%20%20%20%20%20lea%20r10%2C%20%5Brdi%20%2B%20rcx%5D%0D%0A%20%20%20%20%20%20%20%20cmp%20r9%2C%2032%0D%0A%0D%0A%20%20%20%20%20%20%20%20ja%20.LBB17_7%0D%0A%0D%0A%20%20%20%20%20%20%20%20shlx%20r11%2C%20r8%2C%20r9%0D%0A%0D%0A%20%20%20%20%20%20%20%20not%20r11d%0D%0A%0D%0A%20%20%20%20%20%20%20%20kmovd%20k1%2C%20r11d%0D%0A%0D%0A%20%20%20%20%20%20%20%20vmovdqu8%20ymm0%20%7Bk1%7D%20%7Bz%7D%2C%20ymmword%20ptr%20%5Br10%5D%0D%0A%0D%0A%20%20%20%20%20%20%20%20vpmovzxbw%20zmm0%2C%20ymm0%0D%0A%0D%0A%20%20%20%20%20%20%20%20vmovdqu16%20zmmword%20ptr%20%5Brdx%20%2B%202*rcx%5D%20%7Bk1%7D%2C%20zmm0%0D%0A%20%20%20%20%20%20%20%20jmp%20.LBB17_8&syntax=asIntel&uArchs=RKL&tools=uiCA&alignment=0&uiCAHtmlOptions=traceTable&uiCAHtmlOptions=dependencies)
1. `USE_BRANCH=false,USE_BITMASK=true` is at ~41GB/s -  [hot loop](https://uica.uops.info/?code=.LBB17_4%3A%0D%0A%20%20%20%20%20%20%20%20cmp%20rcx%2C%20rsi%0D%0A%20%20%20%20%20%20%20%20ja%20.LBB17_7%0D%0A%0D%0A%20%20%20%20%20%20%20%20cmp%20r9%2C%2032%0D%0A%20%20%20%20%20%20%20%20mov%20r10d%2C%2032%0D%0A%20%20%20%20%20%20%20%20cmovb%20r10%2C%20r9%0D%0A%0D%0A%20%20%20%20%20%20%20%20add%20r9%2C%20-32%0D%0A%20%20%20%20%20%20%20%20dec%20rax%0D%0A%0D%0A%20%20%20%20%20%20%20%20shlx%20r10%2C%20r8%2C%20r10%0D%0A%0D%0A%20%20%20%20%20%20%20%20not%20r10d%0D%0A%0D%0A%20%20%20%20%20%20%20%20kmovd%20k1%2C%20r10d%0D%0A%0D%0A%20%20%20%20%20%20%20%20vmovdqu8%20ymm0%20%7Bk1%7D%20%7Bz%7D%2C%20ymmword%20ptr%20%5Brdi%20%2B%20rcx%5D%0D%0A%0D%0A%20%20%20%20%20%20%20%20vpmovzxbw%20zmm0%2C%20ymm0%0D%0A%0D%0A%20%20%20%20%20%20%20%20vmovdqu16%20zmmword%20ptr%20%5Brdx%20%2B%202*rcx%5D%20%7Bk1%7D%2C%20zmm0%0D%0A%0D%0A%20%20%20%20%20%20%20%20lea%20rcx%2C%20%5Brcx%20%2B%2032%5D%0D%0A%0D%0A%20%20%20%20%20%20%20%20jne%20.LBB17_4&syntax=asIntel&uArchs=RKL&tools=uiCA&alignment=0&uiCAHtmlOptions=traceTable&uiCAHtmlOptions=dependencies)
2. `USE_BRANCH=true,USE_BITMASK=false` 32GB/s - [hot loop](https://uica.uops.info/?code=.LBB17_8%3A%0D%0A%20%20%20%20%20%20%20%20dec%20rax%0D%0A%0D%0A%20%20%20%20%20%20%20%20add%20rcx%2C%2032%0D%0A%20%20%20%20%20%20%20%20add%20r8%2C%20-32%0D%0A%20%20%20%20%20%20%20%20test%20rax%2C%20rax%0D%0A%0D%0A%20%20%20%20%20%20%20%20je%20.LBB17_9%0D%0A%0D%0A.LBB17_4%3A%0D%0A%20%20%20%20%20%20%20%20cmp%20rcx%2C%20rsi%0D%0A%20%20%20%20%20%20%20%20ja%20.LBB17_10%0D%0A%0D%0A%20%20%20%20%20%20%20%20lea%20r9%2C%20%5Brdi%20%2B%20rcx%5D%0D%0A%20%20%20%20%20%20%20%20cmp%20r8%2C%2032%0D%0A%0D%0A%20%20%20%20%20%20%20%20ja%20.LBB17_7%0D%0A%0D%0A%20%20%20%20%20%20%20%20vpbroadcastb%20ymm2%2C%20r8d%0D%0A%0D%0A%20%20%20%20%20%20%20%20vpbroadcastw%20zmm3%2C%20r8d%0D%0A%0D%0A%20%20%20%20%20%20%20%20vpcmpgtb%20k1%2C%20ymm2%2C%20ymm0%0D%0A%0D%0A%20%20%20%20%20%20%20%20vmovdqu8%20ymm2%20%7Bk1%7D%20%7Bz%7D%2C%20ymmword%20ptr%20%5Br9%5D%0D%0A%0D%0A%20%20%20%20%20%20%20%20vpcmpgtw%20k1%2C%20zmm3%2C%20zmm1%0D%0A%0D%0A%20%20%20%20%20%20%20%20vpmovzxbw%20zmm2%2C%20ymm2%0D%0A%0D%0A%20%20%20%20%20%20%20%20vmovdqu16%20zmmword%20ptr%20%5Brdx%20%2B%202*rcx%5D%20%7Bk1%7D%2C%20zmm2%0D%0A%20%20%20%20%20%20%20%20jmp%20.LBB17_8&syntax=asIntel&uArchs=RKL&tools=uiCA&alignment=0&uiCAHtmlOptions=traceTable&uiCAHtmlOptions=dependencies)
3. `USE_BRANCH=false,USE_BITMASK=false` at 2.4GB/s - [hot loop](https://uica.uops.info/?code=.LBB17_4%3A%0D%0A%20%20%20%20%20%20%20%20cmp%20rcx%2C%20rsi%0D%0A%20%20%20%20%20%20%20%20ja%20.LBB17_7%0D%0A%0D%0A%20%20%20%20%20%20%20%20movsx%20r8d%2C%20r8w%0D%0A%0D%0A%20%20%20%20%20%20%20%20mov%20ebx%2C%20127%0D%0A%0D%0A%20%20%20%20%20%20%20%20mov%20r11d%2C%20r8d%0D%0A%20%20%20%20%20%20%20%20sar%20r11d%2C%208%0D%0A%0D%0A%20%20%20%20%20%20%20%20cmp%20r10%2C%20128%0D%0A%0D%0A%20%20%20%20%20%20%20%20cmovb%20ebx%2C%20r10d%0D%0A%0D%0A%20%20%20%20%20%20%20%20cmovae%20r11d%2C%20r9d%0D%0A%0D%0A%20%20%20%20%20%20%20%20add%20r8d%2C%20-8192%0D%0A%20%20%20%20%20%20%20%20add%20r10%2C%20-32%0D%0A%20%20%20%20%20%20%20%20dec%20rax%0D%0A%0D%0A%20%20%20%20%20%20%20%20vpbroadcastb%20ymm2%2C%20ebx%0D%0A%0D%0A%20%20%20%20%20%20%20%20vpbroadcastw%20zmm3%2C%20r11d%0D%0A%0D%0A%20%20%20%20%20%20%20%20vpcmpgtb%20k1%2C%20ymm2%2C%20ymm0%0D%0A%0D%0A%20%20%20%20%20%20%20%20vmovdqu8%20ymm2%20%7Bk1%7D%20%7Bz%7D%2C%20ymmword%20ptr%20%5Brdi%20%2B%20rcx%5D%0D%0A%0D%0A%20%20%20%20%20%20%20%20vpcmpgtw%20k1%2C%20zmm3%2C%20zmm1%0D%0A%0D%0A%20%20%20%20%20%20%20%20vpmovzxbw%20zmm2%2C%20ymm2%0D%0A%0D%0A%20%20%20%20%20%20%20%20vmovdqu16%20zmmword%20ptr%20%5Brdx%20%2B%202*rcx%5D%20%7Bk1%7D%2C%20zmm2%0D%0A%0D%0A%20%20%20%20%20%20%20%20lea%20rcx%2C%20%5Brcx%20%2B%2032%5D%0D%0A%0D%0A%20%20%20%20%20%20%20%20jne%20.LBB17_4&syntax=asIntel&uArchs=RKL&tools=uiCA&alignment=0&uiCAHtmlOptions=traceTable&uiCAHtmlOptions=dependencies)

That last one is interesting as an outlier - uica disagrees that it should be >10x slower. I will investigate these on Sapphire Rapids and Graviton 3. I'm also curious if there's any way to do `i8::try_from(slice.len()).unwrap_or(i8::MAX)` more efficiently. And there also seem to be too many branches per iteration, perhaps it has something to do with how I defined my input & output vectors?

```rust
    let len = bytestring.len();
    output.clear();
    output.reserve(characters);

    unsafe {
        // contents are uninitialized until the loop below ends
        output.set_len(len);
    }

    for i in (0..len).step_by(N) {
        let data = Simd::<u8, N>::load_or_default(&bytestring[i..]);
        let data2: Simd<u16, N> = data.cast();
        data2.masked_store(&mut output[i..], Mask::splat(true));
    }
```

And I've set N=32 for this run, expecting a ymm load that expands into a zmm store

crates/core_simd/src/vector.rs
farnoy: After the rename -- I don't like that this function seems more discoverable than from_slice. Users should prefer `from_slice` if it fits their use-case. This isn't a problem with gather/scatter because that's a specific name with a more complex signature, it's hard to use it by accident.

crates/core_simd/src/vector.rs
```diff
@@ -311,6 +313,104 @@ where
...
+        Simd<<T as SimdElement>::Mask, N>: SimdPartialOrd,
+        Mask<<T as SimdElement>::Mask, N>: core::ops::BitAnd<Output = Mask<<T as SimdElement>::Mask, N>>
+            + core::convert::From<<Simd<<T as SimdElement>::Mask, N> as SimdPartialEq>::Mask>,
+    {
+        Self::load_or(slice, Default::default())
+    }
+
+    #[must_use]
+    #[inline]
+    pub fn load_or(slice: &[T], or: Self) -> Self
```
farnoy, line 334: TODO: I need to test if this works with `T=*const u8` and other pointers. These trait bounds are involved, at least for now, but they shouldn't require `T: Default` for this and other functions that accept an `or: Self`

farnoy:
I optimized `USE_BITMASK=false` to always do the comparison on `i8` masks, which avoids the redundant computation. It's a substantial perf win, but it still misbehaves at specific vector sizes.

crates/core_simd/src/vector.rs
```diff
@@ -974,3 +1119,44 @@ where
...
+{
+    let index = index::<i8, N>();
+    enable & Mask::<M, N>::from(index.simd_lt(Simd::splat(i8::try_from(len).unwrap_or(i8::MAX))))
+}
+
+// This function matches the semantics of the `bzhi` instruction on x86 BMI2
+// TODO: optimize it further if possible
+// https://stackoverflow.com/questions/75179720/how-to-get-rust-compiler-to-emit-bzhi-instruction-without-resorting-to-platform
+#[inline(always)]
+fn bzhi_u64(a: u64, ix: u32) -> u64 {
```
farnoy, line 1156: If we made `ix` a `u64`, this would remove the need to do `std::cmp::min(N, slice.len() as u32)` in call-sites above.

Which is doubly important if we take it further:
```rust

extern "C" {
    #[link_name = "llvm.x86.bmi.bzhi.64"]
    fn x86_bmi2_bzhi_64(x: u64, y: u64) -> u64;
}

#[inline(always)]
fn bzhi_u64(a: u64, ix: u64) -> u64 {
    #[cfg(target_feature = "bmi2")]
    unsafe {
        return x86_bmi2_bzhi_64(a, ix);
    }

    if ix > 63 {
        a
    } else {
        a & (1u64 << ix) - 1
    }
}
```

For whatever reason the intel intrinsic of `_bzhi_u64` takes a `ix: u32`, also in the C/C++ versions. Even a `1u64 as u32` generates a redundant register mov to a 32-bit register, clearing the high bits. All of this is truly unnecessary because the instruction operates on 64b registers.

It's the difference between

```asm
        cmp rsi, 64
        mov r9d, 64
        cmovb r9, rsi
        bzhi r9, r8, r9

```
In the current pure-Rust version,

 ```asm
        mov r9d, esi
        bzhi r9, rdx, r9
```
When calling the `core::arch::x86_64::_bzhi_u64` intrinsic with `slice.len() as u32` (which is incorrect for slices longer than u32::MAX).

```asm
        bzhi r9, rdx, rsi
```

When you use the above code with `#![feature(llvm_link_intrinsics)]`

The layers of mismatched semantics are working against us

farnoy: Nevermind, the version without `min(N, slice.len())` is wrong because the instruction only looks at the lowest 8 bits. So it effectively does a modulo 256 on the second operand.

So `x86_bmi2_bzhi_64(u64::MAX, 525)` is equal to `0x1FFF`, 13 low bits set, because `525 % 256 = 13`

farnoy:
I've updated it for the final version of these intrinsics that shipped in nightly, but I have no idea why it fails on CI, something about the imports for ToBitmask that I've added

workingjubilee (changes requested):
Please add example doctests for these functions.

workingjubilee:
> I've also added temporary toggles for easier benchmarking (USE_BRANCH and USE_BITMASK).

These may be most interesting to keep around via `cfg`, so that people can rebuild the libcore with these using `-Zbuild-std`, but I have no strong opinion here.

I'll take a look at the import problem.

calebzulawski:
I would like fuzz tests especially, I know it's not enabled but I'm particularly suspicious of the bitmask implementation

crates/core_simd/src/vector.rs
```diff
@@ -311,6 +312,104 @@ where
...
+    #[inline]
+    pub fn load_or(slice: &[T], or: Self) -> Self
+    where
+        Mask<<T as SimdElement>::Mask, N>: ToBitMask + ToBitMaskArray,
+        <T as SimdElement>::Mask: Default
+            + core::convert::From<i8>
+            + core::ops::Add<<T as SimdElement>::Mask, Output = <T as SimdElement>::Mask>,
+        Simd<<T as SimdElement>::Mask, N>: SimdPartialOrd,
+        Mask<<T as SimdElement>::Mask, N>: core::ops::BitAnd<Output = Mask<<T as SimdElement>::Mask, N>>
+            + core::convert::From<Mask<i8, N>>,
```
calebzulawski, line 341: These bounds are extreme, and are inflexible enough to make it difficult to adjust the implementation in the future. If it's possible to implement this in a sealed trait or something and get some type erasure to hide the bounds, that would be preferred.

farnoy:
> I would like fuzz tests especially, I know it's not enabled but I'm particularly suspicious of the bitmask implementation

@calebzulawski What do you mean by that, exactly? Do we have facilities in this library already for fuzz testing?

> These may be most interesting to keep around via cfg, so that people can rebuild the libcore with these using -Zbuild-std, but I have no strong opinion here.

@workingjubilee Is that the approach we want to take, given it will have an impact on the trait bounds we publish for these new functions? If we focused on `USE_BRANCH=true` and especially `USE_BITMASK=true/false`, we could simplify the trait bounds. Unless it's not a problem if we use a Sealed trait, but I don't usually write library code so I wouldn't know.

I'm not sure if these toggles are needed, or if the API interface is the final one we should use.

In my testing, even AVX-512 can benefit from `USE_BRANCH=true`, which you'd think is strange because it's generating more code for both cases, but the optimizer is able to unroll loops better when it can promote it to an unmask load. I don't have a good variety of benchmarks though, and I'm not sure what's the best way to explore our options here.

As for the interface potentially not being the final one -- I find that in real code using this, you're often recomputing your own effective Mask that was used for the load/store, as the functions only return the vector, if they return anything at all. This is probably inefficient because user code might compute them slightly differently and the final code will have two computations which are effectively the same? To improve on this, we might want to return the mask that was used by the library - basically with the effects of bounds-checking applied to the user-provided mask.

calebzulawski:
Regarding fuzz testing we have them implemented via wrappers around proptest.  You'll see the helpers in the test_helpers crate. This will be a little more difficult, since it involves pointers.

workingjubilee:
> These may be most interesting to keep around via cfg, so that people can rebuild the libcore with these using -Zbuild-std, but I have no strong opinion here.

@workingjubilee Is that the approach we want to take, given it will have an impact on the trait bounds we publish for these new functions? If we focused on USE_BRANCH=true and especially USE_BITMASK=true/false, we could simplify the trait bounds. Unless it's not a problem if we use a Sealed trait, but I don't usually write library code so I wouldn't know.

Oh, I hadn't thought about that. The Sealed trait pattern still leaves those bounds as visible so we should eschew anything that affects the bounds.

calebzulawski:
I think the direction I would like to see these changes go (in this PR or in a new one, your preference) is for the initial implementation to have no x86 specific code and no bitmask tricks, and no extra bounds on the functions, and then develop it from there.

I don't disagree that there may be better codegen possibilities (though I'd like to see a little more evidence of a particular implementation being faster than another, since the supposedly slower implementation actually has a lower instruction count), but I am getting the feeling that this could be better implemented as improvements to the LLVM intrinsics, or at least in the rustc intrinsics, rather than in library code.

farnoy:
That's fair. Should we worry about backwards compatibility if we introduce a function now but then add extra bounds later? I know that it's an unstable feature for now, just want to confirm that's fine.

Sorry about the delay too. My priorities at work changed and this is not blocking me for now. I still want to ship this as I'll have other uses for it.

calebzulawski:
No worries--I've been too busy to focus much on std::simd lately as well.  My goal is that there should be effectively no bounds on the functions (which is true of all of the other functions for the most part).  This is one of the benefits of fixing the codegen on the LLVM or rust intrinsic side--there are no additional rust bounds to expose.  However, if that's not possible for some reason, it may still be possible to avoid the bounds but still implement this within std::simd by using the intrinsics directly rather than `Simd`'s functions.

jhorstmann:
Of note is that llvm has separate intrinsics for masked loads with an additional len parameter: https://llvm.org/docs/LangRef.html#vector-predication-intrinsics

In my experiments these generate nice code for avx512, and with a mask of all ones and only a len, also good code for avx2 and 32/64 bit values: https://llvm.godbolt.org/z/Yx6fKM853

But I would agree to start with a simpler version, and maybe initially just offer functions taking a mask parameter.

A pattern I often use in my code is to have a helper function that processes a chunk of data, and which has a `const MASKED: bool` generic parameter. Based on that parameter I then either use normal or masked loads, and call the masked version only for the remainder chunk that is smaller than the vector size. Calculating the load mask is then relatively simple, for example for an 8bit bitmask `u8::MAX >> (8 - len)`. Since this is the remainder chunk, the len is guaranteed to be less than 8.

workingjubilee:
aaaaaaaaaaaaaaaaAAAAAAAAAAAAAAAAAA

hi just wanted to let you know I sympathize with
> Sorry about the delay too.

> No worries--I've been too busy to focus much on std::simd lately as well.

anyways, re:
> My goal is that there should be effectively no bounds on the functions (which is true of all of the other functions for the most part).

yeah, ideally. I'd rather any bounds be the kind that type inference can easily take care of 99.9% of the time.

PR #376, opened 2023-11-17
Find first set element in a mask

### LINKED ISSUE #326: Find-first-set style operations directly on masks
A very common use case for the ToBitMask trait is searching for / iterating over set bytes in a vector, i.e. searching for a character in a string.

Neon does not have an equivalent of movemask_epi, and llvm generates very poor code when emulating it. Quite a few libraries have faster implementations of movemask_epi, and this is something llvm should really fix.

However, if you assume:

1. That your mask has all bytes set to -1 or 0
2. All you want to do is find the first set lanes or iterate over set lanes

then [an optimisation such as this](https://community.arm.com/arm-community-blogs/b/infrastructure-solutions-blog/posts/porting-x86-vector-bitmask-optimizations-to-arm-neon) are significantly faster - [you can see it here](https://rust.godbolt.org/z/8YvEfEPT6):

```asm
example::find_first_match:
        sub     sp, sp, #16
        ldr     q0, [x0]
        dup     v1.16b, w1
        cmeq    v0.16b, v0.16b, v1.16b
        umov    w8, v0.b[1]
        and     w8, w8, #0x1
        umov    w9, v0.b[0]

        repeat many times...

        umov    w9, v0.b[15]
        orr     w8, w8, w9, lsl #15
        and     w8, w8, #0xffff
        clz     w8, w8
        sub     w8, w8, #16
        and     w0, w8, #0xffff
        add     sp, sp, #16
        ret
```

vs 

```asm
example::find_first_match_fast:
        dup     v0.16b, w1
        ldr     q1, [x0]
        cmeq    v0.16b, v0.16b, v1.16b
        shrn    v0.8b, v0.8h, #4
        fmov    x8, d0
        rbit    x8, x8
        clz     x8, x8
        ubfx    x0, x8, #2, #30
        ret
```

This simple case could arguably be more suitable for an llvm optimisation pass if there was a way to inform llvm that the mask vectors *always* consisted of -1 and 0, however I don't think that's at all possible right now? The iterator case seems potentially much more complicated to implement in llvm.

This is potentially getting into niche use cases (although in my experience most uses of to_bitmask involve iteration over set bits) and would involve putting architecture specific code into portable_simd instead of relying on llvm intrinsics, so maybe it's not possible, but seems worth discussing. At the moment any sort of string/byte search using std::simd is going to perform poorly on arm, or only be implemented for x86 - [take this commit](https://github.com/rust-lang/rust/pull/103779/files)

calebzulawski:
Depends on #375, so many of those changes are included here (I'll rebase if that is merged).  Closes #326.

### DIFF
--- a/crates/core_simd/src/masks.rs
+++ b/crates/core_simd/src/masks.rs
@@ -13,7 +13,7 @@
 mod mask_impl;
 
 use crate::simd::{
-    cmp::SimdPartialEq, intrinsics, LaneCount, Simd, SimdElement, SupportedLaneCount,
+    cmp::SimdPartialEq, intrinsics, LaneCount, Simd, SimdCast, SimdElement, SupportedLaneCount,
 };
 use core::cmp::Ordering;
 use core::{fmt, mem};
@@ -35,6 +35,10 @@ mod sealed {
 
         fn eq(self, other: Self) -> bool;
 
+        fn as_usize(self) -> usize;
+
+        type Unsigned: SimdElement;
+
         const TRUE: Self;
 
         const FALSE: Self;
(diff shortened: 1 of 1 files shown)


programmerjake (approved):
looks nice! should be rebased on master once #375 is merged.

PR #377, opened 2023-11-19
Follow-up fixes for to_bitmask

calebzulawski:
I also noticed my non-power-of-two code wasn't hit because the `#[cfg]` was wrong (should have been `#[cfg(feature = "all_lane_counts")]`.  It turns out the bitmask intrinsics need a power of two (otherwise they need to use e.g. `u17` as the return type), so I changed to expand to power-of-twos, at least for now.


calebzulawski:
I'm not sure why our clippy config didn't pick up the bad `#[cfg]`, I only saw it when trying to sync back to rust-lang/rust

programmerjake (approved):

PR #380, opened 2023-12-03
Fix bitmask vector bit order

### LINKED ISSUE #379: from_bitmask_vector on big-endian calls simd_select_bitmask with the mask at the wrong end of the byte
This code, running on big-endian
```rust
    let bitmask = u8x4::from_array([0b00001101, 0, 0, 0]);
    assert_eq!(
        mask32x4::from_bitmask_vector(bitmask),
        mask32x4::from_array([true, false, true, true]),
    );
```
ends up calling `simd_select_bitmask` with the mask being `0b10110000`. I think that's wrong; for simd_bitmask the mask lives in the least-sigificant bits (see https://github.com/rust-lang/portable-simd/pull/378) so I would expect that `simd_select_bitmask` should be called with `0b00001011`.

This issue is reminiscent of https://github.com/rust-lang/portable-simd/pull/267.

calebzulawski:
Fixes #379

### DIFF
--- a/crates/core_simd/src/masks/full_masks.rs
+++ b/crates/core_simd/src/masks/full_masks.rs
@@ -157,6 +157,9 @@ where
                 for x in bytes.as_mut() {
                     *x = x.reverse_bits()
                 }
+                if N % 8 > 0 {
+                    bytes.as_mut()[N / 8] >>= 8 - N % 8;
+                }
             }
 
             bitmask.as_mut_array()[..bytes.as_ref().len()].copy_from_slice(bytes.as_ref());
@@ -179,6 +182,9 @@ where
             if cfg!(target_endian = "big") {
                 for x in bytes.as_mut() {
                     *x = x.reverse_bits();
+                }
+                if N % 8 > 0 {
+                    bytes.as_mut()[N / 8] >>= 8 - N % 8;
                 }
             }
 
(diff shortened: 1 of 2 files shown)


calebzulawski:
@RalfJung looks good?

RalfJung:
I honestly can't tell, I don't really understand the code in this library. We'll see if this makes the Miri tests pass -- once it propagates into the rustc repository.

I also have no idea what should happen for cases like `N == 12`. Should the first or the last element of the array be the one that is only partially filled? I thought it'd have to be the first element, but then I looked at an example: on little endian, mask `0b001101001001` is the same as `[0b01001001, 0b0011]`, and on big endian the same mask is written `0b100100101100` as a single integer and `[0b10010010, 0b1100]` as an array. Looks like indeed converting the arrays works by reversing the bits in each element, and then fixing up the last element as that is always the one that is "partial".

workingjubilee:
We should explain why this is the correct behavior before we commit it, which means expanding our written spec for `simd_select_bitmask` and `simd_bitmask`.

calebzulawski:
Perhaps we should expose the intrinsics in something like core::intrinsics::simd and use that documentation as the actual semantics

RalfJung:
Any idea when this will be propagated to the rustc repo? :)

calebzulawski:
I just opened https://github.com/rust-lang/rust/pull/121269

Issue #384, opened 2023-12-13
 UB for certain transmutes could be easier to detect


workingjubilee:
We could insert `core::intrinsics::assume` or `core::hint::unreachable_unchecked` or something?

Ideally this would not pass miri: https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=3b0672dc5b90a1f626050388bb0478cb

PR #385, opened 2023-12-14
Assume masks are correct

workingjubilee:
This allows miri to detect when they are not, and may be exploited by LLVM during optimization. Closes https://github.com/rust-lang/portable-simd/issues/384.

Are there any other places we want to inject such assumptions? I avoided it on the `test_unchecked` and such because I think it is likely unnecessary on those cases and may result in aggressive over-optimization, but anywhere it's likely to be called once instead of once-per-element seems like a fine place to put an `assume`.

### DIFF
--- a/crates/core_simd/src/lib.rs
+++ b/crates/core_simd/src/lib.rs
@@ -4,6 +4,7 @@
     const_maybe_uninit_as_mut_ptr,
     const_mut_refs,
     convert_float_to_int,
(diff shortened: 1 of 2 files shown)


programmerjake (approved):

programmerjake:
maybe wait till the `exposed_provenance` issue is fixed?

calebzulawski (approved):

Issue #381, opened 2023-12-11
Move intrinsic declarations to core and verify behavior


calebzulawski:
The intrinsics should exist in core, which will make them easier to test and document.  There is currently no single set of documentation explaining the semantics of the intrinsics, so this module would serve as the "definitive" description.

Additionally, we should ensure that the semantics match the documentation and is sound.

RalfJung:
> Additionally, we should ensure that the semantics match the documentation and is sound.

Note that there are two semantics:
- the codegen backend, which "just" has to generate correct behavior whenever the safety requirements are met
- Miri, which also has to check whether the safety requirements are actually met

PR #386, opened 2023-12-29
Use core::intrinsics

calebzulawski:
Switches some of the intrinsics for #381. The remainder will be switched after rust-lang/rust#119213


crates/core_simd/tests/pointers.rs
```diff
@@ -1,4 +1,4 @@
-#![feature(portable_simd, strict_provenance)]
+#![feature(portable_simd, strict_provenance, exposed_provenance)]
```
programmerjake, line 1: Please split this into a separate commit and in the commit message link to the PR where exposed_provenance was split out.

programmerjake: @bjorn3 this should allow you to drop [the patch](https://github.com/rust-lang/rust/commit/1095e0165c36bd65f5fb3af9955e565137238b65) once this is merged and subtree synced.

bjorn3: I can actually drop it as soon as this PR is merged. cg_clif's test suite currently directly clones this PR to test it rather than using the copy used by the standard library.

programmerjake (approved):

PR #389, opened 2024-02-09
fix feature flags

RalfJung:
const_intrinsic_copy stabilization got reverted in https://github.com/rust-lang/rust/pull/117905.

### DIFF
--- a/crates/core_simd/src/lib.rs
+++ b/crates/core_simd/src/lib.rs
@@ -12,9 +12,26 @@
     repr_simd,
     simd_ffi,
     staged_api,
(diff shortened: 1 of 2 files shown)


RalfJung:
Okay this is painful, we'll need a bunch of careful `#[cfg]` in the lib.rs that enables *exactly* the stdarch features needed for each configuration. Enabling features that don't exist for the current target just shows "unknown feature" errors...

programmerjake (approved):
assuming it compiles everywhere, lgtm!

calebzulawski:
Thanks!

PR #393, opened 2024-02-17
Assume masks are correct

### LINKED ISSUE #384: UB for certain transmutes could be easier to detect

calebzulawski:
Reopened from #385.  Closes #384.

I should have waited for CI to go green :)

### DIFF
--- a/crates/core_simd/src/masks.rs
+++ b/crates/core_simd/src/masks.rs
@@ -12,7 +12,7 @@
 )]
 mod mask_impl;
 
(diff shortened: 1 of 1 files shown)


calebzulawski:
I think there must be an infinite loop in `valid` because it causes a stack overflow.

PR #399, opened 2024-02-29
Add support for masked loads & stores

farnoy:
Continuation of #374, but simplified.

1. Got rid of the optimization variants `USE_BRANCH`, `USE_BITMASK`
   1. We don't do a branch to check if all lanes are enabled and we don't calculate bitmasks with integer code
   1. This won't be the most optimal approach until LLVM starts to optimize these further
1. Added documentation for load operations. Let me know if this complies with established norms and I'll document stores as well
1. Simplified public bounds to bare essentials

### DIFF
--- a/crates/core_simd/src/vector.rs
+++ b/crates/core_simd/src/vector.rs
@@ -1,5 +1,6 @@
 use crate::simd::{
     cmp::SimdPartialOrd,
+    num::SimdUint,
     ptr::{SimdConstPtr, SimdMutPtr},
     LaneCount, Mask, MaskElement, SupportedLaneCount, Swizzle,
 };
@@ -261,6 +262,7 @@ where
     /// # Panics
     ///
     /// Panics if the slice's length is less than the vector's `Simd::N`.
+    /// Use `load_or_default` for an alternative that does not panic.
     ///
     /// # Example
     ///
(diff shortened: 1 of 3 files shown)


calebzulawski:
Can you add to the documentation for `from_slice` that you can use `load_or` or `load_or_default` for a non-panicking version?

crates/core_simd/src/vector.rs
```diff
@@ -492,6 +622,31 @@ where
         unsafe { core::intrinsics::simd::simd_gather(or, source, enable.to_int()) }
     }
 
+    #[inline]
+    pub fn masked_store(self, slice: &mut [T], mut enable: Mask<<T as SimdElement>::Mask, N>) {
+        enable &= mask_up_to(enable, slice.len());
```
koyeung, line 627: Would `mask_up_to()` already do "and" ops to the input `enable`? 

```rust
enable & Mask::<M, N>::from_bitmask_vector(lt.to_bitmask_vector())
```
i.e. it could be `enable = mask_up_to(enable, slice.len());` here. 

(or dropping `enable &` from function `mask_up_to()`)

farnoy:
Not sure why the new doctests are failing. The errors about store functions being undocumented are valid and I've yet to add them.

crates/core_simd/src/vector.rs
```diff
@@ -979,3 +1134,27 @@ where
...
+
+#[inline]
+fn mask_up_to<M, const N: usize>(len: usize) -> Mask<M, N>
+where
+    LaneCount<N>: SupportedLaneCount,
+    M: MaskElement,
+{
+    let index = lane_indices::<i8, N>();
+    let lt = index.simd_lt(Simd::splat(i8::try_from(len).unwrap_or(i8::MAX)));
+    Mask::<M, N>::from_bitmask_vector(lt.to_bitmask_vector())
```
koyeung, line 1159: Maybe it has been considered. any idea to avoid calling `lane_indices` (and no need to run the for loop inside the function each time the `mask_up_to` is called), by :

```rust
fn mask_up_to<...>() -> Mask<M,N> ... {
    Mask::<M, N>::from_bitmask((1_u64 << len) - 1)
}
```

pls let me know if I misunderstood the function :)

farnoy: The previous version of this PR had the option to use this exact bitmask and indeed, it tends to produce better performing code https://github.com/rust-lang/portable-simd/pull/374#issuecomment-1817863838

Plus, it's just simpler, so I'm in favor. Any objections @calebzulawski? This would not require adding any extra trait bounds ever since you removed ToBitMask.

farnoy: Oh, and it would have to be this to prevent an overflow on the shift:

```rust
fn mask_up_to<...>() -> Mask<M,N> ... {
    if len >= 64 {
        Mask::<M, N>::splat(true)
    } else {
        Mask::<M, N>::from_bitmask((1_u64 << len) - 1)
    }
}
```

koyeung: it might be...if desirable to avoid the literal "64"
```rust
 if len >= u64::BITS as usize { ...
```

calebzulawski, line 1159: Why the to/from bitmask here? Can you just cast?

calebzulawski: I think you need to condition that on x86, because that will be worse on other architectures.  Also, you should condition that on the lane count being no more than 64, because we will support larger in the future. Alternatively, use a bitmask vector which works for larger than 64.

koyeung: any comment if giving shortcuts when `len` not less than lane count?

```rust
fn mask_up_to<...>() -> Mask<M,N> ... {
    if len >= N {
        Mask::<M,N>::splat(true)
    }else{
        let index = lane_indices::<i8, N>();
        let lt = index.simd_lt(Simd::splat(i8::try_from(len).unwrap_or(i8::MAX)));
        // .... Mask::<M, N>::from_bitmask_vector(lt.to_bitmask_vector())
    }
}
```

jhorstmann: I think the comparison based version also generates better code on non-avx512 machines: https://rust.godbolt.org/z/6edK1rx9M

Regarding the initial question, since the length parameter is a const generic we would expect llvm to always precompute the mask vector, so at runtime it is a single vector load.

farnoy: We're at risk of having the same conversation as in the previous PR, which we wanted to avoid with this spin-off PR.

Let's go with the bitmask vector implementation, which is more generic. We can revisit this later, or LLVM can learn to optimize this better, or users who are performance-conscious today can use the unsafe functions to provide their own mask, calculated in a way that best fits their architecture and usecase.

farnoy:
Should be green now. Please review

crates/core_simd/src/vector.rs
```diff
@@ -492,6 +630,77 @@ where
...
+    /// # use simd::{Simd, Mask};
+    /// let mut arr = [0i32; 4];
+    /// let write = Simd::from_array([-5, -4, -3, -2]);
+    /// let enable = Mask::from_array([false, true, true, true]);
+    ///
+    /// write.masked_store(&mut arr[..3], enable);
+    /// assert_eq!(arr, [0, -4, -3, 0]);
+    /// ```
+    #[inline]
+    pub fn masked_store(self, slice: &mut [T], mut enable: Mask<<T as SimdElement>::Mask, N>) {
```
calebzulawski, line 652: For consistency, should this be `store_select`?

crates/core_simd/src/vector.rs
```diff
@@ -314,6 +315,143 @@ where
...
+    /// let result = Simd::load_or(&vec, or);
+    /// assert_eq!(result, Simd::from_array([10, 11, -3, -2]));
+    /// ```
+    #[must_use]
+    #[inline]
+    pub fn load_or(slice: &[T], or: Self) -> Self {
+        Self::load_select(slice, Mask::splat(true), or)
+    }
+
+    /// Reads contiguous elements from `slice`. Each lane is read from memory if its
```
calebzulawski, line 362: Please change all instances of "lane" to "element" in docs

calebzulawski (changes requested):
Looks good to me other than these nits

calebzulawski (approved):

crates/core_simd/src/vector.rs
```diff
@@ -979,3 +1188,27 @@ where
...
+    Simd::from_array(index)
+}
+
+#[inline]
+fn mask_up_to<M, const N: usize>(len: usize) -> Mask<M, N>
+where
+    LaneCount<N>: SupportedLaneCount,
+    M: MaskElement,
+{
+    let index = lane_indices::<i8, N>();
```
programmerjake, line 1211: this should be `isize` instead of `i8`, since we will eventually have types with more than 127 elements.

programmerjake (changes requested):

crates/core_simd/src/vector.rs
programmerjake: well, apparently LLVM is more terrible than I thought: https://rust.godbolt.org/z/vj75dsT51

calebzulawski: What about using M directly?  If type inference doesn't play nice, using intrinsics directly can probably get past it.

farnoy: Doing it on `M` would require `M: core::ops::TryFrom<usize>` which would appear on the public interface, and then there's still no way to convert `len: usize` to `M` safely, nor is there a `core::ops::Max` that would give you a generic `::MAX` constant. Am I missing anything?

calebzulawski: You can add hidden helper functions/consts like that (e.g. a `from_usize`) to the Sealed trait on MaskElement.  Not super elegant, but it will work and eliminate the bounds

programmerjake: ok, I came up with some code that still works with arbitrarily many lanes and compiles to good code: https://rust.godbolt.org/z/Gqx85P4Ex

programmerjake: the `M::Unsigned: SimdCast + Into<u64>` bounds should be added to `Sealed::Unsigned` so the user doesn't need to specify anything more than `M: MaskElement`

farnoy: @programmerjake I don't think I can add `Sealed::Unsigned: Into<u64>` because usize does not implement that trait.

```
error[E0277]: the trait bound `u64: From<usize>` is not satisfied
  --> crates\core_simd\src\masks.rs:96:24
   |
96 | impl_element! { isize, usize }
   |                        ^^^^^ the trait `From<usize>` is not implemented for `u64`, which is required by `usize: Into<u64>`
   |
```

calebzulawski: I don't think `where` bounds are picked up when you use the trait, they only constrain it.  I would just add functions to Sealed anyway, to avoid leaking bounds (that we can't remove in the future if we want to)

programmerjake: I originally had that as `u128` instead of `u64`, though like caleb suggests, a Sealed method may be best.

programmerjake: > I don't think `where` bounds are picked up when you use the trait, they only constrain it.

associated type bounds seem to be picked up: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=4d88c89dd7ee65b133b92a3ccedca39d

farnoy: Pushed the change. I barely understand it but I think it should work?

crates/core_simd/src/vector.rs
```diff
@@ -979,3 +1189,37 @@ where
...
+fn mask_up_to<M, const N: usize>(len: usize) -> Mask<M, N>
+where
+    LaneCount<N>: SupportedLaneCount,
+    M: MaskElement,
+{
+    let index = lane_indices::<N>();
+    let max_value: u64 = M::max_unsigned();
+    macro_rules! case {
+        ($ty:ty) => {
+            if N < <$ty>::MAX as usize && max_value as $ty as u64 == max_value {
```
farnoy, line 1215:
```suggestion
            if N < <$ty>::MAX as usize && max_value == <$ty>::MAX as u64 {
```

Is there a difference?

programmerjake: yes those are not equivalent, `max_value as $ty as u64 == max_value` is `true` when `M = i8` and `$ty = u16` which is what you'd get with `Mask<i8, 256>`. basically it's checking `$ty::BITS >= M::BITS`

programmerjake (approved):

crates/core_simd/src/vector.rs
farnoy: I don't think I can write a test for it until SupportedLaneCount is increased

farnoy:
Squashed to clean up the history

crates/core_simd/src/vector.rs
programmerjake: > I don't think I can write a test for it until SupportedLaneCount is increased

yeah, that's fine. if it's broken we can fix it later, imo the important part is that we tried to not bake in things we know will be broken.

crates/core_simd/src/vector.rs
```diff
@@ -979,3 +1189,37 @@ where
...
+                return index.cast().simd_lt(Simd::splat(len.min(N) as $ty)).cast();
+            }
+        };
+    }
+    case!(u8);
+    case!(u16);
+    case!(u32);
+    case!(u64);
+    index.simd_lt(Simd::splat(len)).cast()
+}
```
calebzulawski, line 1225: How about adding a `from_usize` and doing something like this to avoid casts?
```suggestion
#[inline]
fn lane_indices<M, const N: usize>() -> Simd<M, N>
where
    M: MaskElement,
    LaneCount<N>: SupportedLaneCount,
{
    Simd::from_array(core::array::from_fn(M::from_usize))
}

#[inline]
fn mask_up_to<M, const N: usize>(len: usize) -> Mask<M, N>
where
    LaneCount<N>: SupportedLaneCount,
    M: MaskElement,
{
    if N as u64 < M::max_unsigned() {
        lane_indices::<M, N>().simd_lt(Simd::splat(M::from_usize(len)))
    } else {
        lane_indices::<usize, N>().simd_lt(Simd::splat(len)).cast()
    }
}
```

programmerjake: part of the `case!` logic is picking the minimum viable element size, since that's usually more efficient, so it would pick 16-bit elements for `Simd<u8, 256>` and 32-bit elements for `Simd<u8, 65536>` (yes, afaict that is theoretically possible in a single vector on RISC-V V).

your suggestion would just pick the mask's element size and if that isn't big enough, give up and use full `usize` elements even if 16-bit elements would suffice for all realistic vector types. Unfortunately LLVM doesn't seem to be able to optimize vectors to use a smaller element size unless you explicitly cast to that size before using `simd_lt`

programmerjake: also `lane_indices::<M, N>().simd_lt(Simd::splat(M::from_usize(len)))` is just plain incorrect, if `M` is `i8` and `len` is `256` then it will wrap around (or panic if `from_usize` checks) and act as if `len == 0`

calebzulawski: I think for the usual case (vectors with well under 256 elements) codegen is likely to be better without a cast?  How about keeping as is, but trying to use M first if it fits?

programmerjake: maybe? the casts optimized away completely when I tried...

calebzulawski: Interesting, I've had bad luck with casts, but that's usually with more complex operations. If we think the casts won't matter it's fine with me

programmerjake: > Would it be sufficient to consider only value of N when determine the element size? i.e. use 8-bit elements for `Simd<u64,16>`

no, because iirc LLVM doesn't optimize-out the element-size conversions then.

PR #400, opened 2024-03-03
feat: add SIMD float math functions (exp, exp2, log, log2, log10, sin…

avhz:
Hi,

In relation to #14, I've added some math functions to check with you if I'm on the right track. 

I simply followed the implementation that was already there for `sqrt`, and for functions that I easily found were already in `intrinsics`. 

Let me know if it's the right way to go :) 

Cheers!

Edit: I manually checked the output values against R, and they were all correct.

### DIFF
--- a/crates/std_float/src/lib.rs
+++ b/crates/std_float/src/lib.rs
@@ -65,6 +65,62 @@ pub trait StdFloat: Sealed + Sized {
         unsafe { intrinsics::simd_fsqrt(self) }
     }
 
+    /// Produces a vector where every lane has the sine of the value
+    /// in the equivalently-indexed lane in `self`.
+    #[inline]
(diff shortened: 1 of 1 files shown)


calebzulawski:
Thank you!  I'm now noticing how lacking our tests are in this area, but we can leave that for another day.

avhz:
My pleasure :)

I did notice that too, and I didn't check any special values like `T:MAX`, infinity, etc. 
But I can add tests for these. 

In #14 you mentioned adding intrinsics to the compiler, is this completely done ?

calebzulawski:
I actually thought the intrinsics you used here didn't exist yet.  There might be some still missing, but none stand out to me right now.  Also, I opened #401 to add the tests.

calebzulawski:
After a quick check, the intrinsics that seem to be missing are the `pow`/`powi` intrinsics

PR #406, opened 2024-03-22
Use v1 prelude to match core

calebzulawski:
A workaround for rust-lang/rust#122912 for discovered in rust-lang/rust#122905.


workingjubilee:
Hmm, sure!

PR #410, opened 2024-04-01
Add loongarch64 vendor conversions

heiher:
Add conversions for loongarch64-simd types and portable-simd types.


programmerjake (approved):

programmerjake:
I was going to ask if you can try to add loongarch64 to CI, but `cross` doesn't yet support it: https://github.com/cross-rs/cross/issues/1404

PR #412, opened 2024-04-09
do not run the 4-lane tests in Miri

RalfJung:
This should cut down the CI times for the Miri job.

Also only test one non-power-of-2 configuration (if/when that ever gets enabled) in Miri, again to avoid the tests taking too long. I don't see why odd vs even non-power-of-2 would make a big difference for Miri.

### DIFF
--- a/crates/test_helpers/src/lib.rs
+++ b/crates/test_helpers/src/lib.rs
@@ -539,12 +539,12 @@ macro_rules! test_lanes {
                     #[cfg_attr(target_arch = "wasm32", wasm_bindgen_test::wasm_bindgen_test)];
                     lanes_1 1;
                     lanes_2 2;
-                    lanes_4 4;
                 );
 
                 #[cfg(not(miri))] // Miri intrinsic implementations are uniform and larger tests are sloooow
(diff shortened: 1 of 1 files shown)


programmerjake:
> I don't see why odd vs even non-power-of-2 would make a big difference for Miri.

oh, that's because of alignment -- the layout I proposed would have align 1 for `Simd<u8, 5>` but align 2 for `Simd<u8, 6>`, since basically it picks the biggest alignment that doesn't require padding.

RalfJung:
Current layout adds padding to make the size a power of 2. There does not seem to be consensus on what the layout should be. Anyway that is a discussion for <https://github.com/rust-lang/portable-simd/issues/319>.

But no matter whether alignment is the next power of 2 or the largest power that divides the unpadded size, even length does not seem fundamentally different from odd length to me.

RalfJung:
With this, Miri takes 27min. The second-slowest job seems to be "default on aarch64-unknown-linux-gnu" which takes 19min.

The only way I see to further reduce CI times is to *only* test the 2-lane tests in Miri.

Issue #402, opened 2024-03-06
migration pitfall from packed_simd: PartialOrd behaves differently?


KonradHoeffner:
I've recently [tripped over](https://github.com/KonradHoeffner/rsdict/pull/1) the different behavior in the minimum function when migrating from packed_simd to portable-simd.

I'm a SIMD beginner so sorry if this is obvious for the experts but it seems as if the minimum and maximum functions of packed_simd extract the minimum and maximum value for each element while portable-simd seems to choose one SIMD value in it's entirety based on lexicographic ordering.

For example let's assume a hypothetical 3x8bit type with values represented like an array as a = [1,2,100] and b = [3,4,5].
Now packed_simd min(a,b) would output [1,2,5] (which makes sense to me) while portable-simd seems to output [1,2,100] (because that one comes first in lexicographic ordering), whereas simd_min outputs [1,2,5].

Is that working as intended and are there use cases for this? If yes, a migration warning in the README and/or a comment explaining this on top of the PartialOrd trait impl could be helpful so that others don't repeat this mistake.

Currently the code is:

```
impl<T, const N: usize> PartialOrd for Simd<T, N>
where
    LaneCount<N>: SupportedLaneCount,
    T: SimdElement + PartialOrd,
{
    #[inline]
    fn partial_cmp(&self, other: &Self) -> Option<core::cmp::Ordering> {
        // TODO use SIMD equality
        self.to_array().partial_cmp(other.as_ref())
    }
}
```

Does the "TODO use SIMD equality" mean that the behavior will be changed? Or will it stay the same and just be faster?

calebzulawski:
This is definitely a concern, but the packed_simd behavior was a bit of a hack as well.

In packed_simd, you were not actually interacting with PartialOrd at all, but an associated function called `min` that shadowed `PartialOrd::min`.  If you look at PartialOrd, you'll realize that function is required to return one of the two inputs untouched.  It's a total ordering function, unrelated to and incompatible with the SIMD elementwise minimum.

We changed the name to discourage relying on the scope to select the correct trait and instead used different names.  This can still be confusing, but the only alternative I see would be not implementing PartialOrd for vectors at all.

calebzulawski:
Regarding the TODO, it only affects implementation, not behavior.

KonradHoeffner:
Thank you for the clarification!
What do you think about a rustdoc comment that explains this and links from PartialOrd to simd_min and simd_max?
[...]

PR #413, opened 2024-04-10
document PartialOrd difference to simd_min and simd_max

KonradHoeffner:
Avoids a migration pitfall from packed_simd as described in #402.
This adds a single line of documentation, no code change. 

Fixes: #402

### DIFF
--- a/crates/core_simd/src/vector.rs
+++ b/crates/core_simd/src/vector.rs
@@ -945,6 +945,7 @@ where
 {
 }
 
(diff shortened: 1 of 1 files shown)


programmerjake:
can you add it for `Ord` too?

KonradHoeffner:
@programmerjake Good point! Added the comment for Ord as well.

programmerjake (approved):

PR #417, opened 2024-04-30
Make splat const fn

calebzulawski:


programmerjake (approved):

programmerjake:
looks like github switched to using apple m1, which llvm/rustc can't figure out the native cpu of, also looks like the m1 doesn't support avx properly...

calebzulawski:
The good news is now we can have a native arm test! I'll try to fix that in #419

thomcc:
> m1 doesn't support avx properly...

Yes, it only supports up to SSE4.2

PR #418, opened 2024-04-30
Update rust version

calebzulawski:


calebzulawski:
Whoops, we already had the toolchain locked. Thought I hadn't done it yet. Why did we suddenly get the `inline_const` stabilization errors? Maybe something wrong in the CI script

programmerjake:
> Whoops, we already had the toolchain locked. Thought I hadn't done it yet. Why did we suddenly get the `inline_const` stabilization errors? Maybe something wrong in the CI script

because inline consts were stabilized on nightly: https://github.com/rust-lang/rust/pull/104087#issuecomment-2075679329

calebzulawski:
I meant why was I seeing them when we were supposed to be locked to an older toolchain

programmerjake:
> I meant why was I seeing them when we were supposed to be locked to an older toolchain

the updated version uses today's (or yesterday depending on time zones) nightly...I wouldn't call that old. the stabilization PR was merged about a week ago.

PR #419, opened 2024-04-30
Fix macos tests in CI

calebzulawski:


.github/workflows/ci.yml
```diff
@@ -56,26 +56,19 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        target: [x86_64-pc-windows-msvc, i686-pc-windows-msvc, i586-pc-windows-msvc, x86_64-unknown-linux-gnu, x86_64-apple-darwin]
```
programmerjake, line 59: I think we should still run some tests on `x86_64-apple-darwin`, since it is a tier 1 target. just don't enable any target features.

programmerjake (approved):

PR #429, opened 2024-08-25
Add `abs_diff` function to `SimdInt` and `SimdUint` traits

okaneco:
Implement `abs_diff` for signed and unsigned integer vectors

### DIFF
--- a/crates/core_simd/src/simd/num/int.rs
+++ b/crates/core_simd/src/simd/num/int.rs
@@ -1,6 +1,6 @@
 use super::sealed::Sealed;
 use crate::simd::{
-    cmp::SimdPartialOrd, num::SimdUint, LaneCount, Mask, Simd, SimdCast, SimdElement,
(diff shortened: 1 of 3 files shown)


calebzulawski:
Thank you! Can you move the test under the macros that generate all of the int/uint tests? That will allow the function to be fuzzed over all the types/lane counts/inputs

calebzulawski:
I think it would be the same as this: https://github.com/rust-lang/portable-simd/blob/283acf44cc889a159483c4805cf29c75d6c13735/crates/core_simd/tests/ops_macros.rs#L553

However, it should be under the signed and unsigned tests, not float.

okaneco:
Thanks, I'll give that a try

okaneco:
Removed the original test file I added
Added the tests to `impl_signed_tests` and `impl_unsigned_tests`

With that advice, it was easy to add the tests

calebzulawski:
Looks good, thanks!

PR #431, opened 2024-08-25
Fix avx512vbmi swizzle_dyn implementation

cvijdea-bd:
With target_features = "avx512vbmi", swizzle_dyn with N = 32 did not set output lanes to 0 when the input index was out of range, because it used `_mm256_permutexvar_epi8` (vpermb) which, unlike `_mm256_shuffle_epi8` (vpshufb), does not provide that behaviour.

This PR fixes the problem and adds the avx512vbmi implementation for N = 64.


calebzulawski:
Looks good to me! Thanks! FYI @workingjubilee

workingjubilee:
Thanks!

Issue #432, opened 2024-08-25
f64::reduce_sum::lanes_1 test fails on latest nightly due to negative zero


cvijdea-bd:
I tried this: `cargo test --test f64_ops`

I expected to see this happen: tests pass

Instead, this happened: 
```
---- f64::reduce_sum::lanes_1 stdout ----
thread 'f64::reduce_sum::lanes_1' panicked at /home/cvijdea/dev/portable-simd/crates/test_helpers/src/lib.rs:126:43:
called `Result::unwrap()` on an `Err` value: Fail(Reason("assertion failed: `(left == right)` \n  left: `0.0 (0)`,\n right: `-0.0 (8000000000000000)` at crates/core_simd/tests/f64_ops.rs:5"), [-0.0])
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- f64::reduce_sum::lanes_2 stdout ----
thread 'f64::reduce_sum::lanes_2' panicked at /home/cvijdea/dev/portable-simd/crates/test_helpers/src/lib.rs:126:43:
called `Result::unwrap()` on an `Err` value: Fail(Reason("assertion failed: `(left == right)` \n  left: `0.0 (0)`,\n right: `-0.0 (8000000000000000)` at crates/core_simd/tests/f64_ops.rs:5"), [-0.0, -0.0])
```

### Meta

`rustc --version --verbose`:
```
binary: rustc
commit-hash: 4074d4902dc19ff1bbce1d74ef9ceae15b5f41ce
commit-date: 2024-08-23
host: x86_64-unknown-linux-gnu
release: 1.82.0-nightly
LLVM version: 19.1.0
```

`nightly-2024-08-24` toolchain, portable-simd commit 4697d394138fddbf062dec154270f2a43c4cf64e

PR #438, opened 2024-09-12
Use -0.0 as the neutral additive float

workingjubilee:
-0.0 +  0.0 is  0.0
-0.0 + -0.0 is -0.0

Thus, the float additive-zero is -0.0, not its positive cousin. This aligns with a recent change to the impl of Sum for floats, in rust-lang/rust@490818851860fb257e23fe7aa0ee32eaffc4ba40

Fixes rust-lang/portable-simd#432

### DIFF
--- a/crates/core_simd/src/simd/num/float.rs
+++ b/crates/core_simd/src/simd/num/float.rs
@@ -419,7 +419,7 @@ macro_rules! impl_trait {
                     self.as_array().iter().sum()
                 } else {
                     // Safety: `self` is a float vector
-                    unsafe { core::intrinsics::simd::simd_reduce_add_ordered(self, 0.) }
+                    unsafe { core::intrinsics::simd::simd_reduce_add_ordered(self, -0.) }
                 }
             }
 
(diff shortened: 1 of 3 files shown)


calebzulawski:
I assume this resolves #432

workingjubilee:
yep, tests might require a bit of punching first.

calebzulawski:
I think you just need to bump rust-toolchain.toml

workingjubilee:
Oh, yep. That does it, thanks!

calebzulawski:
Pretty sure the arm failure is unrelated to your changes, looks like std might have added an extra step to is_normal that causes it to flush subnormals?

workingjubilee:
Oh yeah those tests are gonna be absolutely busted for thumbv7neon.

workingjubilee:
I've simply cfg'd those out for now.

workingjubilee:
oh, that doesn't work at all, heh.

workingjubilee:
There.

workingjubilee:
Yes, probably.

calebzulawski (approved):

PR #441, opened 2024-09-27
shift_elements_{left,right}

sammysheep:
This implementation was patterned after the pre-existing `rotate_elements_{left,right}` and is meant to supersede #112 from 2021. If this functionality is buried somewhere else in the API that I missed, please close with a note where I can find it.

The tests pass and the assembly looked reasonable on x86.

### DIFF
--- a/crates/core_simd/src/swizzle.rs
+++ b/crates/core_simd/src/swizzle.rs
@@ -255,7 +255,7 @@ where
     /// default value (e.g., zero) to the right.
     #[inline]
     #[must_use = "method returns a new vector and does not mutate the original inputs"]
-    fn shift_elements_left<const OFFSET: usize>(self) -> Self
+    pub fn shift_elements_left<const OFFSET: usize>(self) -> Self
     where
         T: Default,
     {
@@ -280,7 +280,7 @@ where
     /// default value (e.g., zero) from the left.
     #[inline]
     #[must_use = "method returns a new vector and does not mutate the original inputs"]
-    fn shift_elements_right<const OFFSET: usize>(self) -> Self
+    pub fn shift_elements_right<const OFFSET: usize>(self) -> Self
     where
         T: Default,
     {
--- a/crates/core_simd/src/swizzle.rs
+++ b/crates/core_simd/src/swizzle.rs
@@ -499,18 +499,30 @@ where
     /// `padding` from the right.
     #[inline]
     #[must_use = "method returns a new mask and does not mutate the original inputs"]
-    pub fn shift_elements_left<const OFFSET: usize>(self, padding: T) -> Self {
+    pub fn shift_elements_left<const OFFSET: usize>(self, padding: bool) -> Self {
         // Safety: swizzles are safe for masks
-        unsafe { Self::from_int_unchecked(self.to_int().shift_elements_left::<OFFSET>(padding)) }
+        unsafe {
+            Self::from_int_unchecked(self.to_int().shift_elements_left::<OFFSET>(if padding {
+                T::TRUE
+            } else {
+                T::FALSE
+            }))
+        }
     }
 
     /// Shifts the mask elements to the right by `OFFSET`, filling in with
     /// `padding` from the left.
     #[inline]
     #[must_use = "method returns a new mask and does not mutate the original inputs"]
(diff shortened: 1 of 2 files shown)


calebzulawski:
Not to say we _shouldn't_ have this function, but (devil's advocate) slices for example only have rotating elements, not shifting in default. You could technically use our already implemented SIMD rotates, and then another swizzle to insert the default values.

Should this maybe shift in an arbitrary value, rather than Default?

sammysheep:
Ah, this PR came about because I wrote an extension trait to do this for an algorithm that needs to shift by one element to the left and continue on its merry way. The C code I was porting just used a single intrinsic for what amounts to the same thing, so that motivated me to look for it in portable SIMD's API.

Perhaps also, adding an explicit _shift_elements_ function might help users to stop and realize that the `Shl` / `<<` operator works _within_ each element bitwise and not element-wise. This was an actual mistake I made a few months ago and only realized it after restarting my port project again yesterday. You /can/ use rotate++ as you suggest, but I wrote the extension trait exactly because my doing so would have felt like the wrong level of abstraction.

As to using `default()`, the previous impl in #112 inspired me that it was an easy and natural way to provide zero-padding. However, I'm not stuck on that. If one were to make the pad value arbitrary, would it just need to be another `const` parameter I suppose?

If so, would that rule out things that are invalid as a const generic parameter like `f64`?

Happy to make whatever desired changes based upon your own design goals if you decide to keep this PR.

programmerjake:
> However, I'm not stuck on that. If one were to make the pad value arbitrary, would it just need to be another const parameter I suppose?

no, it can just be a function argument, no need for it to be const.

sammysheep:
I have changed it to accept a `padding` argument and updated the tests and docs. This seems equivalent for the zero case on x86_64: https://godbolt.org/z/E8h9onese

crates/core_simd/src/swizzle.rs
```diff
@@ -451,6 +495,24 @@ where
         unsafe { Self::from_int_unchecked(self.to_int().rotate_elements_right::<OFFSET>()) }
     }
 
+    /// Shifts the mask elements to the left by `OFFSET`, filling in with
+    /// `padding` from the right.
+    #[inline]
+    #[must_use = "method returns a new vector and does not mutate the original inputs"]
+    pub fn shift_elements_left<const OFFSET: usize>(self, padding: T) -> Self {
```
calebzulawski, line 502: I think `padding` should be `bool` for masks

crates/core_simd/src/swizzle.rs
```diff
@@ -451,6 +495,24 @@ where
         unsafe { Self::from_int_unchecked(self.to_int().rotate_elements_right::<OFFSET>()) }
     }
 
+    /// Shifts the mask elements to the left by `OFFSET`, filling in with
+    /// `padding` from the right.
+    #[inline]
+    #[must_use = "method returns a new vector and does not mutate the original inputs"]
```
calebzulawski, line 501:
```suggestion
    #[must_use = "method returns a new mask and does not mutate the original inputs"]
```

crates/core_simd/src/swizzle.rs
```diff
@@ -451,6 +495,24 @@ where
...
+    pub fn shift_elements_left<const OFFSET: usize>(self, padding: T) -> Self {
+        // Safety: swizzles are safe for masks
+        unsafe { Self::from_int_unchecked(self.to_int().shift_elements_left::<OFFSET>(padding)) }
+    }
+
+    /// Shifts the mask elements to the right by `OFFSET`, filling in with
+    /// `padding` from the left.
+    #[inline]
+    #[must_use = "method returns a new vector and does not mutate the original inputs"]
```
calebzulawski, line 510:
```suggestion
    #[must_use = "method returns a new mask and does not mutate the original inputs"]
```

calebzulawski (changes requested):

crates/core_simd/src/swizzle.rs
sammysheep: I added this change but the conversion felt awkward. LMK if there is a more idiomatic way.

calebzulawski: I think that's the way to go

calebzulawski (approved):

programmerjake (approved):

PR #443, opened 2024-10-07
Add armv7 neon mplementation for  `Simd<u8, 16>::swizzle_dyn`

hkratz:
Use arm neon intrinsics to swizzle two u8x8 blocks with a u8x8x2 lookup table.


workingjubilee:
Oh, thanks!

PR #444, opened 2024-11-02
ci: add support for loongarch64-unknown-linux-gnu

heiher:


programmerjake (approved):

.github/workflows/ci.yml
```diff
@@ -216,15 +218,9 @@ jobs:
...
-        # binary. Ideally we wouldn't hardcode a version, but the version number
-        # being part of the tarball means we can't just use the download/latest
-        # URL :(
+        # Install the latest git version for newer targets.
         run: |
-          CROSS_URL=https://github.com/cross-rs/cross/releases/download/v0.2.5/cross-x86_64-unknown-linux-gnu.tar.gz
-          mkdir -p "$HOME/.bin"
-          curl -sfSL --retry-delay 10 --retry 5 "${CROSS_URL}" | tar zxf - -C "$HOME/.bin"
-          echo "$HOME/.bin" >> $GITHUB_PATH
+          cargo install cross --git https://github.com/cross-rs/cross
```
programmerjake, line 223: tbh, installing dependencies from latest git has its downsides, e.g. our CI breaks if they push a broken commit.

the comments wanted to install the latest released version from crates.io, which should be more stable...but loongarch64 isn't in a released version yet. if we wanted to avoid breaks, we could pick a particular git commit to use instead of just latest master.

heiher: Done.

programmerjake (approved):

PR #446, opened 2024-11-25
add rustc_const_unstable attribute to Simd::splat

### LINKED ISSUE #445: Compiler error at `Simd::splat`: usage of `intrinsics::const_eval_select`
The latest nightly compiler gives the following error when compiling the `core_simd` crate from this repository (not `core::simd`):

```
error: const function that might be (indirectly) exposed to stable cannot use `#[feature(const_eval_select)]`
   --> prtable-simd/crates/core_simd/src/vector.rs:171:9
    |
171 |         core::intrinsics::const_eval_select((value,), splat_const, splat_rt)
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
help: if the function is not (yet) meant to be exposed to stable, add `#[rustc_const_unstable]` (this is what you probably want to do)
    |
147 +     #[rustc_const_unstable(feature = "...", issue = "...")]
148 |     pub const fn splat(value: T) -> Self {
    |
help: otherwise, as a last resort `#[rustc_allow_const_fn_unstable]` can be used to bypass stability checks (this requires team approval)
    |
147 +     #[rustc_allow_const_fn_unstable(const_eval_select)]
148 |     pub const fn splat(value: T) -> Self {
    |
```

A quick blame suggests that this is caused by #417

Possbile fixes (some already suggested by the compiler):

 - Using `#[rustc_const_unstable(feature = "...", issue = "...")]` (Which feature and tracking issue would be appropriate here?)
 - Using `#[rustc_allow_const_fn_unstable(const_eval_select)]`
 - Reverting #417 completely

AquaEBM:
Fixes #445

### DIFF
--- a/crates/core_simd/src/vector.rs
+++ b/crates/core_simd/src/vector.rs
@@ -144,6 +144,7 @@ where
     /// assert_eq!(v.as_array(), &[8, 8, 8, 8]);
     /// ```
     #[inline]
+    #[rustc_const_unstable(feature = "portable_simd", issue = "86656")]
     pub const fn splat(value: T) -> Self {
         const fn splat_const<T, const N: usize>(value: T) -> Simd<T, N>
         where


calebzulawski:
Thank you!

PR #448, opened 2025-01-16
Sync upstream

calebzulawski:
Apparently rebases don't really work with git subtree, which is what I was using in the past. I did a merge this time, which had better conflict resolution. (we should really move to josh, if I can ever get a chance to figure it out...)


programmerjake (approved):
lgtm, assuming you did the git subtree ops correctly

PR #451, opened 2025-02-12
Migrate core_simd to Rust 2024

ehuss:
This migrates core_simd to Rust 2024. The intent here is to ensure that when embedded in core that there aren't any regressions when we move core to 2024 (since it is embedded with `#[path]`). Otherwise this shouldn't have any impact.

### DIFF
--- a/crates/core_simd/Cargo.toml
+++ b/crates/core_simd/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "core_simd"
 version = "0.1.0"
-edition = "2021"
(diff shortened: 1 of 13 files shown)


programmerjake (approved):
looks good to me, though I'm not sure what the plan is on when to move to edition 2024? maybe @workingjubilee knows?

workingjubilee:
This should be fine.

programmerjake:
> This should be fine.

so then just merge now?

PR #452, opened 2025-03-02
add examples for `shift_elements_{left, right}` and `rotate_elements_{left, right}`

folkertdev:
I think the text is just complicated enough to warrant an example. The doctest also shows the wrapping behavior of the implementation (for standard integers, shifting by more bits than the type has is UB (c) or panics (rust)).

Maybe the modulo behavior should be called out more explicitly still?

### DIFF
--- a/crates/core_simd/src/swizzle.rs
+++ b/crates/core_simd/src/swizzle.rs
@@ -214,6 +214,17 @@ where
     /// Rotates the vector such that the first `OFFSET` elements of the slice move to the end
     /// while the last `self.len() - OFFSET` elements move to the front. After calling `rotate_elements_left`,
     /// the element previously at index `OFFSET` will become the first element in the slice.
+    /// ```
+    /// # #![feature(portable_simd)]
+    /// # #[cfg(feature = "as_crate")] use core_simd::simd::Simd;
+    /// # #[cfg(not(feature = "as_crate"))] use core::simd::Simd;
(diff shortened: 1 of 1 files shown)


folkertdev:
Looks like CI is using an older version of rustc? I can reproduce the failures with `cargo test` but they disappear with `cargo +nightly test`.

crates/core_simd/src/swizzle.rs
```diff
@@ -214,6 +214,16 @@ where
     /// Rotates the vector such that the first `OFFSET` elements of the slice move to the end
     /// while the last `self.len() - OFFSET` elements move to the front. After calling `rotate_elements_left`,
     /// the element previously at index `OFFSET` will become the first element in the slice.
+    /// ```
+    /// # #![feature(portable_simd)]
+    /// # use core::simd::Simd;
```
calebzulawski, line 219: You probably need to reference the actual crate like this:
https://github.com/rust-lang/portable-simd/blob/abd5d05ec4de9eb11fd3dfd84cfc1c6e75fb5eba/crates/core_simd/src/vector.rs#L122-L123

calebzulawski (changes requested):

crates/core_simd/src/swizzle.rs
folkertdev: yup, that did it, thanks!

calebzulawski:
Thanks!

PR #456, opened 2025-03-19
Rust upstream 2025-03-19

calebzulawski:


calebzulawski:
hmm that's not right

calebzulawski:
I'll try again after rust-lang/rust#138687

PR #465, opened 2025-06-12
Add const to `as_mut_array`, `copy_to_slice`

clarfonthey:
These were never marked as const when `const_mut` was stabilised, and pass all tests when marked as const.


calebzulawski:
Thanks!

PR #466, opened 2025-06-13
Make `Mask::splat` const

clarfonthey:
This is a small step toward making more methods `const`.

While removing the `Default` bound was not *strictly* necessary, I found that it made sense to do given the implementation. Better to mention that it's not used here in the bounds than leave it in.

There are probably other mask methods that could be made const with minimal changes, but I didn't want to touch anything that used intrinsics and/or litter everything with `const_eval_select` to make it work.

### DIFF
--- a/crates/core_simd/src/lane_count.rs
+++ b/crates/core_simd/src/lane_count.rs
@@ -18,7 +18,11 @@ impl<const N: usize> LaneCount<N> {
 /// Only SIMD vectors with supported lane counts are constructable.
 pub trait SupportedLaneCount: Sealed {
     #[doc(hidden)]
-    type BitMask: Copy + Default + AsRef<[u8]> + AsMut<[u8]>;
+    type BitMask: Copy + AsRef<[u8]> + AsMut<[u8]>;
+    #[doc(hidden)]
+    const EMPTY_BIT_MASK: Self::BitMask;
+    #[doc(hidden)]
+    const FULL_BIT_MASK: Self::BitMask;
 }
 
 impl<const N: usize> Sealed for LaneCount<N> {}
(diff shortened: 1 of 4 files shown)


crates/core_simd/src/lane_count.rs
```diff
@@ -28,6 +32,15 @@ macro_rules! supported_lane_count {
         $(
             impl SupportedLaneCount for LaneCount<$lanes> {
                 type BitMask = [u8; ($lanes + 7) / 8];
+                const EMPTY_BIT_MASK: Self::BitMask = [u8::MIN; ($lanes + 7) / 8];
+                const FULL_BIT_MASK: Self::BitMask = {
+                    const LEN: usize = ($lanes + 7) / 8;
+                    let mut array = [u8::MAX; LEN];
```
programmerjake, line 38:
```suggestion
                const EMPTY_BIT_MASK: Self::BitMask = [0; ($lanes + 7) / 8];
                const FULL_BIT_MASK: Self::BitMask = {
                    const LEN: usize = ($lanes + 7) / 8;
                    let mut array = [!0u8; LEN];
```

imo literals are clearer since we care about which bits are set, rather than whatever the maximum/minimum is.

clarfonthey: I'm fine with that and agree: I mostly just copied from the existing code which preferred using `u8::MIN` and `u8::MAX`.

clarfonthey: Also following up on this, I also changed the `u8::MAX` below to be `!0`, and also the `&=` to just `=` since it's a bit clearer.

crates/core_simd/src/masks/bitmask.rs
```diff
@@ -78,17 +78,16 @@ where
...
-        if value {
-            mask.as_mut().fill(u8::MAX)
-        } else {
-            mask.as_mut().fill(u8::MIN)
-        }
-        if N % 8 > 0 {
-            *mask.as_mut().last_mut().unwrap() &= u8::MAX >> (8 - N % 8);
-        }
-        Self(mask, PhantomData)
+    #[rustc_const_unstable(feature = "portable_simd", issue = "86656")]
```
calebzulawski, line 81: Since this has never been stable, what's the attribute for?

clarfonthey: I'm not 100% sure why, but it's required. My guess is since it's a constructor for an unstable type instead of just a method on it.

PR #480, opened 2025-09-08
swizzle_dyn: 64 byte swizzle_dyn for AVX2

87flowers:
Implemented a 64 bit `swizzle_dyn` for AVX2.

Use case: Encountered this while [implementing a chess move generator](<https://87flowers.com/byteboard-attack-tables-1/>) in rust with portable simd.

Would like an alternative suggestion for use of `mem::transmute` here.


Shnatsel:
FWIW https://github.com/rust-lang/portable-simd/pull/540 applied a similar operation generically, for all cases where swizzle_dyn operates on double the native vector width.

87flowers:
Thanks!

PR #483, opened 2025-09-15
Mask element v2

calebzulawski:
#322 redux

Arguments for this change:
* Integer elements impose a very specific view of masks (AVX512, SVE, RVV, etc use bitmasks). `Mask<i32, 4>` over `Simd<f32, 4>` only maps to one type of implementation. `Mask<f32, 4>` just means "a mask over f32s"
* Some API surface is reduced: `T::Mask` disappears from everywhere except `Mask::to_simd`/`Mask::from_simd`. `MaskElement` is removed from the public API.

This comment from @thomcc summarizes the arguments against this change in the previous PR: https://github.com/rust-lang/portable-simd/pull/322#issuecomment-1464710468
* I think the first comment regarding API friction is no longer relevant, as this PR includes `From` implementations for all masks. Additionally, `Select` was changed in #482 to allow any mask to select any vector as long as it has the same number of elements. Some code will require slightly more calls `into()`/`cast()` in some situations but otherwise there should be minimal overhead.
* Regarding the second comment about complicated types and projections, this PR doesn't really make this worse anywhere, and probably makes it better in a lot of places, e.g. you can make a function generic over `T: SimdElement` and that will also be your mask type.

### DIFF
--- a/crates/core_simd/src/masks.rs
+++ b/crates/core_simd/src/masks.rs
@@ -114,7 +114,7 @@ impl_element! { i32, u32 }
 impl_element! { i64, u64 }
 impl_element! { isize, usize }
 
-/// A SIMD vector mask for `N` elements of width specified by `Element`.
+/// A SIMD vector mask for `N` elements matching the element type `T`.
 ///
 /// Masks represent boolean inclusion/exclusion on a per-element basis.
 ///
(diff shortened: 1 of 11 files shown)


programmerjake (approved):

programmerjake:
ping @workingjubilee -- it's been a while, are you ready to post your review yet?

programmerjake:
> ping @workingjubilee -- it's been a while, are you ready to post your review yet?

it's been another several months. maybe you'll be ready soon?

workingjubilee:
HELLO YES.

PR #487, opened 2025-11-05
Implement `SimdElement` for `f16`

sayantn:
After rust-lang/rust#147520, now we can use safely use f16 in portable SIMD.

TODO: we cannot test it because `proptest` and `float_cmp` don't support `f16` yet. @workingjubilee what should I do in this case? I can send a PR to them


folkertdev:
This has now been subsumed by https://github.com/rust-lang/portable-simd/pull/513 I think.

PR #492, opened 2025-11-16
Add mul_sub() fused multiply-add operation

GrigoryEvko:
This PR adds and `mul_sub()` methods to the `SimdFloat` trait. This operation compute `(self * a) - b` respectively, using a single fused multiply-add instruction where available.


GrigoryEvko:
Maybe move mul_add from std_float::StdFloat to core_simd/src/simd/num/float.rs as I accidentaly tried?

PR #493, opened 2025-11-27
Implement traits over scalars with `self: Simd<T, N>`

calebzulawski:
This PR has been overhauled a little bit because we decided to try using `self: Simd<T, N>`.

Original comments:

This improves type inference by removing most associated types. For example, the following function doesn't actually work:
```rust
fn foo<T: SimdElement>(x: Simd<T, 4>) -> Mask<T::Mask, 4>
where
    Simd<T, 4>: SimdFloat,
{
    x.is_sign_negative()
}
```
You actually need the constraint `Simd<T, 4>: SimdFloat<Mask = Mask<T::Mask, 4>>`.

With this PR, the function should look a little more like:
```rust
fn foo<T: SimdElement>(x: Simd<T, 4>) -> Mask<T::Mask, 4>
where
    Simd<T, 4>: SimdFloat<T, 4>,
{
    x.is_sign_negative()
}
```
There are still a few instances that need associated types, but we can explore those individually to see if there's a better API.

Also, note that we now end up with a lot of instances of `Mask<<T as SimdElement>::Mask, N>`. We can either remove the `SimdElement::Mask` associated type with rust-lang/portable-simd#483 or we can provide an alias to avoid writing that out.

### DIFF
--- a/crates/core_simd/src/simd/cmp/eq.rs
+++ b/crates/core_simd/src/simd/cmp/eq.rs
@@ -4,35 +4,32 @@ use crate::simd::{
 };
 
 /// Parallel `PartialEq`.
-pub trait SimdPartialEq {
-    /// The mask type returned by each comparison.
-    type Mask;
-
+pub trait SimdPartialEq<T, const N: usize>
+where
+    T: SimdElement,
+{
     /// Test if each element is equal to the corresponding element in `other`.
     #[must_use = "method returns a new mask and does not mutate the original value"]
-    fn simd_eq(self, other: Self) -> Self::Mask;
+    fn simd_eq(self, other: Self) -> Mask<<T as SimdElement>::Mask, N>;
 
     /// Test if each element is not equal to the corresponding element in `other`.
     #[must_use = "method returns a new mask and does not mutate the original value"]
-    fn simd_ne(self, other: Self) -> Self::Mask;
+    fn simd_ne(self, other: Self) -> Mask<<T as SimdElement>::Mask, N>;
 }
 
 macro_rules! impl_number {
     { $($number:ty),* } => {
         $(
-        impl<const N: usize> SimdPartialEq for Simd<$number, N>
-        {
-            type Mask = Mask<<$number as SimdElement>::Mask, N>;
-
+        impl<const N: usize> SimdPartialEq<$number, N> for Simd<$number, N> {
             #[inline]
-            fn simd_eq(self, other: Self) -> Self::Mask {
+            fn simd_eq(self, other: Self) -> Mask<<$number as SimdElement>::Mask, N> {
                 // Safety: `self` is a vector, and the result of the comparison
                 // is always a valid mask.
                 unsafe { Mask::from_simd_unchecked(core::intrinsics::simd::simd_eq(self, other)) }
             }
 
             #[inline]
-            fn simd_ne(self, other: Self) -> Self::Mask {
+            fn simd_ne(self, other: Self) -> Mask<<$number as SimdElement>::Mask, N> {
                 // Safety: `self` is a vector, and the result of the comparison
                 // is always a valid mask.
                 unsafe { Mask::from_simd_unchecked(core::intrinsics::simd::simd_ne(self, other)) }
@@ -47,19 +44,16 @@ impl_number! { f16, f32, f64, u8, u16, u32, u64, usize, i8, i16, i32, i64, isize
 macro_rules! impl_mask {
     { $($integer:ty),* } => {
         $(
-        impl<const N: usize> SimdPartialEq for Mask<$integer, N>
-        {
-            type Mask = Self;
-
+        impl<const N: usize> SimdPartialEq<$integer, N> for Mask<$integer, N> {
             #[inline]
-            fn simd_eq(self, other: Self) -> Self::Mask {
+            fn simd_eq(self, other: Self) -> Mask<<$integer as SimdElement>::Mask, N> {
                 // Safety: `self` is a vector, and the result of the comparison
                 // is always a valid mask.
                 unsafe { Self::from_simd_unchecked(core::intrinsics::simd::simd_eq(self.to_simd(), other.to_simd())) }
             }
 
             #[inline]
-            fn simd_ne(self, other: Self) -> Self::Mask {
+            fn simd_ne(self, other: Self) -> Mask<<$integer as SimdElement>::Mask, N> {
                 // Safety: `self` is a vector, and the result of the comparison
                 // is always a valid mask.
                 unsafe { Self::from_simd_unchecked(core::intrinsics::simd::simd_ne(self.to_simd(), other.to_simd())) }
@@ -71,30 +65,26 @@ macro_rules! impl_mask {
 
 impl_mask! { i8, i16, i32, i64, isize }
 
-impl<T, const N: usize> SimdPartialEq for Simd<*const T, N> {
-    type Mask = Mask<isize, N>;
-
+impl<T, const N: usize> SimdPartialEq<*const T, N> for Simd<*const T, N> {
     #[inline]
-    fn simd_eq(self, other: Self) -> Self::Mask {
+    fn simd_eq(self, other: Self) -> Mask<isize, N> {
         self.addr().simd_eq(other.addr())
     }
 
     #[inline]
-    fn simd_ne(self, other: Self) -> Self::Mask {
+    fn simd_ne(self, other: Self) -> Mask<isize, N> {
         self.addr().simd_ne(other.addr())
     }
 }
 
-impl<T, const N: usize> SimdPartialEq for Simd<*mut T, N> {
-    type Mask = Mask<isize, N>;
-
+impl<T, const N: usize> SimdPartialEq<*mut T, N> for Simd<*mut T, N> {
     #[inline]
-    fn simd_eq(self, other: Self) -> Self::Mask {
+    fn simd_eq(self, other: Self) -> Mask<isize, N> {
         self.addr().simd_eq(other.addr())
     }
 
     #[inline]
-    fn simd_ne(self, other: Self) -> Self::Mask {
+    fn simd_ne(self, other: Self) -> Mask<isize, N> {
         self.addr().simd_ne(other.addr())
     }
 }
--- a/crates/core_simd/src/simd/cmp/eq.rs
+++ b/crates/core_simd/src/simd/cmp/eq.rs
@@ -4,32 +4,41 @@ use crate::simd::{
 };
 
 /// Parallel `PartialEq`.
-pub trait SimdPartialEq<T, const N: usize>
-where
-    T: SimdElement,
-{
+pub trait SimdPartialEq: SimdElement {
     /// Test if each element is equal to the corresponding element in `other`.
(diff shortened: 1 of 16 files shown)


crates/core_simd/src/simd/cmp/eq.rs
```diff
@@ -4,35 +4,32 @@ use crate::simd::{
 };
 
 /// Parallel `PartialEq`.
-pub trait SimdPartialEq {
-    /// The mask type returned by each comparison.
-    type Mask;
-
+pub trait SimdPartialEq<T, const N: usize>
```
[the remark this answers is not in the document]
programmerjake: further reduced feature gates: https://play.rust-lang.org/?version=nightly&mode=debug&edition=2024&gist=b57a420a75142e3eda48993dc2f981f1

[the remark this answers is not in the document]
programmerjake: note the message you quoted is for achieving a similar effect while specifically avoiding mGCA:

> > it also has just as nice ergonomics when the length is a const generic: https://play.rust-lang.org/?version=nightly&mode=debug&edition=2024&gist=29fba8f75be0e771bb5fd580e77f9c89
> 
> Having been playing around with mGCA. I will say it does allow for very nice API via associated consts, and combine that with associated const equality. You can create very nice trait interfaces that just need to specify one trait bound for a generic function. Where as putting the `<const N: usize>` in your traits works on stable, but you end having to carry that N everywhere when you start working generically over some a set of types.

if you're trying to be generic over multiple lengths, of course you need a `const N: usize` generic. if you instead have one or more known lengths that you're not generic over, I intended that you shouldn't need bounds mentioning generic lengths.

> @programmerjake For stable I have something like this as a backbone of something I have been working on to avoid specifying N every where since associated types are stable, and I can also perform associated type equality on constraints on types. https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=8190a2e8e45e7dbbbbaabc90c75c3c74

the `Inc`/`Dec` parts seems very much like [`typenum`](https://docs.rs/typenum/1.19.0/typenum/) which I think is something Rust should avoid in stable interfaces in the standard library since I think Rust should have generic const expressions for that kind of thing instead of using types to do arithmetic. Also, the finite list of impls for specific integers is what we just removed in #485.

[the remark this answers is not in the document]
Keith-Cancel: > > if you're trying to be generic over multiple lengths, of course you need a `const N: usize` generic. if you instead have one or more known lengths that you're not generic over, I intended that you shouldn't need bounds mentioning generic lengths.
> 
> imo `Simd<T, N>` ideally shouldn't need more or less trait bounds for lengths than `[T; N]`, so if you're trying to be generic over length having `fn f<const N: usize>(v: Simd<f32, N>)` seems good to me. imo you shouldn't need to mention known or generic lengths anywhere other than the `N` in `Simd<T, N>` (or `Mask`), so I don't like having to have `Simd<T, N>: SomeTrait<N> + OtherTrait<N>` where you have to repeat the `N` in all the trait bounds.

Yea but an associated `#[type_const]` would allow such a function to just be `fn f<T: Simd<f32>>(v: T)`.  

-Also if you need constraint the length you could do something like:
`fn f<T: Simd<f32, N = 4>>(v: T)` ect..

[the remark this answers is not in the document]
programmerjake: > Yea but an associated `#[type_const]` would allow such a function to just be `fn f<T: Simd<f32>>(v: T)`.

Yes (if you used an actual trait rather than the `Simd` type). That said, I think using the `Simd<f32, N>` type is waay better, similar to how it's standard practice to write  `fn f<const N: usize>(v: [f32; N])` instead of `fn f<T: ArrayOf<f32>>(v: T)`.

> -Also if you need constraint the length you could do something like: `fn f<T: Simd<f32, N = 4>>(v: T)` ect..

if you already know both the length and element type then there's only one `Simd` type that matches, using a generic is just extra complexity for no reason, just use `Simd<f32, 4>` or the `f32x4` type alias. this is similar to how you'd write `[f32; 4]` instead of `T: Array<f32, 4>`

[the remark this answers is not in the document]
programmerjake: we spawned a discussion: https://rust-lang.zulipchat.com/#narrow/channel/257879-project-portable-simd/topic/Simd.20traits.20and.20assoc.2E.20.60type.20const.60.20vs.2E.20other.20API.20designs/near/592861697

programmerjake:
after thinking some more, I think I found the best API so far: https://rust-lang.zulipchat.com/#narrow/channel/257879-project-portable-simd/topic/Simd.3CT.2C.20N.3E.3A.20Receiver.3CTarget.20.3D.20T.3E/with/593431978

crates/core_simd/src/simd/num/float.rs
```diff
@@ -57,102 +47,110 @@ pub trait SimdFloat: Copy + Sealed {
...
     /// same size and number of elements.
     #[must_use = "method returns a new vector and does not mutate the original value"]
-    fn to_bits(self) -> Self::Bits;
+    fn to_bits<const N: usize>(self: Simd<Self, N>) -> Simd<Self::Bits, N>;
 
     /// Raw transmutation from an unsigned integer vector type with the
     /// same size and number of elements.
     #[must_use = "method returns a new vector and does not mutate the original value"]
-    fn from_bits(bits: Self::Bits) -> Self;
+    fn from_bits<const N: usize>(bits: Simd<Self::Bits, N>) -> Simd<Self, N>;
```
programmerjake: maybe we should rename this to `simd_from_bits` or something so you can use it as `f32::simd_from_bits(v)` instead of having to write `<f32 as SimdFloat>::from_bits(v)`. `SimdFloat::from_bits(v)` only works when something else constrains the return type to `Simd<f32, N>`. `f32::from_bits(v)` doesn't work since that gives the scalar method.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/portable-simd/493/cc032ad5ed5b69d606c4eabbc9c9ae1f412ef2fb..f935cb3739f6227091d76ee86b736f5c62daafe2)*

crates/core_simd/src/simd/num/float.rs
```diff
@@ -261,141 +254,144 @@ macro_rules! impl_trait {
             // tracked in: https://github.com/rust-lang/rust/issues/135982
```
programmerjake: this workaround is no longer necessary since LLVM 20 was already made the minimum version a while ago, maybe remove it as a drive-by fix?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/portable-simd/493/cc032ad5ed5b69d606c4eabbc9c9ae1f412ef2fb..f935cb3739f6227091d76ee86b736f5c62daafe2)*

crates/core_simd/src/simd/ptr/const_ptr.rs
```diff
@@ -1,35 +1,25 @@
...
-    fn is_null(self) -> Self::Mask;
+    fn is_null<const N: usize>(self: Simd<Self, N>) -> Mask<isize, N>;
 
     /// Casts to a pointer of another type.
     ///
     /// Equivalent to calling [`pointer::cast`] on each element.
-    fn cast<T>(self) -> Self::CastPtr<T>;
+    fn cast<U, const N: usize>(self: Simd<Self, N>) -> Simd<*const U, N>
+    where
+        U: core::ptr::Pointee<Metadata = ()>;
```
programmerjake: why have the `Pointee<Metadata = ()>` bound when that's already implied by `Sized`? or did you just forget the `?Sized` bound here and on `Self::Pointee`?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/portable-simd/493/cc032ad5ed5b69d606c4eabbc9c9ae1f412ef2fb..f935cb3739f6227091d76ee86b736f5c62daafe2)*

crates/core_simd/src/simd/num/float.rs
calebzulawski: Yeah, something like that probably makes sense for any free functions

crates/core_simd/src/simd/ptr/const_ptr.rs
calebzulawski: Hmm, that's a good point. I guess technically extern types are unsized and still have thin pointers if we want to add `?Sized`, but it's probably better to just assume `Sized`.

crates/core_simd/src/simd/num/float.rs
```diff
@@ -234,168 +232,138 @@ pub trait SimdFloat: Copy + Sealed {
...
 
             #[inline]
-            fn copysign(self, sign: Self) -> Self {
-                let sign_bit = sign.to_bits() & Self::splat(-0.).to_bits();
-                let magnitude = self.to_bits() & !Self::splat(-0.).to_bits();
-                Self::from_bits(sign_bit | magnitude)
+            fn copysign<const N: usize>(self: Simd<Self, N>, sign: Simd<Self, N>) -> Simd<Self, N> {
+                let sign_bit = sign.to_bits() & Simd::splat(-0. as $ty).to_bits();
+                let magnitude = self.to_bits() & !Simd::splat(-0. as $ty).to_bits();
+                <Self as SimdFloat>::simd_from_bits(sign_bit | magnitude)
```
programmerjake:
```suggestion
                Self::simd_from_bits(sign_bit | magnitude)
```

now that it's renamed, the above should work.

programmerjake (commented):
generally looks good to me, though we'll want to wait on https://github.com/rust-lang/rust/issues/156270 before merging.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/portable-simd/493/94a3ba2675fb37f5683323a7cfd29a888d737f23..e9c3147ca83dff9570fc0ad593f9ad41be1f6696)*

workingjubilee (approved):
So this looks fine and my only primary question atm is the detail of why the `const N` is in every function instead of a generic as part of the trait definition _per se_. Though as I believe has been mentioned it might be better off as an associated constant of sorts but we don't want to expose that immediately because of mGCA stability concerns?

My primary question probably has an answer that has already been converged on and I missed it in the discussion, apologies.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/portable-simd/493/94a3ba2675fb37f5683323a7cfd29a888d737f23..71b9f6c90fe9450935440ab68a9fd35930069e1e)*

calebzulawski:
> So this looks fine and my only primary question atm is the detail of why the `const N` is in every function instead of a generic as part of the trait definition _per se_. Though as I believe has been mentioned it might be better off as an associated constant of sorts but we don't want to expose that immediately because of mGCA stability concerns?
> 
> My primary question probably has an answer that has already been converged on and I missed it in the discussion, apologies.

In short having it on the function means you can do `T: SimdUint` and then use `Simd<T, 4>` and `Simd<T, 8>` without introducing new bounds, because "being able to be put in a Simd" is a characteristic of the scalar. It's not strictly necessary, but it will be easier when we get to things like `<T as SimdFloat>::Bits` being independent of the vector size.

PR #498, opened 2025-12-22
Remove `Simd: Index`

scottmcm:
(Inspired by this conversation on Discord <https://discord.com/channels/273534239310479360/592856094527848449/1452407229843247197>)

This PR removes the `Simd<_, _>: Index` and `Simd<_, _>: IndexMut` trait implementations, and instead adds `get` and `set` methods.

IMHO this is a good idea as `v[1] = 2;` looks innocuous but is actually surprisingly bad for codegen, as it forces the vector to be in-memory in order to return the `&T` -- and there's no way to return a `T` from `Index`.  One *can* always `v.as_mut_array()[1] = 2;` if you want (that still *works*) but that extra speed bump is, I posit, a good thing to help push people to better methods.

What are those better methods?  Well, this also adds `Simd::set` and `Simd::get` using the `simd_{insert,extract}_dyn` intrinsics internally.  Thus instead of `let x = v[i];`, you'd write `let x = v.get(i);`.  Doing that solves the optimization problem that the person was asking about on discord, and seeing that it did fix it (and that there's no use of `simd_extract_dyn` in the current library) is why I went and made this PR 🙂 

Discussion topics:
- Should `Index(Mut)` be left there for a bit, to help smooth the transition?  Can we even mark `impl`s as deprecated?
- What should the names be?  I like `set`/`get` for shortness, but maybe something domain-specific would be better.  I considered using `insert`/`extract`, but <https://doc.rust-lang.org/std/simd/struct.Simd.html#method.extract> already exists.  (It could be `extract_vector` or something, though.)
- Should `set` be `&mut self` or `self`?  I noticed that <https://doc.rust-lang.org/std/simd/struct.Mask.html#method.set> is `&mut self`, but using `&mut self` also forces it through memory in codegen, albeit in a way much easier for the optimizer to remove.

### DIFF
--- a/crates/core_simd/src/vector.rs
+++ b/crates/core_simd/src/vector.rs
@@ -218,6 +218,11 @@ where
     /// `idx` must be in-bounds (`idx < N`)
     #[inline]
     unsafe fn get_inner(self, idx: usize) -> T {
+        // FIXME: This is a workaround for a CI failure in #498
+        if cfg!(target_family = "wasm") {
+            return self.as_array()[idx];
+        }
+
         // SAFETY: our precondition is also that the value is in-bounds
         // and this type is a simd type of the correct element type.
         unsafe { core::intrinsics::simd::simd_extract_dyn(self, idx as u32) }
@@ -281,6 +286,13 @@ where
     #[inline]
     #[must_use = "This returns a new vector, rather than updating in-place"]
     unsafe fn set_inner(self, idx: usize, val: T) -> Self {
+        // FIXME: This is a workaround for a CI failure in #498
+        if cfg!(target_family = "wasm") {
+            let mut temp = self;
+            temp.as_mut_array()[idx] = val;
+            return temp;
+        }
+
         // SAFETY: our precondition is also that the value is in-bounds
         // and this type is a simd type of the correct element type.
         unsafe { core::intrinsics::simd::simd_insert_dyn(self, idx as u32, val) }
--- a/crates/core_simd/src/vector.rs
+++ b/crates/core_simd/src/vector.rs
@@ -173,6 +173,11 @@ where
 
     /// Returns an array reference containing the entire SIMD vector.
     ///
+    /// While this exists and *can* be used to read an arbitrary element from a
+    /// vector, like `v.as_array()[i]`, that's discouraged because it forces the
+    /// vector to memory which tends to perform poorly.  Instead, use
+    /// [`Self::get`].  (This is also why `Index` is not implemented.)
+    ///
     /// # Examples
     ///
     /// ```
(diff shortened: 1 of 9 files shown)


workingjubilee:
Yes, both? Fucking hell.

workingjubilee:
I don't see how anything that is "In My Opinion" matters anyways unless you have an opinion in the exact shape of certain optimization passes OR are willing to write an optimization pass in the shape of your opinions. We should be trying to get results not speculation.

workingjubilee:
Ah. We actually do represent scalars "directly" in LLVMIR.

I was primarily looking to make the function-crossing of vectors be represented directly in LLVMIR (where ABI compatible). That was my main goal, as it enables optimizations on its own. Adjusting things to represent vectors directly in the intra-function IR is also possible, easier than it might sound, and kind of almost required once you have vectors pass directly.

Basically just a lot of things have not been thoroughly examined and are only a few small touches away. The main thing is just that previous attempts to improve things have not been thoughtful in how they adjusted, more like blind stabs in the dark. They wind up requiring reverts too often. I'm trying to at least get a candle, maybe a few, shining on the issue, so that I can make something stickier.

workingjubilee:
My apologies for excess heat. I was nervy and rattled and having a bad time for unrelated reasons. I think I should have taken more time in replying to... think about word choice, if nothing else.

Are we still game to accept this PR?

crates/core_simd/src/vector.rs
```diff
@@ -204,6 +214,147 @@ where
...
+            // SAFETY: just checked in-bounds with the if
+            Some(unsafe { self.get_inner(idx) })
+        } else {
+            None
+        }
+    }
+
+    /// Gets the value at `idx` to `val`.
+    ///
+    /// This typically faster than reading via `as_array`.
```
programmerjake, line 271:
```suggestion
    /// This is typically faster than reading an element via `self.as_array()[idx]`.
```

also for the other occurrences

crates/core_simd/src/vector.rs
```diff
@@ -204,6 +214,147 @@ where
...
+        }
+
+        // SAFETY: our precondition is also that the value is in-bounds
+        // and this type is a simd type of the correct element type.
+        unsafe { core::intrinsics::simd::simd_insert_dyn(self, idx as u32, val) }
+    }
+
+    /// Sets the value at `idx` to `val`, returning the updated vector.
+    ///
+    /// This typically faster than updating via `as_mut_array`.
```
programmerjake, line 303:
```suggestion
    /// This is typically faster than updating an element via `as_mut_array`.
```

also for the other occurrences

programmerjake (approved):
this looks good to me other than the doc comment changes I noted.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/portable-simd/498/4215b7228d28cb84016e8f0e41c0a5e0935a222f..3bbbc4fd34e72eae2edcade0de6d04046910173f)*

PR #516, opened 2026-04-11
sync to `nightly-2026-04-11`

folkertdev:
Not sure the sync quite worked with that extra merge commit? maybe it's fine though.

I'd like to actually sync the other way, but that'll only work when we use an up-to-date nightly here.

I could also open this as my own commit with the 3 lines changed to what they should be, I'm not sure how that synchronizes.

### DIFF
--- a/crates/core_simd/src/simd/num/float.rs
+++ b/crates/core_simd/src/simd/num/float.rs
@@ -385,13 +385,13 @@ macro_rules! impl_trait {
             #[inline]
             fn simd_min(self, other: Self) -> Self {
                 // Safety: `self` and `other` are float vectors
-                unsafe { core::intrinsics::simd::simd_fmin(self, other) }
+                unsafe { core::intrinsics::simd::simd_minimum_number_nsz(self, other) }
             }
 
             #[inline]
             fn simd_max(self, other: Self) -> Self {
                 // Safety: `self` and `other` are floating point vectors
-                unsafe { core::intrinsics::simd::simd_fmax(self, other) }
+                unsafe { core::intrinsics::simd::simd_maximum_number_nsz(self, other) }
             }
 
             #[inline]
--- a/crates/std_float/tests/float.rs
+++ b/crates/std_float/tests/float.rs
@@ -33,23 +33,6 @@ macro_rules! unary_approx_test {
     }
 }
 
-macro_rules! binary_approx_test {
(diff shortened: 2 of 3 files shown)


folkertdev:
the bots are not happy with merge commits, so I guess this'll have to do.

What's the status on switching to a josh subtree?

calebzulawski:
If you ran the sync script then it's correct. I believe there should be a merge commit. rust-lang/rust has a rustbot rule to allow merge commits for subtree syncs

folkertdev:
Hmm, another precision issue 

https://triage.rust-lang.org/gha-logs/rust-lang/portable-simd/70922828512?pr=516


I'll submit a separate PR that displays that error in a nicer way, with that it renders as:

```
    thread 'f32::log::lanes_3' (1001) panicked at crates/test_helpers/src/lib.rs:149:5:
    Test failed: assertion failed: `(left == right)` 
      left: `[NaN (ffc00000), NaN (ffc00000), -1.7069013 (bfda7bbe)]`,
     right: `[NaN, NaN, -1.7069033]` at crates/test_helpers/src/lib.rs:374.
    minimal failing input: (
        [
            -4120549500000.0,
            inf,
            6.2068386e-14,
        ],
        [
            3.0238625e25,
            -1.0388e-38,
            54636652.0,
        ],
    )
```

At least for me locally this is the only test that fails. It could be that we just now hit an unlucky seed. Miri did double the margins recently (in https://github.com/rust-lang/miri/pull/4942) if both sides are subject to the error, which is the case here. We (correctly, because both sides have an error applied) already use the doubled ulps value of 16 though.

So, do we double it again? Or disable the random errors?

(I also noticed that we don't run these tests for `f16` yet, in part because https://crates.io/crates/float-cmp does not support `f16`. Another side quest)

folkertdev:
I think I have an answer: the implementation of `log` uses two inexact operations

```rust
    /// Produces a vector where every element has the logarithm with respect to an arbitrary
    /// in the equivalently-indexed elements in `self` and `base`.
    #[inline]
    #[must_use = "method returns a new vector and does not mutate the original value"]
    fn log(self, base: Self) -> Self {
        unsafe { intrinsics::simd_div(self.ln(), base.ln()) }
    }
```

and hence we need to increase the `ulps` for the comparison here.

PR #518, opened 2026-04-12
run `f16` float tests

folkertdev:
the `float-cmp` and `num-traits` libraries don't (yet) support `f16`, and likely won't do so before the type is stabilized.

Luckily it turns out that we only really need one function from them for checking whether two floats are "close enough". Miri already has this logic, which I've adapted here.

https://github.com/rust-lang/miri/blob/871d3d06c549cb554dec9a3a7c37cdd723254d06/tests/pass/float.rs#L40-L51

With that we can actually drop the dependency on `float-cmp`, which is neat too.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -72,15 +72,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5baebc0774151f905a1a2cc41989300b1e6fbb29aff0ceffa1064fdd3088d582"
 
 [[package]]
-name = "float-cmp"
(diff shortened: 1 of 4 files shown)


workingjubilee (approved):
Sweet.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/portable-simd/518/81e7d7d99ed1f2890f8e2a32291a27d7088e094e..eb7cb897ea71f57eed0495f1fbc12b7cca0e55df)*

PR #520, opened 2026-04-16
disable `f16` tests for the cranelift backend

### LINKED ISSUE #519: Missing cfg(target_has_reliable_f16_math) check for f16 tests
https://github.com/rust-lang/portable-simd/pull/513 added a bunch of f16 tests, which fail to link with cg_clif on x86_64-pc-windows-gnu due to missing intrinsics. cg_clif does declare f16 support as unreliable on this target, but portable-simd never checks it. This causes CI to fail for the rustc update. https://github.com/rust-lang/rustc_codegen_cranelift/actions/runs/24462910206/job/71481840311

folkertdev:
fixes rust-lang/portable-simd#519 

@bjorn3 does this look like it would work?

### DIFF
--- a/crates/core_simd/tests/f16_ops.rs
+++ b/crates/core_simd/tests/f16_ops.rs
@@ -1,5 +1,7 @@
 #![feature(portable_simd)]
 #![feature(f16)]
+#![expect(internal_features)]
+#![feature(cfg_target_has_reliable_f16_f128)]
 
 #[macro_use]
 mod ops_macros;
@@ -7,4 +9,5 @@ mod ops_macros;
 // FIXME: some f16 operations cause rustc to hang on wasm simd
 // https://github.com/llvm/llvm-project/issues/189251
 #[cfg(not(all(target_arch = "wasm32", target_feature = "simd128")))]
+#[cfg(target_has_reliable_f16_math)]
 impl_float_tests! { f16, i16 }
--- a/crates/core_simd/Cargo.toml
+++ b/crates/core_simd/Cargo.toml
@@ -31,3 +31,11 @@ path = "../test_helpers"
 
 [dev-dependencies]
 std_float = { path = "../std_float/", features = ["as_crate"] }
+
+[lints.rust.unexpected_cfgs]
+level = "warn"
+check-cfg = [
+    # Internal features aren't marked known config by default, we use these to
+    # gate tests.
+    'cfg(target_has_reliable_f16_math)',
+]


bjorn3:
It still fails: https://github.com/rust-lang/rustc_codegen_cranelift/actions/runs/24515088390/job/71656650354

folkertdev:
So actually it looks like the failure is in `proptest`? And there we certainly don't want to cfg on f16 math given that we want to stabilize `f16` soon.

The libc implementation seems doable to implement, actually? 

https://github.com/llvm/llvm-project/blob/b96818f2e0118fcbf908b0a168ca1d23190ac295/libc/src/__support/FPUtil/NearestIntegerOperations.h#L54-L105

bjorn3:
> And there we certainly don't want to cfg on f16 math given that we want to stabilize f16 soon.

There are still a bunch of cases where the LLVM backend also says it doesn't support f16 math. So either way `cfg(target_has_reliable_f16_math)` is necessary to make it possible to test portable-simd on those targets at all.

folkertdev:
I believe that is because we still support older LLVM versions? 

https://github.com/rust-lang/rust/blob/e8e4541ff19649d95afab52fdde2c2eaa6829965/compiler/rustc_codegen_llvm/src/llvm_util.rs#L373-L389

With LLVM 22 `f16` should work everywhere as far as I know. Unless we're running into this case here 

```rust
        (Arch::X86_64, Os::Windows) if *target_env == Env::Gnu && *target_abi != CfgAbi::Llvm => {
```

bjorn3:
To unblock the cg_clif rustc update, I have for now added a patch that remove all f16 support in portable-simd again.

crates/core_simd/tests/f16_ops.rs
```diff
@@ -1,10 +1,15 @@
...
 
 #[macro_use]
 mod ops_macros;
 
 // FIXME: some f16 operations cause rustc to hang on wasm simd
 // https://github.com/llvm/llvm-project/issues/189251
 #[cfg(not(all(target_arch = "wasm32", target_feature = "simd128")))]
+// FIXME: the cranelift backend runs these tests for x86_64-pc-windows-gnu
+// where it does not have full support for f16 yet.
+#[cfg(target_has_reliable_f16_math)]
```
tgross35: I'd drop this comment, it's not a cranelift-specific issue and the `target_has_reliable_f16_math` is effectively already a fixme.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/portable-simd/520/81e7d7d99ed1f2890f8e2a32291a27d7088e094e..1f19d30c40349566cbb9eb713db771ef73806943)*

tgross35:
At this time I'd still consider it more or less a requirement to gate anything that may codegen behind `cfg(target_has_reliable_f16)`. We still have these edge cases and they're kind of bootstrapped by the fact that the `cfg` is used pretty consistently (out of necessity) like here. It will be at least ~4 months until our minimum LLVM version is high enough to drop the `cfg` completely.

calebzulawski:
Should implementation of `SimdElement` for f16 be gated on reliable f16 math?

folkertdev:
It makes no difference

```rust
    // Assume that working `f16` means working `f16` math for most platforms, since
    // operations just go through `f32`.
    cfg.has_reliable_f16_math = cfg.has_reliable_f16;
```

https://github.com/rust-lang/rust/blob/e8e4541ff19649d95afab52fdde2c2eaa6829965/compiler/rustc_codegen_llvm/src/llvm_util.rs#L416

Unless that gets overridden somehow.

workingjubilee (approved):

PR #526, opened 2026-05-08
Fix pointer API to match what was stabilized

calebzulawski:


crates/core_simd/src/simd/ptr.rs
```diff
@@ -9,3 +9,49 @@ mod sealed {
...
+
+/// Converts addresses back to pointers, picking up some previously "exposed" provenance.
+///
+/// Equivalent to calling [`core::ptr::with_exposed_provenance`] on each element.
+#[inline]
+pub fn with_exposed_provenance<T, const N: usize>(addr: Simd<usize, N>) -> Simd<*const T, N> {
+    // FIXME(strict_provenance_magic): this should be backed by a dedicated SIMD intrinsic,
+    // but currently that path can ICE on some nightly toolchains.
+    // SAFETY: Integer-to-pointer transmutes are valid and model pointer-int-pointer casts.
+    unsafe { core::mem::transmute_copy(&addr) }
```
programmerjake: is there an issue for the ICEs?
saying integer to pointer transmutes are valid seems to contradict the comments in `without_provenance` above, as well as the docs in latest nightly: https://github.com/rust-lang/rust/blob/63b1dfc0e00fd6f8ad7cd8817fc712e7d9b7be59/library/core/src/ptr/mod.rs#L942-L946

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/portable-simd/526/cc032ad5ed5b69d606c4eabbc9c9ae1f412ef2fb..bb0ef6c0822616dbd1a52a5bc9fae04a8038b1c3)*

programmerjake: It also contradicts the stable docs of `transmute`: https://doc.rust-lang.org/std/mem/fn.transmute.html#transmutation-between-pointers-and-integers

I think that means we'll have to wait until the intrinsic is working properly.

calebzulawski: Whoops, that was an outdated comment from when I saw another problem. We were already using transmutes, but do you think we should use the scalar functions for now?

programmerjake: yeah, using the scalar functions is at least correct.

calebzulawski: I confused myself, I had copy paste mistakes and thought we were talking about `without_provenance`. I fixed `with_exposed_provenance` and updated the comment on `without_provenance` to match std

programmerjake (approved):

PR #528, opened 2026-05-28
Optimize `swizzle_dyn` for AArch64 with N > 16

Kmeakin:
We can do swizzles for 24, 32, 48 and 64 byte vectors by stacking multiple TBL instructions.

See https://godbolt.org/z/PE95nrqjj for a comparison of the generated assembly.

I would do the same optimization for 32-bit Arm but I can't get at the intrinsics for some reason: https://godbolt.org/z/54jqTEeeW


programmerjake:
to ensure it's actually tested, you'll need to uncomment the lane counts that you have special cases for, so uncomment 48 here: https://github.com/rust-lang/portable-simd/blob/94a3ba2675fb37f5683323a7cfd29a888d737f23/crates/test_helpers/src/lib.rs#L681

programmerjake (approved):

calebzulawski:
Looks good, thank you!

PR #529, opened 2026-05-30
Use `impl` restrictions over hand-sealed traits

jhpratt:
As part of rust-lang/rust#105077, I'm beginning to use `#![feature(impl_restriction)]` within the standard library. As portable SIMD is relied upon by the standard library (as a submodule), I thought I would start here.

I have tested the feature myself in a number of situations and have found it works as expected. Feel free to ask any questions you may have.

### DIFF
--- a/crates/core_simd/src/masks.rs
+++ b/crates/core_simd/src/masks.rs
@@ -29,7 +29,7 @@ macro_rules! impl_fix_endianness {
 
 impl_fix_endianness! { u8, u16, u32, u64 }
 
-mod sealed {
+mod private_methods {
     use super::*;
 
     /// Not only does this seal the `MaskElement` trait, but these functions prevent other traits
@@ -38,7 +38,7 @@ mod sealed {
     /// For example, `eq` could be provided by requiring `MaskElement: PartialEq`, but that would
     /// prevent us from ever removing that bound, or from implementing `MaskElement` on
     /// non-`PartialEq` types in the future.
-    pub trait Sealed {
+    pub impl(super) trait PrivateMethods {
         fn valid<const N: usize>(values: Simd<Self, N>) -> bool
         where
             Self: SimdElement;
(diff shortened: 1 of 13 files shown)


jhpratt:
Note: `rustdoc` does not currently have support for `impl` restrictions, so if that's a concern then you'll want to hold off.

calebzulawski:
Because std::simd is just a module of the core library, we can't actually use this feature until core does.

jhpratt:
Gotcha. rust-lang/rust#157170 will need to be merged first then.

jhpratt:
FYI that was merged, so there should be no blockers for this now.

calebzulawski:
Thanks! @programmerjake any thoughts?

crates/core_simd/src/masks.rs
```diff
@@ -38,7 +38,7 @@ mod sealed {
     /// For example, `eq` could be provided by requiring `MaskElement: PartialEq`, but that would
     /// prevent us from ever removing that bound, or from implementing `MaskElement` on
     /// non-`PartialEq` types in the future.
-    pub trait Sealed {
+    pub trait PrivateMethods {
```
programmerjake: shouldn't this also be `impl(self)`? (or `super`)

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/portable-simd/529/be090a7e3622887fe0350686844e1c8bd6e962c8..d0145147c262ac782ed220a139a44b00b777a0b4)*

programmerjake:
other than that, looks good to me! Glad that rustdoc has support now.

crates/core_simd/src/masks.rs
jhpratt: Not strictly necessary because it's still the "sealing" trait, but I've added it for consistency.

programmerjake (approved):

calebzulawski (approved):

calebzulawski:
Thank you!

PR #530, opened 2026-05-31
Provide `From` impls for AArch64 ACLE tuple types.

Kmeakin:
`From` impls were already provided for the normal AArch64 ACLE vector types, but not for the x2,x3,x4 tuples of ACLE vectors. Rectify that.

Also provide impls for vectors of `f16`, as I noticed they were missing.


programmerjake (approved):

calebzulawski:
Thanks!

PR #531, opened 2026-06-12
Bump `rand` version to address GHSA-cq8v-f236-94qc.

ruyrocha:
https://github.com/advisories/GHSA-cq8v-f236-94qc


workingjubilee:
We do not meet the other criteria for this advisory being applicable.

PR #533, opened 2026-06-22
Optimize `swizzle_dyn` for LoongArch64 with N is 16 or 32

heiher:
We can optimize swizzles for 16-byte and 32-byte vectors using LoongArch LSX/LASX shuffle instructions.

For reference, see the Unofficial LoongArch Intrinsics Guide:
https://jia.je/unofficial-loongarch-intrinsics-guide

I've run all `swizzle_dyn::lanes_*` tests locally, and they all passed.

### DIFF
--- a/crates/core_simd/src/swizzle_dyn.rs
+++ b/crates/core_simd/src/swizzle_dyn.rs
@@ -48,6 +48,8 @@ impl<const N: usize> Simd<u8, N> {
                     target_endian = "little"
                 ))]
                 16 => transize(armv7_neon_swizzle_u8x16, self, idxs),
+                #[cfg(all(target_arch = "loongarch64", target_feature = "lsx"))]
+                16 => transize(loong64_lsx_swizzle, self, idxs),
                 #[cfg(all(target_feature = "avx2", not(target_feature = "avx512vbmi")))]
                 32 => transize(avx2_pshufb, self, idxs),
                 #[cfg(all(target_feature = "avx512vl", target_feature = "avx512vbmi"))]
@@ -62,6 +64,8 @@ impl<const N: usize> Simd<u8, N> {
                     };
                     transize(swizzler, self, idxs)
                 }
+                #[cfg(all(target_arch = "loongarch64", target_feature = "lasx"))]
+                32 => transize(loong64_lasx_swizzle, self, idxs),
                 // Notable absence: avx512bw pshufb shuffle
                 #[cfg(all(target_feature = "avx512vl", target_feature = "avx512vbmi"))]
                 64 => {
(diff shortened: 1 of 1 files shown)


crates/core_simd/src/swizzle_dyn.rs
```diff
@@ -220,6 +224,42 @@ unsafe fn avx2_pshufb(bytes: Simd<u8, 32>, idxs: Simd<u8, 32>) -> Simd<u8, 32> {
...
+///
+/// # Safety
+/// This requires LoongArch LSX to work
+#[cfg(all(target_arch = "loongarch64", target_feature = "lsx"))]
+unsafe fn loong64_lsx_swizzle(bytes: Simd<u8, 16>, idxs: Simd<u8, 16>) -> Simd<u8, 16> {
+    use core::arch::loongarch64::{lsx_vand_v, lsx_vandi_b, lsx_vshuf_b, lsx_vslei_bu};
+    // SAFETY: Caller promised loongarch lsx support
+    unsafe {
+        let mask = lsx_vslei_bu::<15>(idxs.into());
+        let idxs = lsx_vandi_b::<15>(idxs.into());
```
programmerjake: is the bitwise-and necessary? can't you just leave the original indexes since the output is masked? or is out-of-bounds indexes some sort of UB here?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/portable-simd/533/d7a525f66ace336c4b67bf1576c4ce341ac110c9..9ef2797a2266e8ca935268d4557b77e5e8a66ef1)*

crates/core_simd/src/swizzle_dyn.rs
```diff
@@ -220,6 +224,42 @@ unsafe fn avx2_pshufb(bytes: Simd<u8, 32>, idxs: Simd<u8, 32>) -> Simd<u8, 32> {
...
+unsafe fn loong64_lasx_swizzle(bytes: Simd<u8, 32>, idxs: Simd<u8, 32>) -> Simd<u8, 32> {
+    use core::arch::loongarch64::{
+        lasx_xvand_v, lasx_xvandi_b, lasx_xvpermi_q, lasx_xvshuf_b, lasx_xvslei_bu,
+    };
+    // SAFETY: Caller promised loongarch lasx support
+    unsafe {
+        let lolo = lasx_xvpermi_q::<0x00>(bytes.into(), bytes.into());
+        let hihi = lasx_xvpermi_q::<0x11>(bytes.into(), bytes.into());
+        let mask = lasx_xvslei_bu::<31>(idxs.into());
+        let idxs = lasx_xvandi_b::<31>(idxs.into());
```
programmerjake: same here

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/portable-simd/533/d7a525f66ace336c4b67bf1576c4ce341ac110c9..9ef2797a2266e8ca935268d4557b77e5e8a66ef1)*

crates/core_simd/src/swizzle_dyn.rs
heiher: Good question.

According to the LoongArch ISA manual (Chinese) for `[X]VSHUF.B`:

```
When the upper three bits of a byte containing an index value are non-zero, the selection
result is unpredictable.
```

The wording is somewhat ambiguous and does not explicitly state whether "unpredictable" applies to the corresponding output element only or to the entire instruction result.

From testing on real hardware, the behavior appears to be per-element rather than per-instruction: lanes with out-of-range indices may produce unspecified results, while other lanes are unaffected. This suggests that the manual's "unpredictable" wording should be interpreted on a lane-by-lane basis.

Given that the output is already masked afterwards, I think you're right that the extra bitwise-and on the indices is unnecessary, and we can probably remove it.

programmerjake:
> I've run all `swizzle_dyn::lanes_*` tests locally, and they all passed.

do you have LASX? afaik it isn't tested unless you enable it as a target feature when building portable-simd, since afaict CI doesn't enable it.

heiher:
> do you have LASX? afaik it isn't tested unless you enable it as a target feature when building portable-simd, since afaict CI doesn't enable it.

Yeah. I tested it locally with:

`RUSTFLAGS="-C target-feature=+lasx" cargo test -r`

```asm
000000000005ae80 <_RNvNtNtCs1775If9jsmw_9core_simd9core_simd11swizzle_dyn19loong64_lsx_swizzle>:
   5ae80:	2c0000a0 	vld         	$vr0, $a1, 0
   5ae84:	2c0000c1 	vld         	$vr1, $a2, 0
   5ae88:	0d508000 	vshuf.b     	$vr0, $vr0, $vr0, $vr1
   5ae8c:	72843c21 	vslei.bu    	$vr1, $vr1, 0xf
   5ae90:	71260400 	vand.v      	$vr0, $vr0, $vr1
   5ae94:	2c400080 	vst         	$vr0, $a0, 0
   5ae98:	4c000020 	ret

000000000005aea0 <_RNvNtNtCs1775If9jsmw_9core_simd9core_simd11swizzle_dyn20loong64_lasx_swizzle>:
   5aea0:	2c8000a0 	xvld        	$xr0, $a1, 0
   5aea4:	2c8000c1 	xvld        	$xr1, $a2, 0
   5aea8:	77d40002 	xvori.b     	$xr2, $xr0, 0x0
   5aeac:	77ec0002 	xvpermi.q   	$xr2, $xr0, 0x0
   5aeb0:	77ec4400 	xvpermi.q   	$xr0, $xr0, 0x11
   5aeb4:	0d608800 	xvshuf.b    	$xr0, $xr0, $xr2, $xr1
   5aeb8:	76847c21 	xvslei.bu   	$xr1, $xr1, 0x1f
   5aebc:	75260400 	xvand.v     	$xr0, $xr0, $xr1
   5aec0:	2cc00080 	xvst        	$xr0, $a0, 0
   5aec4:	4c000020 	ret
```

programmerjake (approved):

PR #535, opened 2026-06-24
explain SIMD terminology and commonly-confused terms

enthropy7:
Part of rust-lang/portable-simd#29. Second PR in the series.

This PR broadens the beginner vocabulary and adds new terms, bringing the guide closer to how the finished documentation should look.

Addresses these checklist items:

- [x] "field"
- [x] "element"
- [x] "intrinsics"
- [x] intrinsics vs. instructions
- [x] instructions vs. operations
- [x] lane vs. field vs. element
- [x] scalar vs. vector

New `Terms` entries: `Element`, `Splat`, `Mask`, `Horizontal`, and `Intrinsic` (covering both vendor `core::arch` intrinsics and compiler/LLVM intrinsics, so readers can see what intrinsics actually look like). New "Words That Are Easy to Mix Up" section for the four distinctions above. `Splat`/`Mask`/`Horizontal` aren't on the checklist, but they're core beginner-facing vocabulary for this crate (splat appears in the README's hello-world, masks have a whole README section, and horizontal pairs with the existing "vertical" term), so I included them while we're here.

PR #536, opened 2026-06-24
explain how SIMD relates to SIMT, SPMD, scalable hardware, and tensors

enthropy7:
Part of rust-lang/portable-simd#29. My third PR in the series.

Adds the conceptual "how does SIMD relate to other models and hardware" material, placed right after Quick Background so it reads as the landscape before the vocabulary. 

Fills these checklist gaps:

- [x] Differences between CPU SIMD and GPU SIMT
- [x] Differences between SIMD vectors and SPMD threads
- [x] Differences between "wide vectors" and "vector architectures" (Cray, RISC-V V, Arm SVE)

added two new sections:
- SIMD, SIMT, and SPMD - warp/wavefront and divergence; why a vector's lanes aren't independent threads (lockstep, masking instead of branches); ISPC as SPMD-on-CPU-SIMD.
- Fixed-Width vs. Scalable Vectors - packed SSE/AVX/NEON vs. length-agnostic Cray/RVV/SVE (RVV `vsetvl`, SVE vector length-agnostic loops), with concrete widths in the wild (A64FX 512-bit, Graviton3 256-bit, most Armv9 mobile 128-bit).
Notes that portable-simd's fixed-size types lower to scalable instructions where that's better, but it has no variable-size types yet - expanding on the point @programmerjake raised in rust-lang/portable-simd#535.

NB: the issue already had "SIMD vectors vs SPMD threads" checked, but the guide had no text on it, which was corrected in this PR.


workingjubilee:
My thinking has changed somewhat over time and I don't really think we need a bunch of documentation on tensors and all that anymore.

enthropy7:
> My thinking has changed somewhat over time and I don't really think we need a bunch of documentation on tensors and all that anymore.

so, we should get rid of tensors and related stuff here?

workingjubilee:
Sorry, to clarify, by "all that" I mean this entire PR's contents. My apologies.

PR #538, opened 2026-07-04
cleanup `FlushSubnormals` implementation

folkertdev:
let's see if that powerpc fix made it into our qemu.


folkertdev:
cc @calebzulawski looks like that workaround is no longer needed.

folkertdev:
@calebzulawski ping?

calebzulawski:
Whoops, thanks!

PR #542, opened 2026-07-22
Optimize swizzle_dyn AVX2 implementation

Shnatsel:
llvm-mca shows:

- no change on Haswell and Broadwell
- 20% improvement on Skylake when using the 512-bit decomposition from rust-lang/portable-simd#540, no change otherwise
- 23% improvement on Zen 1, Zen 2 and Tiger Lake
- 7-20% improvement on Zen 3

End-to-end benchmarks on a [toy base64 implementation](https://github.com/Shnatsel/intrepid-base64) using 512-bit shuffles improve by 7% to 13% on my Zen4 workstation, depending on how you count.

<details><summary>Benchmark details</summary>
<p>

```
$ RUSTFLAGS='--cfg disable_dispatch_avx512' cargo bench  --bench codec -- 1048576
    Finished `bench` profile [optimized] target(s) in 0.02s
     Running benches/codec.rs (target/release/deps/codec-0676c1840143d734)
encode/1048576          time:   [103.20 µs 103.31 µs 103.45 µs]
                        thrpt:  [9.4398 GiB/s 9.4524 GiB/s 9.4631 GiB/s]
                 change:
                        time:   [−11.900% −11.804% −11.685%] (p = 0.00 < 0.05)
                        thrpt:  [+13.231% +13.384% +13.507%]
                        Performance has improved.
Found 13 outliers among 100 measurements (13.00%)
  12 (12.00%) high mild
  1 (1.00%) high severe

decode/1048576          time:   [163.61 µs 163.63 µs 163.65 µs]
                        thrpt:  [5.9674 GiB/s 5.9682 GiB/s 5.9689 GiB/s]
                 change:
                        time:   [−7.6401% −7.5954% −7.5538%] (p = 0.00 < 0.05)
                        thrpt:  [+8.1710% +8.2197% +8.2721%]
                        Performance has improved.
Found 6 outliers among 100 measurements (6.00%)
  4 (4.00%) high mild
  2 (2.00%) high severe
```

</p>
</details> 

This is a port/upstreaming of my work on `fearless_simd`: https://github.com/linebender/fearless_simd/pull/276

Disassembly and llvm-mca measurements are from `fearless_simd` because this repo does not include instructions on how to disassemble and benchmark it (as opposed to the `std::simd` re-export of its older version).

Correctness of this formulation was verified using random testing against the scalar reference on 100k inputs: https://github.com/linebender/fearless_simd/pull/276/changes/4e8a56accb23ab4f0b96db89147b3c1fadc8f07e
[...]

PR #543, opened 2026-07-25
Optimize swizzle_dyn on ssse3

Shnatsel:
The idea was [suggested](https://github.com/rust-lang/portable-simd/pull/542#issuecomment-5077221668) by @axnsan12. 

Changes the control sequence from vpcmpgtb + vpblendvb to vpaddb + vpor. 15% throughput improvement on Haswell, 20% improvement on skylake, zen1, zen3. For the 256-bit case, 38% improvement on haswell, 12% on skylake, zen1 and zen3 unchanged. Disassembly measured [on fealrless_simd](https://github.com/linebender/fearless_simd/pull/276) similarly to rust-lang/portable-simd#542.

Expands test coverage to cover byte values 128-255 which would misbehave without `idxs | zeroing_bits`.

### DIFF
--- a/crates/core_simd/src/swizzle_dyn.rs
+++ b/crates/core_simd/src/swizzle_dyn.rs
@@ -338,7 +338,8 @@ unsafe fn transize<T, const N: usize>(
 #[allow(unused)]
 #[inline(always)]
 fn zeroing_idxs<const N: usize>(idxs: Simd<u8, N>) -> Simd<u8, N> {
-    use crate::simd::{Select, cmp::SimdPartialOrd};
-    idxs.simd_lt(Simd::splat(N as u8))
-        .select(idxs, Simd::splat(u8::MAX))
+    // Adding this sets the high bit for indices N..=127, while PSHUFB ignores
+    // the other changed bits. The OR preserves the high bit for indices 128..=255.
+    let zeroing_bits = idxs + Simd::splat((127 - N + 1) as u8);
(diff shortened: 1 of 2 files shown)


SuperSamus:
Couldn't this version of `zeroing_idxs` also be used on the AVX2 shuffle like this?
```rs
        let lolo = avx2_cross_shuffle::<0x00>(bytes.into(), bytes.into());
        let hihi = avx2_cross_shuffle::<0x11>(bytes.into(), bytes.into());

        let control = zeroing_idxs(idxs).into();

        // Move index bit 4 into each byte's sign bit for VPBLENDVB.
        let select_high = x86::_mm256_slli_epi16::<3>(idxs.into());
        let from_low = avx2_half_pshufb(lolo, control);
        let from_high = avx2_half_pshufb(hihi, control);
        x86::_mm256_blendv_epi8(from_low, from_high, select_high).into()
```

Shnatsel:
You could, but according to llvm-mca that keeps the throughput unchanged and makes latency slightly worse, so there's no point in doing that.

axnsan12:
Yeah I think my suggestion to use wrapping_add was wrong because I misremembered my experiments from ~1 year ago. saturating_add (expecting paddsb to be emitted) would actually be correct and not require the additional OR.

And IIRC after doing that the custom avx2 shuffle became unnecessary because LLVM just generated the same code using the generic recursive one

